PR #1, opened 2015-08-28
Add missing webtest module in install_requires

asaunier:
When typing
```
make -f config/alex serve
```
I had the following error:
```
...
File "/home/alex/v6_api/app_api/tests/__init__.py", line 7, in <module>
    from webtest import TestApp
ImportError: No module named webtest
```

Is it the correct place to add the module?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@ requires = [
     'cornice',
     'colander',
     'ColanderAlchemy>=0.3.2',
+    'webtest',
     ]
 
 setup(name='app_api',


tsauerwein:
`webtest` is already listed in the [dev-requirements.txt](https://github.com/c2corg/v6_api/blob/master/dev-requirements.txt#L4). The `dev-requirements` are not installed when simply running the target `serve`.

I think we should not add `dev-requirements` as a requirment for the target `serve`, because we would have the same problem when running on Apache. The actual question is: Why is the `tests` package loaded when running the application? This is because [config.scan()](http://docs.pylonsproject.org/projects/pyramid//en/latest/api/config.html#pyramid.config.Configurator.scan) is used in the [init.py](https://github.com/c2corg/v6_api/blob/master/app_api/__init__.py#L21). Here we should ignore the `tests` package:

    config.scan(ignore='app_api.tests')

elemoine:
Or not using `config.scan` at all… You can do without `config.scan` if you imperatively declare routes and views using `config.add_view` and `config.add_route`.

asaunier:
> The dev-requirements are not installed when simply running the target serve.

Well, I think I had run the "install" target first but had the problem anyway!?

elemoine:
And it's when you run the `test` or `flake8` target the first time that the dev requirements are installed. I'd just use add_view and remove the scanning phase.

tsauerwein:
> I'd just use add_view and remove the scanning phase. 

This would be more explicit, but I am not sure how we can do that with the `@resource` annotation of Cornice.

elemoine:
I am not a Cornice expert but can't you use the `Service` constructor rather than the `@resource` decorator? I don't know really. Just wondering…

tsauerwein:
With Cornice you either directly define a [service](https://cornice.readthedocs.org/en/latest/tutorial.html#defining-the-services) or use a [resource](https://cornice.readthedocs.org/en/latest/resources.html). I decided to use resources because then the services that belong together are nicely grouped into a class (e.g. all services for waypoints in one class).

tsauerwein:
I [committed](https://github.com/c2corg/v6_api/commit/4e669a0af412ecc14cb61186a15174a057567011) the change with `config.scan(ignore=...)` to make the demo work on Apache. This fixes the original issue.

Feel free to open the ticket again, if you further want to discuss `config.scan` vs. explicit route definitions.

PR #2, opened 2015-09-10
Add models/views/tests for routes

asaunier:
It's mostly copy-pasted from the waypoints classes, adapting the attributes.
Next step is to factorize the common parts between waypoints, routes, etc. I think.

### DIFF
--- a/app_api/models/route.py
+++ b/app_api/models/route.py
@@ -1,0 +1,137 @@
+from sqlalchemy import (
+    Column,
+    Integer,
+    SmallInteger,
+    String,
+    ForeignKey,
+    Enum
+    )
+
+from colanderalchemy import SQLAlchemySchemaNode
+
+from . import schema
+from utils import copy_attributes
+from document import (
+    ArchiveDocument, Document, DocumentLocale, ArchiveDocumentLocale)
+
+# TODO: move to a common place for outings etc.
+activities = [
+    'skitouring',
+    'snow_ice_mixed',
+    'mountain_climbing',
+    'rock_climbing',
+    'ice_climbing',
+    'hiking',
+    'snowshoeing',
+    'paragliding',
+    'mountain_biking',
+    'via_ferrata'
+    ]
+
+
+class _RouteMixin(object):
+    activities = Column(
+        Enum(name='activities', inherit_schema=True, *activities),
+        nullable=False)
+
+    height = Column(SmallInteger)
+
+    __mapper_args__ = {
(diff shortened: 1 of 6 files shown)


app_api/models/route.py
```diff
@@ -0,0 +1,137 @@
...
+    'snow_ice_mixed',
+    'mountain_climbing',
+    'rock_climbing',
+    'ice_climbing',
+    'hiking',
+    'snowshoeing',
+    'paragliding',
+    'mountain_biking',
+    'via_ferrata'
+    ]
```
elemoine: Is this actually used?

app_api/models/route.py
```diff
@@ -0,0 +1,137 @@
...
+        copy_attributes(self, route, Route._ATTRIBUTES)
+
+        return route
+
+    def get_archive_locales(self):
+        locales = []
+
+        for locale in self.locales:
+            archive_local = locale.to_archive()
+            locales.append(archive_local)
```
elemoine: How about using a list comprehension?

```py
    def get_archive_locales(self):
        return [locale.to_archive() for locale in self.locales]
```

app_api/models/route.py
asaunier: Yes i'ts used at line 34.
"Activities" is one of the main attributes of routes and outings. See for instance: http://www.camptocamp.org/routes/54417/fr/petite-dent-de-veisivi-traversee-des-aretes

By the way routes/outings may have several activities at the same time (for instance: http://www.camptocamp.org/routes/46065/fr/cheval-blanc-depuis-emosson)
=> not sure the current Enum type is correct?

elemoine: Not sure either. I guess this can be changed later.

asaunier: Yep, that's what I thought when adding the attribute. :)

app_api/tests/__init__.py
```diff
@@ -61,6 +61,10 @@ def setUp(self):  # noqa
         self.trans = connection.begin()
 
         # bind an individual Session to the connection
+        # Next line is needed to make several tests run in a row.
+        # See https://github.com/Pylons/webtest/issues/5
+        # FIXME Is there a better solution?
+        DBSession.remove()
```
elemoine: That part is weird. But that entire setup process looks weird to me. I'd like to take a look at it, but I suggest that you merge your branch first.

asaunier:
OK thanks @elemoine !
Merging

PR #10, opened 2015-09-11
Database session-related clean-up in the tests

elemoine:
This PR suggests changes to the code that configures `DBSession` and the test Session (`self.session`) in the tests.

With this change the SQLAlchemy warning "SAWarning: At least one scoped session is already present.  configure() can not affect sessions that have already been created." goes away.


asaunier:
Thanks for the improvements!
I have tested in my instance, tests pass and actual actions (read/list/create) work as expected.
Looks good to me.

PR #11, opened 2015-09-11
Factorize code in views and models

asaunier:


asaunier:
I think the PR is ready for reviewing.

I have not really factorized the ``to_archive()`` function in the models. Is it really relevant since the function is small as for now.
By the way I have noticed that we programmatically copy the current doc version in the archives tables (instead of using triggers in the DB as first planned). Why not if not too complicated.

elemoine:
Looks good to me.

Issue #13, opened 2015-09-11
Use GeoJSON format


asaunier:
Since most of the items returned by the API are georeferenced objects, we should change the output to GeoJSON, not only JSON.

PR #14, opened 2015-09-14
[WIP] Add support of geometries

asaunier:
[Work in progress]

This PR adds the support of the geometry column in the waypoints tables. A "geom" attribute is than available in the API output (as for now only for waypoints).

Tests pass and everything looks normal when no geom is set. But if setting a WP geom (using for instance ``ST_GeomFromText('POINT(764132.607042082 5753542.30518111)', 3857)`` in psql, requesting the waypoint returns an error:
```
AttributeError
AttributeError: Neither 'Function' object nor 'Comparator' object has an attribute 'items'
```

The colander geometry type is taken from c2cgeoform: https://github.com/camptocamp/c2cgeoform/blob/master/c2cgeoform/ext/colander_ext.py

A next step is most likely to return GeoJSON instead of JSON as the API output format.

Related to #13

### DIFF
--- a/c2corg_api/ext/colander_ext.py
+++ b/c2corg_api/ext/colander_ext.py
@@ -1,0 +1,89 @@
+# Inspired from c2cgeoform colander extension
+# https://github.com/camptocamp/c2cgeoform/blob/master/c2cgeoform/ext/
+
+from colander import (null, Invalid)
+
+from geoalchemy2 import WKBElement
+from geoalchemy2.shape import to_shape, from_shape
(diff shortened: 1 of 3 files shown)


asaunier:
I have made a rebase to integrate a change in the geom field definition. The comments about the geom defintiion are available in the old commit: https://github.com/c2corg/v6_api/commit/6efb29134e8a212fc628baf4be10925edd50e87e#commitcomment-13230880

tsauerwein:
See #42

Issue #6, opened 2015-09-11
Server caching


asaunier:
Caching the output of some of the API actions would be relevant to increase the performances of the API. Using dogpile for instance.

I don't think caching the list output is relevant because new items are often added. On the other hand existing documents do not change every day: it's probably a good idea to cache the "get" output (perhaps the history as well).

asaunier:
We need to pay attention that we have many documents available and that caching all of them might take quite a lot of space (in v5 the symfony cache grows really big).
We will probably also use Varnish in front of the API.

mfournier:
One big issue we have with the current cache implementation is the overhead of using the filesystem. Dazillions of small files and directories hurt performance and are a nightmare to handle in production.

According to the doc, dogpile offers configurable cache backends, redis and memcached among others, which won't suffer from this problem. So it at least offers the flexibility we lack currently.

Regarding caching elements or not, IMHO the question to ask is not if the element is frequently changed or not but rather what is the cost of computing this piece of data.

Thinking this way you don't need to worry about the dilemma of which TTL to set on cached objects, but rather on an efficient strategy to purge or update outdated cached data (or change the identifier and let the LRU get rid of them).

HTH

tsauerwein:
> Regarding caching elements or not, IMHO the question to ask is not if the element is frequently changed or not but rather what is the cost of computing this piece of data. 

> Thinking this way you don't need to worry about the dilemma of which TTL to set on cached objects, but rather on an efficient strategy to purge or update outdated cached data (or change the identifier and let the LRU get rid of them).

Good point!

asaunier:
Should we have server caching both in the API app and in the UI app?

Having server caching in the API would make sense because it might be accessed by several clients (UI as well as mobile apps, etc.).

On the other hand, the UI pages will probably be server-cached as well.
It means that when a change is done, the cache should be invalidated both in the UI and in the API.

What about using Varnish in front of the API instead of using some server caching? How to make Varnish know that a document has changed?

asaunier:
Some precisions from @mfournier 

>There's 2 levels of caching here (in-app using dogpile, and at the
http-level using http-protocol mechanisms). I think we should use both for
both services.

>You shouldn't look too much at the varnish config from v5 for inspiration.
This was slapped onto the system several years after it's inception, partly
to decrease the number of apache workers by serving static content directly
from the cache, and partly to work around perf issues we weren't able to
fix in the code (slow queries generated from the ORM) on highly requested
[...]

PR #15, opened 2015-09-15
Add sever side caching

asaunier:
This PR makes use of dogpile.cache to cache the output of the "get" method (as for now lists are not cached).

I have used "as is" the dogpile implementation from c2cgeoportal:
https://github.com/camptocamp/c2cgeoportal/blob/1.6/c2cgeoportal/lib/caching.py
We should probably check if all this functions are needed or generic enough (at first sight yes). It provides for instance a ``set_common_headers()`` function to set the cache control HTTP headers.

The backend is currently the standard dogpile.cache.memory backend. dogpile also supports advanced backend such as memcached or Redis:
https://dogpilecache.readthedocs.org/en/latest/api.html#module-dogpile.cache.backends.memory
Using one of them would probably improve the performances. Please note that Discourse uses a Redis cache: http://www.discourse.org/faq/#tech

By the way the configuration of dogpile is at the moment hardcoded (even though I have added it in common.ini.in as well). Using the config files would indeed make sense but I was wondering if would use those ini file or another system such as yaml files.

No test is provided. Not sure it is easy to test if the cache works as expected. At least we could for instance if the attributes values of a WP are unchanged after 2 requests whereas the value in the DB has actually changed? We would need also to test the invalidation of the cache when a new version of the document is created (not available yet).

Related to #6

### DIFF
--- a/app_api/__init__.py
+++ b/app_api/__init__.py
@@ -6,6 +6,8 @@ from .models import (
     Base,
     )
 
+from .ext import caching
+
 
 def main(global_config, **settings):
     """ This function returns a Pyramid WSGI application.
@@ -15,5 +17,12 @@ def main(global_config, **settings):
     Base.metadata.bind = engine
     config = Configurator(settings=settings)
     config.include('cornice')
+
+    # dogpile.cache configuration
+    # TODO: get config from settings
+    # caching.init_region(settings["cache"])
+    caching.init_region({'backend': 'dogpile.cache.memory'})
+    caching.invalidate_region()
+
     config.scan(ignore='app_api.tests')
     return config.make_wsgi_app()
--- a/app_api/ext/caching.py
+++ b/app_api/ext/caching.py
@@ -1,0 +1,151 @@
+# -*- coding: utf-8 -*-
+
+# Taken from c2cgeoportal at
+# https://github.com/camptocamp/c2cgeoportal/blob/1.6/c2cgeoportal/lib/caching.py
+
+# Copyright (c) 2012-2015, Camptocamp SA
+# All rights reserved.
+
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+
+# 1. Redistributions of source code must retain the above copyright notice,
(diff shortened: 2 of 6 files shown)


asaunier:
Concerning the choice of Redis or Memcached:
http://www.infoworld.com/article/2825890/application-development/why-redis-beats-memcached-for-caching.html

In short: Redis has more advanced features but for string storage (JSON or HTML output for instance) memcached seems more performant. What about storing objects such as model instances?

As mentioned above, Discourse uses Redis, which means we will need it anyway.

tsauerwein:
> No test is provided. Not sure it is easy to test if the cache works as expected. At least we could for instance if the attributes values of a WP are unchanged after 2 requests whereas the value in the DB has actually changed? We would need also to test the invalidation of the cache when a new version of the document is created (not available yet).

In the tests you could directly query the cache, see http://dogpilecache.readthedocs.org/en/latest/api.html#dogpile.cache.region.CacheRegion.cache_on_arguments

     view._get_by_id.get(1)
     view._get_by_id.invalidate(1)

elemoine:
> By setting a fixed max_age don't we risk that a browser is using an old version of a document even though a newer is available? Wouldn't it make more sense to use etag (e.g. the version number) and/or last-modified?

Yes, maybe. Cliquet uses Etag. See http://cliquet.readthedocs.org/en/latest/api/timestamps.html. They also rely on Etag to avoid race conditions, which may be relevant to our project as well.

But note that the `set_common_headers` function, which is the one setting `max_age`, is not currently used. I think it should be removed.

asaunier:
> But note that the set_common_headers function, which is the one setting max_age, is not currently used. I think it should be removed.

If we are sure this function is useless and/or not adapted to our needs, we should indeed remove it.

> Cliquet uses Etag.

I have noticed that you mentioned Cliquet (built upon Cornice) as a possible tool for building the API. Why haven't we used it finally?

asaunier:
Since https://github.com/c2corg/v6_api/pull/336 I guess this PR is no longer relevant and can be closed, its branch being deleted. Right?

PR #17, opened 2015-09-16
Move activities and waypoint_types values in a dedicated file

asaunier:
Because
* some lists may be used in several models/views
* some lists may be long and pollute the models/views code
* it's good to have everything at the same place instead of dispatched in several places. (Should we use a YAML file instead of python?)

By the way:
* would it be relevant to be more consistent when importing modules?
Sometimes we have ``from ..models`` or ``from .models``, sometimes ``from app_api.models``. What's the best practise?
* I don't really like "app_api" (too complicated), would simply using "api" make sense?

### DIFF
--- a/app_api/attributes.py
+++ b/app_api/attributes.py
@@ -1,0 +1,43 @@
+activities = [
+    'skitouring',
+    'snow_ice_mixed',
+    'mountain_climbing',
+    'rock_climbing',
+    'ice_climbing',
+    'hiking',
+    'snowshoeing',
+    'paragliding',
+    'mountain_biking',
+    'via_ferrata'
+]
+
+waypoint_types = [
+    'summit',
+    'pass',
+    'lake',
+    'bisse',
+    'waterfall',
+    'cave',
+    'glacier',
+    'cliff',
+    'waterpoint',
+    'canyon',
+    'valley',
+    'locality',
+    'dam',
+    'webcam',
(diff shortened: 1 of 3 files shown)


tsauerwein:
Looks good.

> would it be relevant to be more consistent when importing modules? Sometimes we have from ..models or from .models, sometimes from app_api.models. What's the best practise?

The advantage of using relative imports is that you can move a package without having to change the imports for each sub-package (see http://explore-flask.readthedocs.org/en/latest/conventions.html#relative-imports). The [Python Style Guide](https://www.python.org/dev/peps/pep-0008/#imports) recommends absolute imports though. I would be fine with that.

> I don't really like "app_api" (too complicated), would simply using "api" make sense?

I don't mind.

asaunier:
Thanks for reviewing. Merging...

> The advantage of using relative imports is that you can move a package without having to change the imports for each sub-package (see http://explore-flask.readthedocs.org/en/latest/conventions.html#relative-imports). The Python Style Guide recommends absolute imports though. I would be fine with that.

I think we should use a consistent approach for all imports. I will probably suggest a PR. I think I prefer the absolute paths because the origin/location is more explicit.

>> I don't really like "app_api" (too complicated), would simply using "api" make sense?
> I don't mind.

OK :) Perhaps I will suggest a PR with "api" only if no one else objects.

PR #20, opened 2015-09-24
Rename app and use absolute module paths

asaunier:
As discussed here: https://github.com/c2corg/v6_api/pull/17#issuecomment-141966225

I think that using "absolute" paths are less confusing.
Concerning the app name, it's not a big deal but I prefer a simpler form :P


gberaudo:
LGTM

elemoine:
If it was me I would have used `c2capi` or `c2corgapi` as there is a higher risk of conflicts with `api`.

asaunier:
c2capi or c2corgapi are not that much simpler than app_api :)
But I admit it would be more explicit and avoid possible conflicts, so why not.

asaunier:
I have renamed the module to c2corg_api instead of simply api. It is longer than app_api but at least a little more explicit.
It adds a new commit but I can rebase to merge all in a single one if it makes sense.

Issue #16, opened 2015-09-16
Auth management


asaunier:
Interesting resources:

* http://blog.octo.com/en/design-a-rest-api/#security

> We recommend securing your API with OAuth2.

> Unlike OAuth1, OAuth2 allows you to manage authentication and resource authorization for any type of application (native mobile app, native tablet app, JavaScript app, server side web app, batch processing…) with or without the resource owner’s consent.
> OAuth2 is the de facto standard for securing APIs. Using another technology would slow down your API development and adoption.

* http://stackoverflow.com/questions/549/the-definitive-guide-to-form-based-website-authentication

Interesting (?) implementations:
* https://github.com/elliotpeele/pyramid_oauth2_provider
* https://github.com/tilgovi/pyramid-oauthlib
* https://github.com/idan/oauthlib
* https://github.com/sneridagh/osiris
* https://github.com/hathawsh/yasso
* http://peterhudec.github.io/authomatic/ - https://github.com/peterhudec/authomatic (but used to authenticate against a third-party oauth provider - not sure it may help us)
* https://github.com/omab/python-social-auth (auth through social networks tools)

gberaudo:
According to the wikipedia page about OAuth, OAuth2 suffers from many
security flaws.
Are there some requirements about the version we need to use?

asaunier:
> Are there some requirements about the version we need to use?

No. Using OAuth is not even required.

tsauerwein:
Best-practice on how to implement a remember-me function: http://stackoverflow.com/questions/244882/what-is-the-best-way-to-implement-remember-me-for-a-website

tsauerwein:
What requirements do we have for the authentication/authorization?

 * We must be able to support multiple clients.
 * It should scale. That is it should be possible to run multiple instances of the REST API.
 * ?

Two more resources that are worth reading:
https://scotch.io/tutorials/the-ins-and-outs-of-token-based-authentication
https://scotch.io/tutorials/the-anatomy-of-a-json-web-token

asaunier:
Something that might be useful/required in the future: be able to sign in using a Facebook/Google/Twitter/... account (ie. a third-party auth service).

asaunier:
> Three more resources that are worth reading:

Thanks @tsauerwein those articles are indeed really interesting. In short they explain what token-based auth brings: stateless/scalability, increased security, mobile app support, etc.

tsauerwein:
[...]

PR #21, opened 2015-09-25
Add user table and class

asaunier:
[Still a looot of work to do]

- [x] Create model for userdata => ~~Question: should we move the user-related tables (userdata, users_groups as in v5? etc.) out of schema "guidebook" to a dedicated "users" (?) DB schema?~~ => done
- [x] use v5 password hashed passwords

Related to #16

### DIFF
--- a/c2corg_api/models/user.py
+++ b/c2corg_api/models/user.py
@@ -1,0 +1,57 @@
+import bcrypt
+
+from sqlalchemy import (
+    Boolean,
+    Column,
+    Integer,
+    String
+    )
+
+from c2corg_api.models import Base, users_schema
+
+
+class User(Base):
+    """
+    Class containing the users' private and authentication data.
+    """
+    __tablename__ = 'user'
+    __table_args__ = {"schema": users_schema}
+
+    id = Column(Integer, primary_key=True)
+    username = Column(String(200), nullable=False, unique=True)
+    email = Column(String(200), nullable=False, unique=True)
+    email_validated = Column(Boolean, nullable=False)
+    _password = Column("password", String(255), nullable=False)
+    temp_password = Column(String(255))
+
+    def _get_password(self):
+        return self._password
+
+    def _set_password(self, password):
+        self._password = self.__encrypt_password(password)
+
+    def set_temp_password(self, password):
+        self.temp_password = self.__encrypt_password(password)
+
+    def __encrypt_password(self, password):
+        return bcrypt.hashpw(password, bcrypt.gensalt())
+
+    def validate_password(self, passwd):
+        """Check the password against existing credentials.
+        this method _MUST_ return a boolean.
+        @param passwd: the password that was provided by the user to
+        try and authenticate. This is the clear text version that we will
+        need to match against the encrypted one in the database.
+        @type password: string
+        """
+        if self._password == bcrypt.hashpw(passwd, self._password):
+            return True
+        if self.temp_password is not None and \
+           self.temp_password != "" and \
+           self.temp_password == bcrypt.hashpw(passwd, self.temp_password):
+            self._password = self.temp_password
+            self.temp_password = None
+            return True
+        return False
+
+    password = property(_get_password, _set_password)
--- a/c2corg_api/tests/models/test_user.py
+++ b/c2corg_api/tests/models/test_user.py
@@ -1,0 +1,17 @@
+from c2corg_api.models.user import User
+
+from c2corg_api.tests import BaseTestCase
+
+
+class Testuser(BaseTestCase):
+
+    def test_validate_password(self):
+        tony = User(
+            username='tonymontana',
+            email='tony@montana.com', password='foobar'
+        )
+
+        self.assertFalse(tony.validate_password('foobare'))
+        self.assertTrue(tony.validate_password('foobar'))
+        tony.set_temp_password('bouchon')
+        self.assertTrue(tony.validate_password('bouchon'))
--- a/c2corg_api/models/__init__.py
+++ b/c2corg_api/models/__init__.py
@@ -8,6 +8,7 @@ from sqlalchemy.orm import (
 from zope.sqlalchemy import ZopeTransactionExtension
 
 schema = 'guidebook'
+users_schema = 'users'
 
 DBSession = scoped_session(sessionmaker(extension=ZopeTransactionExtension()))
 
@@ -24,3 +25,4 @@ from c2corg_api.models import waypoint  # noqa
 from c2corg_api.models import route  # noqa
 from c2corg_api.models import document_history  # noqa
 from c2corg_api.models import image  # noqa
+from c2corg_api.models import user  # noqa
--- a/scripts/create_user_db.sh
+++ b/scripts/create_user_db.sh
@@ -4,6 +4,7 @@ create database c2corg_$USER owner "www-data";
 \c c2corg_$USER
 create extension postgis;
 create schema guidebook authorization "www-data";
+create schema users authorization "www-data";
 \q
 EOF
 .build/venv/bin/initialize_c2corg_api_db development.ini
--- a/scripts/create_user_db_test.sh
+++ b/scripts/create_user_db_test.sh
@@ -8,5 +8,6 @@ create database c2corg_${USER}_tests owner "www-data";
 \c c2corg_${USER}_tests
 create extension postgis;
 create schema guidebook authorization "www-data";
+create schema users authorization "www-data";
 \q
 EOF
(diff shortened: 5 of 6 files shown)


elemoine:
> Question: should we move the user-related tables (userdata, users_groups as in v5? etc.) out of schema "guidebook" to a dedicated "users" (?) DB schema?

A "user" DB schema would make sense to me.

elemoine:
Don't you want to store hashed passwords? As in c2cgeoportal. See https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/models.py#L185.

asaunier:
> Don't you want to store hashed passwords?

Sure. Passwords are already indeed hashed in v5.
Thanks for pointing what exists for c2cgeoportal.

tsauerwein:
See also: http://security.stackexchange.com/questions/17421/how-to-store-salt

asaunier:
I have moved the userdata table to schema "users" (I think the "user" table/document will stay in schema "guidebook").

I have also added the password handling functions from c2cgeoportal. And tests. The encryption used is from c2cgeoportal as well, but not sure we will keep it as is since we must be compatible with the encrypted passwords from the existing database (can we really convert the already encrypted passwords?). See https://github.com/c2corg/camptocamp.org/blob/master/lib/model/doctrine/UserPrivateData.class.php

asaunier:
Using HTTPS scheme for the API URLs is needed (highly recommended for OAuth and that's a wish from the client). I don't know if it is really possible for dev instances since waitress seems to not directly support https.

elemoine:
> Using HTTPS scheme for the API URLs is needed (highly recommended for OAuth and that's a wish from the client). I don't know if it is really possible for dev instances since waitress seems to not directly support https (unless by using a proxy in front of it and setting url_scheme = https in development.ini).

We may want to use mod_wsgi-express instead of waitress. With mod_wsgi-express the mod_wsgi author has done an excellent job on making it easier to use mod_wsgi.

gberaudo:
The PHP function used in v5 seems more secure than what is used in
c2cgeoportal. Why not keep the hashing logics?

asaunier:
Anyway, we'll probably have to stick to the v5 encryption unless we re-encrypt the existing passwords (is it  even possible?).

tsauerwein:
> The PHP function used in v5 seems more secure than what is used in c2cgeoportal. Why not keep the hashing logics?

The [PHP function password_hash](http://php.net/manual/en/function.password-hash.php) uses bcrypt, which is [recommended](http://codahale.com/how-to-safely-store-a-password/). So I would also keep using bcrypt, for example with this [module](https://pypi.python.org/pypi/bcrypt/1.1.0).

> ...unless we re-encrypt the existing passwords (is it even possible?).

No, you can not get the plain-text passwords again. If we *would* want to use a different hashing method, we could do the following: Have two columns "old_password" (with the old password) and "password". Once a user logs in for the first time (at this moment we have the plain text password), we check with the column "old_password", calculate the new password hash and fill the column "password". For the next logins we will only check with the new password hash. 
But I am not saying we should do that, because bcrypt is a good option.

elemoine:
http://codahale.com/how-to-safely-store-a-password/ is a very interesting article. +1 on relying on bcrypt.

asaunier:
> So I would also keep using bcrypt, for example with this module.

Done at ff26661

c2corg_api/models/userdata.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class UserData(Base):
+    """
+    Class containing the users' private and authentication data.
+    """
+    __tablename__ = 'userdata'
+    __table_args__ = {"schema": users_schema}
+    __mapper_args__ = {
+        'polymorphic_identity': 'ud'
```
gberaudo: Do we need it to be polymorphic?

c2corg_api/models/userdata.py
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    document_id = Column(Integer, primary_key=True)
```
gberaudo: Is it a document?

c2corg_api/models/userdata.py
asaunier: Not sure. But in c2cgeoportal we use it even for models with no inheritance:
https://github.com/camptocamp/c2cgeoportal/blob/master/c2cgeoportal/models.py#L196

c2corg_api/models/userdata.py
asaunier: No. It's different from the User documents.
But the user id should be related to the User document id.

By the way I'm a bit annoyed: that would make sense to rename this model to User but it would conflict with the future User document model. Or maybe we could name the user document as Profile for instance?

asaunier: Perhaps we should rename this field to "id" and add a FK towards the User (or Profile?) document_id

tsauerwein: > Perhaps we should rename this field to "id" and add a FK towards the User (or Profile?) document_id

I would make a field `id` as PK and a field `document_id` as FK pointing to the profile (which is a document).

gberaudo:
I renamed userdata to user and rebased on master.
I also changed the PK to `id`. We will add later the FK to the profile `document_id`.
I think we can merge it. Then in a subsequent PR I will add a view for creating a user.

c2corg_api/models/user.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    __tablename__ = 'user'
+    __table_args__ = {"schema": users_schema}
+
+    id = Column(Integer, primary_key=True)
+    username = Column(String(200), nullable=False, unique=True)
+    email = Column(String(200), nullable=False, unique=True)
+    email_validated = Column(Boolean, nullable=False)
+    _password = Column("password", String(255), nullable=False)
+    temp_password = Column(String(255))
```
tsauerwein: What is `temp_password`?

c2corg_api/models/user.py
```diff
@@ -0,0 +1,57 @@
...
+    Class containing the users' private and authentication data.
+    """
+    __tablename__ = 'user'
+    __table_args__ = {"schema": users_schema}
+
+    id = Column(Integer, primary_key=True)
+    username = Column(String(200), nullable=False, unique=True)
+    email = Column(String(200), nullable=False, unique=True)
+    email_validated = Column(Boolean, nullable=False)
+    _password = Column("password", String(255), nullable=False)
```
tsauerwein: Why is there an underscore?

c2corg_api/models/user.py
asaunier: It's used when the user asks for a new password ("forgot your password").

c2corg_api/tests/models/test_user.py
```diff
@@ -0,0 +1,17 @@
+from c2corg_api.models.user import User
+
+from c2corg_api.tests import BaseTestCase
+
+
+class Testuser(BaseTestCase):
+
+    def test_to_archive(self):
```
tsauerwein: The name `test_to_archive` does not seem right.

c2corg_api/models/user.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def __encrypt_password(self, password):
+        return bcrypt.hashpw(password, bcrypt.gensalt())
+
+    def validate_password(self, passwd):
+        """Check the password against existing credentials.
+        this method _MUST_ return a boolean.
+        @param passwd: the password that was provided by the user to
+        try and authenticate. This is the clear text version that we will
+        need to match against the (possibly) encrypted one in the database.
```
tsauerwein: Why "possibly" encrypted?

c2corg_api/models/user.py
tsauerwein: Sorry, I made the comment before seeing the property definition.

asaunier: Because it is the encrypted version of the password. It is set using ``_set_password``:
https://github.com/c2corg/v6_api/blob/oauth/c2corg_api/models/user.py#L30-L31

tsauerwein:
Looks good to me.

c2corg_api/models/user.py
asaunier: I don't know, this code comes from GeoMapFish. I think we can remove "(possibly)".

c2corg_api/tests/models/test_user.py
asaunier: Indeed. What about ``test_validate_password``?

asaunier:
Looks good to me too (except the test function name).

asaunier:
I have made the 2 minor changes mentioned previously (renamed test function and remove "possibly" in the validate password function comment).

I'm waiting for the Travis approval before merging the PR.

gberaudo:
Rebased, waiting for Travis.

asaunier:
Thanks, Travis passes. Merging...

PR #22, opened 2015-09-25
Add lint target to Makefile

gberaudo:
Currently a simple alias to flake8.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,10 +25,10 @@ help:
 	@echo
 	@echo "Secondary targets:"
 	@echo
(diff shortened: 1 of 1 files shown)


tsauerwein:
Could you also please update the docs at the top of the Makefile?

```
 "- flake8		Run flake8 checker on the Python code"
```

gberaudo:
Can I get ride of the flake8 target? (rename flake8 to lint instead of an alias)

tsauerwein:
`flake8` is a target that we are also using in other Python projects. But I have no objections renaming it.

PR #23, opened 2015-09-25
Add image document

gberaudo:


gberaudo:
It is basically a copy of route model, view and tests, with small adaptations of the tests to better leverage on the test frameworks:
- use integrated instanceof test;
- use json capabilities of webtest.

I am wondering if copying the tests is the right approach.

Shouldn't we introduce helper functions to simplify testing?
For example we could use reflection (inspect?) to simplify asserts like:
```
        self.assertEqual(body.get('document_id'), self.image.document_id)
        self.assertEqual(
            body.get('activities'), self.image.activities)
```

When more attributes are added to the models, are we going to test if each of them is correctly
present in the response json? If yes I think using inspect would simplify greatly the task. If no, where
should we stop?

Another improvement could be to introduce a base class for functional tests where to put
base methods to factorize the checks of the attributes common to all classes of documents/locales. For example, all locales have a title, a description and an id.

What do you think about these ideas? Have you other thoughts?

asaunier:
I'am +1 for merging commit e01b3c2
Perhaps you could also add the other images attributes from the v5 DB?

For the 2 other commits, what do you think @tsauerwein ?
Perhaps you should move them in a dedicated PR and make the changes apply to the other document types? Or even in a PR that factorizes stuff in views/models/tests of the various doc types?

tsauerwein:
> Perhaps you should move them in a dedicated PR and make the changes apply to the other document types? Or even in a PR that factorizes stuff in views/models/tests of the various doc types?

Sounds good.

gberaudo:
Sorry @tsauerwein, did not see your comment about absolute imports.
That is weird I did not received an email notification for it.
I will fix the import in a future commit.

PR #24, opened 2015-09-29
Introduce base class for testing views

gberaudo:
Introduce a base class for testing views and factorize basics test methods.

Used by:
- [X] images;
- [x] routes;
- [x] waypoints

PR #26, opened 2015-10-01
Add support for versioning

tsauerwein:
- [x] Document how it works.
- [x] Activate/implement the versioning for routes and images (refactor the views at the same time)
- [x] Base branch on #24 
- [ ] ~~Add error handler so that JSON is returned in case of an error.~~ -> #31

### DIFF
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -89,7 +89,11 @@ class _DocumentLocaleMixin(object):
             Integer, ForeignKey(schema + '.documents.document_id'),
             nullable=False)
 
-    culture = Column(String(2), nullable=False)  # TODO as fk
+    @declared_attr
+    def culture(self):
+        return Column(
+            String(2), ForeignKey(schema + '.cultures.culture'),
+            nullable=False)
 
     title = Column(String(150), nullable=False)
     description = Column(String)
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -21,6 +21,13 @@ quality_types = [
     ]
 
 
+class Culture(Base):
+    """The supported languages.
+    """
+    __tablename__ = 'cultures'
+    culture = Column(String(2), primary_key=True)
+
+
 class _DocumentMixin(object):
     """
     Contains the attributes that are common for `Document` and
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -11,6 +11,7 @@ from sqlalchemy.orm import relationship
 from colander import MappingSchema, SchemaNode, String as ColanderString
 import abc
 import uuid
+import enum
 
 from c2corg_api.models import Base, schema
 from utils import copy_attributes
@@ -22,6 +23,8 @@ quality_types = [
     'good',
     'excellent'
     ]
+
+UpdateType = enum.Enum('UpdateType', 'FIGURES_ONLY LANG_ONLY ALL NONE')
 
 
 class Culture(Base):
@@ -93,7 +96,57 @@ class Document(Base, _DocumentMixin):
         return [locale.to_archive() for locale in self.locales]
 
     def update(self, other):
+        """Copies the attributes from `other` to this document.
+        Also updates all locales.
+        """
         copy_attributes(other, self, Document._ATTRIBUTES_WHITELISTED)
+
+        for locale_in in other.locales:
+            locale = self.get_locale(locale_in.culture)
+            if locale:
+                locale.update(locale_in)
+                locale.document_id = self.document_id
+            else:
+                self.locales.append(locale_in)
+
+    def get_versions(self):
+        """Get the version hashs of this document and of all its locales.
+        """
+        return {
+            'document': self.version,
+            'locales': {
+                locale.culture: locale.version for locale in self.locales
+            }
+        }
+
+    def get_update_type(self, old_versions):
+        """Get the update type (only figures have changed, only locales have
+        changed, both have changed or nothing) and the languages that have
+        changed.
+        This is done by comparing the old version hashs (before flushing to
+        the database) with the current hashs. Because SQLAlchemy automatically
+        changes the hash, when something has changed, we can easily detect
+        what has changed.
+        """
+        figures_equal = self.version == old_versions['document']
+
+        changed_langs = []
+        locale_versions = old_versions['locales']
+        for locale in self.locales:
+            locale_version = locale_versions.get(locale.culture)
+
+            if not (locale_version and locale_version == locale.version):
+                # new locale or locale has changed
+                changed_langs.append(locale.culture)
+
+        if not figures_equal and changed_langs:
+            return (UpdateType.ALL, changed_langs)
+        elif not figures_equal:
+            return (UpdateType.FIGURES_ONLY, [])
+        elif changed_langs:
+            return (UpdateType.LANG_ONLY, changed_langs)
+        else:
+            return (UpdateType.NONE, [])
 
     def get_locale(self, culture):
         """Get the locale with the given culture or `None` if no locale
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -68,8 +68,11 @@ class Document(Base, _DocumentMixin):
             'version_id_generator': lambda version: uuid.uuid4().hex
     }
 
+    _ATTRIBUTES_WHITELISTED = \
+        ['document_id', 'version']
+
     _ATTRIBUTES = \
-        ['document_id', 'version', 'protected', 'redirects_to', 'quality']
+        _ATTRIBUTES_WHITELISTED + ['protected', 'redirects_to', 'quality']
 
     @abc.abstractmethod
     def to_archive(self):
@@ -87,6 +90,19 @@ class Document(Base, _DocumentMixin):
 
     def get_archive_locales(self):
         return [locale.to_archive() for locale in self.locales]
+
+    def update(self, other):
+        copy_attributes(other, self, Document._ATTRIBUTES_WHITELISTED)
+
+    def get_locale(self, culture):
+        """Get the locale with the given culture or `None` if no locale
+        is present.
+        """
+        l = [locale for locale in self.locales if locale.culture == culture]
+        if l:
+            return l[0]
+        else:
+            return None
 
 
 class ArchiveDocument(Base, _DocumentMixin):
@@ -140,6 +156,9 @@ class DocumentLocale(Base, _DocumentLocaleMixin):
     def to_archive(self, locale):
         copy_attributes(self, locale, DocumentLocale._ATTRIBUTES)
         return locale
+
+    def update(self, other):
+        copy_attributes(other, self, DocumentLocale._ATTRIBUTES)
 
 
 class ArchiveDocumentLocale(Base, _DocumentLocaleMixin):
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -8,6 +8,7 @@ from sqlalchemy import (
     )
 from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.orm import relationship
+from colander import MappingSchema, SchemaNode, String as ColanderString
 import abc
 import uuid
 
@@ -168,3 +169,14 @@ class ArchiveDocumentLocale(Base, _DocumentLocaleMixin):
         'polymorphic_identity': 'd',
         'polymorphic_on': _DocumentLocaleMixin.type
     }
+
+
+def get_update_schema(document_schema):
+    """Create a Colander schema for the update view which contains an update
+    message and the document.
+    """
+    class UpdateSchema(MappingSchema):
+        message = SchemaNode(ColanderString(), missing='')
+        document = document_schema.clone()
+
+    return UpdateSchema()
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -81,7 +81,7 @@ class Document(Base, _DocumentMixin):
         """
         return
 
-    def to_archive(self, doc):
+    def _to_archive(self, doc):
         """Copy the attributes of this document into a passed in
         `Archive*` instance.
         """
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -9,6 +9,7 @@ from sqlalchemy import (
 from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.orm import relationship
 import abc
+import uuid
 
 from c2corg_api.models import Base, schema
 from utils import copy_attributes
@@ -34,6 +35,7 @@ class _DocumentMixin(object):
     Contains the attributes that are common for `Document` and
     `ArchiveDocument`.
     """
+    version = Column(String(32), nullable=False)
     # move to metadata?
     protected = Column(Boolean)
     redirects_to = Column(Integer)
@@ -61,7 +63,13 @@ class Document(Base, _DocumentMixin):
     # TODO constraint that there is at least one locale
     locales = relationship('DocumentLocale')
 
-    _ATTRIBUTES = ['document_id', 'protected', 'redirects_to', 'quality']
+    __mapper_args__ = {
+            'version_id_col': _DocumentMixin.version,
+            'version_id_generator': lambda version: uuid.uuid4().hex
+    }
+
+    _ATTRIBUTES = \
+        ['document_id', 'version', 'protected', 'redirects_to', 'quality']
 
     @abc.abstractmethod
     def to_archive(self):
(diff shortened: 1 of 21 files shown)


c2corg_api/models/document.py
```diff
@@ -53,15 +67,101 @@ class (e.g. waypoint, route, ...) a separate table will be created, which
...
+        version = self.version
+
+        locale_versions = {}
+        for locale in self.locales:
+            locale_versions[locale.culture] = locale.version
+
+        return {
+            'document': version,
+            'locales': locale_versions
+        }
```
elemoine: Would the following work?

```py
    def get_versions(self):
        return {
            'document': self.version,
            'locales': {locale.culture: locale.version for locale in self.locales}
        }
```

tsauerwein: Yep!

c2corg_api/models/document.py
```diff
@@ -53,15 +67,101 @@ class (e.g. waypoint, route, ...) a separate table will be created, which
...
+
+    def get_locale(self, culture):
+        """Get the locale with the given culture or `None` if no locale
+        is present.
+        """
+        l = [locale for locale in self.locales if locale.culture == culture]
+        if l:
+            return l[0]
+        else:
+            return None
```
elemoine: Do you like `itertools`? ;)

```py
    from itertools import ifilter

    def get_locale(self, culture):
        return next(ifilter(lambda: locale.culture == culture, self.locales), None)
```

Feel free to disregard if you don't like it :)

elemoine: The advantage of using a generator is saving memory by not creating a list.

c2corg_api/views/waypoint.py
```diff
@@ -34,9 +33,96 @@ def get(self):
...
+        waypoint_in = \
+            schema_waypoint.objectify(self.request.validated['document'])
+        self._check_document_id(id, waypoint_in.document_id)
+
+        waypoint = self._get_waypoint(id)
+        self._check_versions(waypoint, waypoint_in)
+        old_versions = waypoint.get_versions()
+        waypoint.update(waypoint_in)
+
+        DBSession.merge(waypoint)
```
elemoine: Is the `merge` really required? In your case you know you have a valid waypoint object, so can't you just update it and call `flush`?

c2corg_api/views/waypoint.py
```diff
@@ -34,9 +33,96 @@ def get(self):
...
+            waypoint = DBSession. \
+                query(Waypoint). \
+                join(Waypoint.locales). \
+                filter(Waypoint.document_id == id). \
+                options(contains_eager(Waypoint.locales)). \
+                filter(DocumentLocale.culture == culture). \
+                first()
+
+        if not waypoint:
+            raise HTTPNotFound('document not found')
```
elemoine: To make the code more readable in `put` I'd leave the responsibility to throw the HTTP exception to the `put` method.

elemoine: I see that you do the same in `_check_document_id`, so forget about this comment maybe.

c2corg_api/views/waypoint.py
```diff
@@ -34,9 +33,96 @@ def get(self):
...
+            schema_waypoint.objectify(self.request.validated['document'])
+        self._check_document_id(id, waypoint_in.document_id)
+
+        waypoint = self._get_waypoint(id)
+        self._check_versions(waypoint, waypoint_in)
+        old_versions = waypoint.get_versions()
+        waypoint.update(waypoint_in)
+
+        DBSession.merge(waypoint)
+        DBSession.flush()
```
elemoine: So the flush operation will cause an SQLAlchemy `StaleDataError` if the versions don't match. Don't you want to convert this to an `HTTPConflict` error?

c2corg_api/views/document.py
```diff
@@ -30,3 +31,64 @@ def _create_new_version(self, document):
...
+                culture=locale.culture,
+                version=999,
+                document_archive=archive,
+                document_locales_archive=locale_archive,
+                history_metadata=meta_data
+            )
+            locale_versions.append(version)
+
+        DBSession.add(archive)
+#         DBSession.add_all(archive_locales)
```
elemoine: Why is this line commented out?

c2corg_api/views/document.py
```diff
@@ -30,3 +31,64 @@ def _create_new_version(self, document):
...
+        DBSession.add(meta_data)
+        DBSession.add_all(locale_versions)
+        DBSession.flush()
```
elemoine: You probably can remove that `flush`?

c2corg_api/views/document.py
```diff
@@ -30,3 +31,64 @@ def _create_new_version(self, document):
...
+            self._get_cultures_to_update(document, update_type, changed_langs)
+        locale_versions = []
+        for culture in cultures:
+            locale = document.get_locale(culture)
+            locale_archive = self._get_locale_archive(locale, changed_langs)
+
+            version = DocumentVersion(
+                document_id=document.document_id,
+                culture=locale.culture,
+                version=999,
```
elemoine: What's this magic number? This at least deserves a comment, no?

c2corg_api/models/document.py
```diff
@@ -97,12 +202,38 @@ def document_id(self):
 class DocumentLocale(Base, _DocumentLocaleMixin):
     __tablename__ = 'documents_locales'
 
-    _ATTRIBUTES = ['document_id', 'culture', 'title', 'description']
+    __mapper_args__ = {
+        'polymorphic_identity': 'd',
+        'polymorphic_on': _DocumentLocaleMixin.type,
+        'version_id_col': _DocumentLocaleMixin.version,
+        'version_id_generator': lambda version: uuid.uuid4().hex
```
elemoine: I was wondering about the use of `uuid4` versus `uuid1`. Using `uuid4` should be safe. See http://stackoverflow.com/questions/703035/when-are-you-truly-forced-to-use-uuid-as-part-of-the-design/786541#786541 for an interesting SO answer on that subject.

c2corg_api/models/document.py
```diff
@@ -53,15 +67,101 @@ class (e.g. waypoint, route, ...) a separate table will be created, which
...
+        changed_langs = []
+        locale_versions = old_versions['locales']
+        for locale in self.locales:
+            locale_version = locale_versions.get(locale.culture)
+
+            if not (locale_version and locale_version == locale.version):
+                # new locale or locale has changed
+                changed_langs.append(locale.culture)
+
+        if not figures_equal and changed_langs:
```
elemoine: I'd use `len(changed_langs) > 0` for clarity reasons…

elemoine:
I am done reviewing. I agree with the general approach, based on the use of SQLAlchemy's [versioning stuff](http://docs.sqlalchemy.org/en/latest/orm/versioning.html) to deal with concurrent updates and know what has changed. Very clever.

asaunier:
What is the correct way to push a new version of a document?

I have created a WP using:
```
curl -X POST -v \
-H "Content-Type: application/json" \
-d '{"waypoint_type": "summit", "elevation": 3779, "locales": [{"culture": "en", "title": "Mont Pourri"}]}' \
http://foobar/waypoints
```
I have then tried to update the waypoint but have not figured out how. The best I could do was to PUT on waypoints/1 but I got:
```
{
    "status": "error",
    "errors": [
        {
            "location": "body",
            "name": "document",
            "description": "document is missing"
        }
    ]
}
```

By the way: I have realized that if one adds a ``"document_id":1234`` attributes when creating the document, the created document will received the passed document_id value. Isn't it a problem that anyone may set whatever id to a document?

c2corg_api/models/document.py
tsauerwein: So instead of creating a list you are creating an iterator. In this case `self.locales` contains at max 7 entries and the list `l` should contains at max 1 entry. So, memory is not really an argument.

But the iterator version is more concise, I will change it.

elemoine: > So, memory is not really an argument.

Yes yes I know. I did not mean to use this as an argument.

c2corg_api/views/waypoint.py
tsauerwein: Right, `merge` is not required.

c2corg_api/views/waypoint.py
tsauerwein: `_get_waypoint` is also used in `WaypointRest.get` and I didn't want to repeat the `...raise ...` part.

c2corg_api/views/waypoint.py
tsauerwein: Done.

c2corg_api/views/document.py
tsauerwein: Cleaned up.

c2corg_api/views/document.py
tsauerwein: This makes the tests fail, because the changes are not visible. Though I would have expected that an implicit commit is done?!

c2corg_api/views/document.py
tsauerwein: This is actually still a *TODO*. I am not sure how to increment the version in a safer manner. Maybe just drop the version column and calculate a number (like in this [answer](http://dba.stackexchange.com/a/47794))? Or use a version hash?

c2corg_api/models/document.py
tsauerwein: Thanks for the link!

c2corg_api/models/document.py
tsauerwein: That wouldn't be very *pythonic*! :)

https://google-styleguide.googlecode.com/svn/trunk/pyguide.html?showone=True/False_evaluations#True/False_evaluations

tsauerwein:
> What is the correct way to push a new version of a document?

You can take a look at the tests:
https://github.com/c2corg/v6_api/blob/versioning-modify/c2corg_api/tests/views/test_waypoint.py#L261-L279

We'll have to document the API somewhere.

> By the way: I have realized that if one adds a "document_id":1234 attributes when creating the document, the created document will received the passed document_id value. Isn't it a problem that anyone may set whatever id to a document?

Good catch!

asaunier:
> We'll have to document the API somewhere.

What about the README (at least for a few examples)? See #27

tsauerwein:
For some examples, yes. But I was thinking of a more detailed documentation.

c2corg_api/views/document.py
asaunier: I have made a few tests in my instance:
```
# select * from guidebook.documents_versions;
 id | document_id | culture | version | document_archive_id | document_locales_archive_id | history_metadata_id 
----+-------------+---------+---------+---------------------+-----------------------------+---------------------
  1 |           1 | fr      |       1 |                   1 |                           1 |                   1
  2 |           3 | fr      |       1 |                   2 |                           2 |                   2
  3 |           1 | fr      |     999 |                   3 |                           3 |                   3
```
Do we really need this "version" field?
In v5, it is used to display the version number in the history list:
http://www.camptocamp.org/routes/history/53806/fr
or to access an old version:
http://www.camptocamp.org/routes/53806/fr/14
or to diff 2 versions:
http://www.camptocamp.org/routes/diff/53806/fr/16/14

In wikipedia there is no version number displayed in the history:
https://fr.wikipedia.org/w/index.php?title=GitHub&action=history
An older version is accessed a kind of id:
https://fr.wikipedia.org/w/index.php?title=GitHub&oldid=116321058

@gberaudo suggests to simply use the table's id field (since auto-incremented and unique).

tsauerwein: We were also discussing this. I would be in favor of removing the column.

The requirements demand that the old URLs are kept in place. I don't know if this is also important for these URLs. If not we still could write are wrapper that works with the old version numbers.

c2corg_api/models/document.py
```diff
@@ -20,12 +24,22 @@
...
+    __tablename__ = 'cultures'
+    culture = Column(String(2), primary_key=True)
+
 
 class _DocumentMixin(object):
     """
     Contains the attributes that are common for `Document` and
     `ArchiveDocument`.
     """
+    version = Column(String(32), nullable=False)
```
gberaudo: Is the version hash needed in archives?

c2corg_api/models/document.py
```diff
@@ -53,15 +67,101 @@ class (e.g. waypoint, route, ...) a separate table will be created, which
     # TODO constraint that there is at least one locale
     locales = relationship('DocumentLocale')
 
-    _ATTRIBUTES = ['document_id', 'protected', 'redirects_to', 'quality']
+    __mapper_args__ = {
+            'version_id_col': _DocumentMixin.version,
+            'version_id_generator': lambda version: uuid.uuid4().hex
```
gberaudo: Shouldn't we use a smaller number?
An uuid4 is billions of billions of billions of possible values: 16^32 = 340282366920938463463374607431768211456.

asaunier:
I have played a bit with the branch in my instance. It looks excellent :)
Conflicts (pushing changes from an old version) are actually detected.

Please note that reverting to an older version may be useful, for instance to cancel a version (vandalism, incorrect change).

c2corg_api/views/document.py
asaunier: > The requirements demand that the old URLs are kept in place.

I think that's mostly important for the documents main URL, for instance http://www.camptocamp.org/outings/681428/fr/le-luisin-de-vallorcine-a-salanfe

Ditto for documents lists (with or without parameters): http://www.camptocamp.org/outings/list/areas/14421/orderby/date/order/desc

For secondary pages such as history, diff, old version, etc. if it's not such a big deal if the URLs change.

c2corg_api/models/document.py
```diff
@@ -53,15 +67,101 @@ class (e.g. waypoint, route, ...) a separate table will be created, which
...
+        return {
+            'document': version,
+            'locales': locale_versions
+        }
+
+    def get_update_type(self, old_versions):
+        """Get the update type (only figures have changed, only locales have
+        changed, both have changed or nothing) and the languages that have
+        changed.
+        This is done by comparing the old version hashs (before flushing to
```
gberaudo: In the tests below, `get_update_type` is called after the flush.
Is the `merge` operation where SQLAlchemy will update the version?

c2corg_api/models/document.py
tsauerwein: See also https://github.com/c2corg/v6_api/pull/26#discussion-diff-40932824

c2corg_api/models/document.py
tsauerwein: In the `flush`, yes.

c2corg_api/models/document.py
elemoine: Ok. It's just me then :)

gberaudo:
@tsauerwein, out of curiosity, why did you choose UUIDs instead of [simple counters](http://docs.sqlalchemy.org/en/latest/orm/versioning.html#simple-version-counting)?

tsauerwein:
> @tsauerwein, out of curiosity, why did you choose UUIDs instead of simple counters?

I wanted to make it very obvious that this version is not the same as the version number (that you see in the UI in v5). But now that we got rid of that version number, we could also use integer versions for Document/DocumentLocale.

gberaudo:
Not sure to understand, doesn't the v5 use a counter? Couldn't we have simply continued to use it in an SqlAlchemy way?

tsauerwein:
> Not sure to understand, doesn't the v5 use a counter? Couldn't we have simply continued to use it in an SqlAlchemy way?

We are talking about two different version fields. In v5 there was a version field on (what is now called) `DocumentVersion`. In v6 there are version fields on `Document` and `DocumentLocale`. To make it obvious that these versions are not the same thing, I originally used a hash for the fields on `Document` and `DocumentLocale`. But now that we got rid of the version on `DocumentVersion`, I made the version fields on `Document` and `DocumentLocale` simple integer counters (see https://github.com/c2corg/v6_api/commit/c18d8ad47b101821cba5b943d89dbf287446e7c6).

tsauerwein:
I added the documentation. Not to block further work, I would suggest to merge this PR (in case there are no more comments).

asaunier:
OK it makes sense to me. Is this PR ready for merging?

tsauerwein:
From my side yes.

asaunier:
>  I would suggest to merge this PR (in case there are no more comments).

I'm currently looking at the doc but +1 for merging.
Thanks for the great work.

asaunier:
What about renaming https://github.com/c2corg/v6_api/blob/versioning-modify/c2corg_api/models/versioning-explained.md to README.md?

This way it will appear directly in the folder when browsing the code in github. That would make it more obvious.

tsauerwein:
> What about renaming https://github.com/c2corg/v6_api/blob/versioning-modify/c2corg_api/models/versioning-explained.md to README.md?

Good idea!

gberaudo:
I added two comments: one about typo and one about adding some details about two tables.
+1 for merging when this is addressed.

tsauerwein:
Thanks for the comments, merging.

PR #28, opened 2015-10-05
Pagination

### LINKED ISSUE #18: Pagination
As for now the collection_get() method returns the whole list of documents, which is not realistic since we have too many items.

Applying a default limit and using pagination technics is then required.

By the way it seems the usual LIMIT/OFFSET approach [1]  might cause performance issues when the offset increases (because postgresql has to deal with more rows before slicing them).

Some resources on the web recommend to use the "seek method":
http://leopard.in.ua/2014/10/11/postgresql-paginattion/
http://chrisdone.com/posts/postgresql-pagination
http://winand.at/slides/PaginationDoneTheRightWay.pdf
that does not use OFFSET but a standard WHERE clause combined to LIMIT.

A drawback is that you need to know the id of the last item of the current page to get the x items whose ids are next. It seems it is especially recommended for "infinite scrolling" requests or for "next page" links.

[1] With SQLAlchemy: most likely done using slice() http://docs.sqlalchemy.org/en/rel_1_0/orm/query.html#sqlalchemy.orm.query.Query.slice

### LINKED ISSUE #49: Do not return JSON array as top-level element
`/waypoints` currently returns an array which is a security problem: http://flask.pocoo.org/docs/0.10/security/#json-security

Though new browsers are not affected (see http://stackoverflow.com/a/16880162/119937), it would be safer simply not to return an array.

gberaudo:
Allow pagination either using an `offset` or `after` a provided document id.
TODO:
- [x] rebase when @tsauerwein work is merged in;
- [x] merge pagination code (not query builder commits)
- [x] create WIP PR with query builder commits (the branch is here: https://github.com/c2corg/v6_api/tree/gberaudo_query_builder)

Closes #49
Closes #18

### DIFF
--- a/c2corg_api/views/document.py
+++ b/c2corg_api/views/document.py
@@ -18,11 +18,10 @@ class DocumentRest(object):
         self.request = request
 
     def _paginate_offset(self, clazz, schema, adapt_schema):
-        offset = self.request.GET.get('offset')
-        limit = self.request.GET.get('limit')
-        total = self.request.GET.get('total')
-        offset = 0 if offset is None else int(offset)
-        limit = 30 if limit is None else int(limit)
+        validated = self.request.validated
+        offset = validated['offset'] if 'offset' in validated else 0
+        limit = validated['limit'] if 'limit' in validated else 30
+        total = validated['total'] if 'total' in validated else None
 
         base_query = DBSession. \
             query(clazz)
@@ -34,7 +33,7 @@ class DocumentRest(object):
             all()
         set_available_cultures(documents)
 
-        total = base_query.count() if total is None else int(total)
+        total = base_query.count() if total is None else total
 
         return {
             'documents': [
@@ -47,8 +46,8 @@ class DocumentRest(object):
         }
 
     def _paginate_after(self, clazz, schema, adapt_schema):
-        after = self.request.GET.get('after')
-        limit = self.request.GET.get('limit')
+        after = self.request.validated['after']
+        limit = self.request.validated['limit']
         after_id = int(after)
         limit = 30 if limit is None else int(limit)
 
@@ -74,7 +73,7 @@ class DocumentRest(object):
         }
 
     def _paginate(self, clazz, schema, adapt_schema):
-        if self.request.GET.get('after') is not None:
+        if 'after' in self.request.validated:
             return self._paginate_after(clazz, schema, adapt_schema)
         else:
             return self._paginate_offset(clazz, schema, adapt_schema)
--- a/c2corg_api/views/document.py
+++ b/c2corg_api/views/document.py
@@ -28,7 +28,7 @@ class DocumentRest(object):
 
         documents = base_query. \
             options(joinedload(getattr(clazz, 'locales'))). \
-            order_by(clazz.document_id). \
+            order_by(clazz.document_id.desc()). \
             slice(offset, offset + limit). \
             all()
         set_available_cultures(documents)
@@ -56,8 +56,8 @@ class DocumentRest(object):
 
         documents = base_query. \
             options(joinedload(getattr(clazz, 'locales'))). \
-            order_by(clazz.document_id). \
-            filter(clazz.document_id > after_id). \
+            order_by(clazz.document_id.desc()). \
+            filter(clazz.document_id < after_id). \
             limit(limit). \
             all()
         set_available_cultures(documents)
--- a/c2corg_api/views/document.py
+++ b/c2corg_api/views/document.py
@@ -12,52 +12,43 @@ from c2corg_api.views.validation import check_required_fields, \
     check_duplicate_locales
 
 
+# the maximum number of documents that can be returned in a request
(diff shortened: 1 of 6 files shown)


c2corg_api/views/document.py
```diff
@@ -7,18 +13,115 @@ class DocumentRest(object):
...
+        limit = self.request.GET.get('limit')
+        offset = 0 if offset is None else int(offset)
+        limit = 30 if limit is None else int(limit)
+
+        base_query = DBSession. \
+            query(clazz)
+
+        documents = base_query. \
+            options(joinedload(getattr(clazz, 'locales'))). \
+            order_by(clazz.document_id). \
```
gberaudo: By default in v5, the ordering is newest documents first.

c2corg_api/views/document.py
```diff
@@ -7,18 +13,115 @@ class DocumentRest(object):
...
+        after_id = int(after)
+        limit = 30 if limit is None else int(limit)
+
+        base_query = DBSession. \
+            query(clazz)
+
+        documents = base_query. \
+            options(joinedload(getattr(clazz, 'locales'))). \
+            order_by(clazz.document_id). \
+            filter(clazz.document_id > after_id). \
```
gberaudo: By default in v5, the ordering is newest documents first.

c2corg_api/views/document.py
```diff
@@ -7,18 +13,115 @@ class DocumentRest(object):
...
+
+    def _paginate(self, clazz, schema):
+        if self.request.GET.get('after') is not None:
+            return self._paginate_after(clazz, schema)
+        else:
+            return self._paginate_offset(clazz, schema)
+
+    def _collection_get(self, clazz, schema):
+        res = self._paginate(clazz, schema)
+        return res['documents']
```
gberaudo: The total should also be returned to the client.

asaunier: Shouldn't we make a separated request for counting?
I mean it is not really needed to count the total number of results for every page of results returned by the API.

gberaudo: Right, computing total count only if not provided in request and return it to client.
The returned json is now:
```
{
  'documents': the_list,
  'total': total count
}
```

asaunier:
As discussed, taking filtering criteria into account in the paginate functions will be needed.

gberaudo:
As we saw together, here is how the transformation from the REST query string to the ORM is done in v5:
https://github.com/c2corg/camptocamp.org/blob/master/lib/model/doctrine/generated/BaseDocument.class.php#L214

asaunier:
051f98b is not really related to pagination. It would have made sense to make a dedicated PR for it.

c2corg_api/views/document.py
gberaudo: done

asaunier:
Is there still a lot of work to do before merging this PR?

c2corg_api/views/document.py
```diff
@@ -17,20 +17,69 @@ class DocumentRest(object):
...
+            query(clazz)
+
+        documents = base_query. \
+            options(joinedload(getattr(clazz, 'locales'))). \
+            order_by(clazz.document_id.desc()). \
+            slice(offset, offset + limit). \
+            all()
+        set_available_cultures(documents)
+
+        total = base_query.count() if total is None else total
```
tsauerwein: @gberaudo Can you explain me the idea behind this `count` property? You can pass the count as parameter which is then returned again?

gberaudo: If the client code does not want the count to be computed (again), it is
possible to pass it in the request.
The reply stays the same.

tsauerwein: But how can you be sure that no document was added since the last request? The client does not know.

Instead we could cache the query on the server-side. What do you think?

gberaudo: The technique of passing the count was taken from
http://sluggo.scrapping.cc/python/WebHelpers/modules/paginate.html
It looked like a standard pyramid mechanism so I implemented it.
It is OK for me to simply drop the feature and come back to it later to
test with real data if it has a noticable performance impact.

tsauerwein:
I rebased and reviewed this PR. @gberaudo, do you want to take a quick look before merging?

gberaudo:
Yes @tsauerwein, looking at it.

gberaudo:
I added some comments, if you want @tsauerwein, I can do these small changes.

tsauerwein:
Thanks for your comments, merging.

asaunier:
Pagination works only for waypoints as for now, right?

It seems the code is not the same for routes and in my instance
http://c2corgv6-demo.gis.internal:6545/waypoints?limit=1 indeed returns only one item whereas http://c2corgv6-demo.gis.internal:6545/routes?limit=1 returns 30 or so routes...

tsauerwein:
> Pagination works only for waypoints as for now, right?

Oh yes, you are right. I will add it for routes and images.

PR #30, opened 2015-10-06
Add db creation scripts

gberaudo:
Extracted the README improvement from #28.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,8 @@ install:
 - make -f config/travis install
 
 before_script:
-- psql -f scripts/travis-db.sql -U postgres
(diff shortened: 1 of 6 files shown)


asaunier:
Looks good to me.
See also @tsauerwein's comments https://github.com/c2corg/v6_api/pull/28#commitcomment-13593708
but could perhaps be done later if you prefer?

asaunier:
> See also @tsauerwein's comments #28 (comment):

> In scripts/ we now have two shell script and one SQL script for Travis. Could we at least combine the Travis script and create_user_db_test.sh?

gberaudo:
Thank you for the review.

PR #32, opened 2015-10-06
Fix sql scripts

gberaudo:


asaunier:
Won't those scripts always create the same databases for a given user (eg. "c2corg_alex")?
Then it is not possible to create the DB for the main instance, since they should not be created according to the username of the person that runs the scripts.

What about adding an argument to the scripts? Something like:
```
scripts/create_user_db.sh alex
scripts/create_user_db.sh main
```

gberaudo:
It is already possible, with the following syntax:
```
USER=alex scripts/create_user_db.sh
USER=main scripts/create_user_db.sh
```

tsauerwein:
Could you please mention this in the readme?

asaunier:
Updating the README would indeed be useful. Then I'm +1 for merging.

gberaudo:
done

PR #33, opened 2015-10-07
Prevent that document_id and version can be set

tsauerwein:
... when adding a document.

Thanks @asaunier for noticing!


asaunier:
Looks good to me, thanks.

PR #34, opened 2015-10-07
Add JSON error handler

### LINKED ISSUE #31: Add error handler
... so that always JSON is returned.

tsauerwein:
... so that always JSON is returned.

Closes #31

### DIFF
--- a/c2corg_api/views/__init__.py
+++ b/c2corg_api/views/__init__.py
@@ -1,6 +1,32 @@
 import collections
 import datetime
 from colander import null
+from pyramid.httpexceptions import HTTPError, HTTPNotFound
+from pyramid.view import view_config
+from cornice import Errors
+from cornice.util import json_error
+
+
+@view_config(context=HTTPNotFound)
+@view_config(context=HTTPError)
+def http_error_handler(exc, request):
+    """In case of a HTTP error, return the error details as JSON, e.g.:
+
+        {
+            "status": "error",
+            "errors": [
+                {
+                    "location": "request",
+                    "name": "Not Found",
+                    "description": "document not found"
+                }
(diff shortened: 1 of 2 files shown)


c2corg_api/views/__init__.py
```diff
@@ -1,6 +1,32 @@
...
 import datetime
 from colander import null
+from pyramid.httpexceptions import HTTPError, HTTPNotFound
+from pyramid.view import view_config
+from cornice import Errors
+from cornice.util import json_error
+
+
+@view_config(context=HTTPNotFound)
+@view_config(context=HTTPError)
```
asaunier: Does this means that when an HTTPNotFound or HTTPError is detected, http_error_handler() will automatically be triggered to add the JSON error output as the response?

tsauerwein: Yes.

Actually HTTPNotFound [should](http://docs.pylonsproject.org/projects/pyramid/en/latest/api/httpexceptions.html#pyramid.httpexceptions.HTTPNotFound) also be an instance of HTTPError, but I had to specify it explicitly.

asaunier: OK, looks good to me.

Out of curiosity: the error names (such as "Conflict" - see the tests above) are set by Cornice?

tsauerwein: By Pyramid in this case. When you raise/return HTTPConflict, Pyramid sets the title (see [here](http://docs.pylonsproject.org/projects/pyramid/en/latest/_modules/pyramid/httpexceptions.html#HTTPConflict)).

tsauerwein:
Thanks for the review, Alex.

PR #35, opened 2015-10-07
Install Cornice from PyPi

tsauerwein:
Cornice 1.1.0 was released last week, which includes https://github.com/mozilla-services/cornice/pull/309, so that we no longer need to install from the GitHub repository.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,4 @@
 --index-url http://pypi.camptocamp.net/pypi
 -e git+https://github.com/camptocamp/pyramid_closure#egg=pyramid_closure
 
-# needed for Cornice: https://github.com/mozilla-services/cornice/pull/309
-git+https://github.com/mozilla-services/cornice.git
-
 -e .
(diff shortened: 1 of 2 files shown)


elemoine:
LGTM.

asaunier:
By the way, would it be relevant to fix the "requires" versions?

tsauerwein:
You mean this?
https://github.com/c2corg/v6_api/commit/6de2431c79a77f495fda4c20e89fe1fbb2def56d#diff-2eeaed663bd0d25b7e608891384b7298R20

asaunier:
Yes
The current way is great to get up-to-date depencies but we don't really control what version will be installed.

tsauerwein:
Ah I see, you mean to set a fixed version for all dependencies? That would make sense.

elemoine:
It makes sense. The dependency versions should be set in the `requirements.txt` file as opposed to `setup.py`.

PR #39, opened 2015-10-09
Make version fields default to 1

asaunier:
I wonder if we don't need to set a default version value in the document* tables in order to ease the migration of the existing documents.

Not sure the suggested change is enough though: I have tested to run scripts/create_user_db.sh but not default value is provided for version fields in document* tables.


tsauerwein:
Sounds good.

PR #42, opened 2015-10-14
Add geometry field to document

tsauerwein:
The geometries are stocked in a separate table, to avoid that potentially big geometries (e.g. tracks) are copied when only a figure was changed.


asaunier:
Looks good, thanks!

tsauerwein:
Thanks for the review.

PR #43, opened 2015-10-14
Set content-type for POST/PUT services

### LINKED ISSUE #8: Bad error message when header Content-Type is missing
When POSTing a new waypoint, if the Content-Type header is missing, one gets a "missing waypoint_type" error even though the waypoint_type is correctly set.
This is confusing.

tsauerwein:
Closes #8

### DIFF
--- a/c2corg_api/tests/views/test_route.py
+++ b/c2corg_api/tests/views/test_route.py
@@ -57,6 +57,9 @@ class TestRouteRest(BaseTestRest):
         }
         self.post_non_whitelisted_attribute(body)
 
+    def test_post_missing_content_type(self):
+        self.post_missing_content_type({})
+
     def test_post_success(self):
         body = {
             'activities': 'hiking',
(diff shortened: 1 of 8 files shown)


c2corg_api/tests/views/test_route.py
```diff
@@ -57,6 +57,9 @@ def test_post_non_whitelisted_attribute(self):
         }
         self.post_non_whitelisted_attribute(body)
 
+    def test_post_missing_content_type(self):
```
asaunier: Since this test is the same for all document types, can't we move it to the parent class BaseTestRest? Or maybe it would not be executed for each doc type then?

asaunier:
That looks good to me, thanks!

c2corg_api/tests/views/test_route.py
tsauerwein: You are right, this test could also be in `BaseTestRest`. But actually I would prefer to also list it in the children classes. So that by looking at `TestRouteRest` you directly see which tests are run for the routes.

PR #47, opened 2015-10-20
Add CORS headers

asaunier:
In the UI when I try to push data to the API I get the following error in my JS dev tool:
```
XMLHttpRequest cannot load http://c2corgv6-demo.gis.internal:6545/waypoints. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://c2corgv6-demo.gis.internal:6555' is therefore not allowed access. The response had HTTP status code 405.
```

I have found this solution:
http://stackoverflow.com/a/22862087

but not sure it is correct/enough since I then get the following error:
```
MLHttpRequest cannot load http://c2corgv6-demo.gis.internal:6545/waypoints. Invalid HTTP status code 405
```

### DIFF
--- a/c2corg_api/__init__.py
+++ b/c2corg_api/__init__.py
@@ -1,10 +1,24 @@
 from pyramid.config import Configurator
 from sqlalchemy import engine_from_config
+from pyramid.events import NewRequest
 
 from c2corg_api.models import (
     DBSession,
     Base,
     )
+
+
+def add_cors_headers_response_callback(event):
+    def cors_headers(request, response):
+        response.headers.update({
+            'Access-Control-Allow-Origin': '*',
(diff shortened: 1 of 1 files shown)


elemoine:
Just in case: do you know that Cornice includes support for CORS? See https://cornice.readthedocs.org/en/latest/api.html?highlight=cors.

tsauerwein:
I would rather use Cornice to handle the CORS support. This would also allow us to configure it on a per-service base.

This article is interesting:
https://blog.mozilla.org/services/2013/02/04/implementing-cross-origin-resource-sharing-cors-for-cornice/

asaunier:
Thanks for the tips!
You're right, it's better to use Cornice built-in tools, I will make a new PR with this solution.

PR #48, opened 2015-10-22
Add CORS parameters to Cornice services

asaunier:
Replace #47 

Based upon
https://blog.mozilla.org/services/2013/02/04/implementing-cross-origin-resource-sharing-cors-for-cornice/
https://cornice.readthedocs.org/en/latest/api.html?highlight=cors

I have tried to add the cors_policy to the DocumentRest class directly but it failed, most likely because I had omitted the path params in the @resource decorator. Anyway it's probably better to have all the cornice params explicitly set for each service.

### DIFF
--- a/c2corg_api/views/__init__.py
+++ b/c2corg_api/views/__init__.py
@@ -12,6 +12,12 @@ from shapely.geometry import mapping
 import json
 
 
+cors_policy = dict(
+    headers=('Content-Type'),
+    origins=('*')
+)
+
+
 @view_config(context=HTTPNotFound)
 @view_config(context=HTTPError)
 def http_error_handler(exc, request):
--- a/c2corg_api/views/image.py
+++ b/c2corg_api/views/image.py
@@ -2,11 +2,12 @@ from cornice.resource import resource, view
 
 from c2corg_api.models.image import Image, schema_image, schema_update_image
 from c2corg_api.views.document import DocumentRest
-from c2corg_api.views import json_view
+from c2corg_api.views import json_view, cors_policy
 from c2corg_api.views.validation import validate_id
 
 
-@resource(collection_path='/images', path='/images/{id}')
+@resource(collection_path='/images', path='/images/{id}',
+          cors_policy=cors_policy)
 class ImageRest(DocumentRest):
 
     def collection_get(self):
--- a/c2corg_api/__init__.py
+++ b/c2corg_api/__init__.py
@@ -5,6 +5,7 @@ from c2corg_api.models import (
     DBSession,
     Base,
     )
+from c2corg_api.views import cors_policy
 
 
 def main(global_config, **settings):
@@ -14,6 +15,11 @@ def main(global_config, **settings):
     DBSession.configure(bind=engine)
     Base.metadata.bind = engine
     config = Configurator(settings=settings)
+
+    origins = settings['cors.origins']
+    cors_policy['origins'] = origins.split(',')
+
     config.include('cornice')
     config.scan(ignore='c2corg_api.tests')
+
     return config.make_wsgi_app()
--- a/c2corg_api/views/route.py
+++ b/c2corg_api/views/route.py
@@ -5,7 +5,7 @@ from c2corg_api.models.route import Route, schema_route, schema_update_route
 from c2corg_api.models.schema_utils import restrict_schema
 from c2corg_api.views.document import DocumentRest, make_validator_create, \
     make_validator_update, make_schema_adaptor
-from c2corg_api.views import json_view
+from c2corg_api.views import json_view, cors_policy
 from c2corg_api.views.validation import validate_id
 from c2corg_common.fields_route import fields_route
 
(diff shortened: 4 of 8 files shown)


asaunier:
I have made a rebase to fix some conflicts with master. In the process it seems that @tsauerwein following comment is now less visible: https://github.com/c2corg/v6_api/commit/68e20c3d73b92cf368e58317a45eadcf47c050b0#commitcomment-13926722

tsauerwein:
I pushed a commit to your branch making the origins configurable. Could you please test with that?

asaunier:
I have juste made a rebase origin/master to take your #50 changes into account and fixed a conflict. The rebase is pushed. The API works as expected. I have not had the opportunity to test with the UI calls, it's currently broken in my instance...

asaunier:
Thanks for the improvements by the way!

asaunier:
Well, I think there's a problem:
http://c2corgv6-demo.gis.internal:6545/waypoints/1?l=fr  returns attributes as expected but I get the following error when calling the URL in Angular:
> GET http://c2corgv6-demo.gis.internal:6545/waypoints/1?l=fr 
> XMLHttpRequest cannot load http://c2corgv6-demo.gis.internal:6545/waypoints/1?l=fr. No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'http://c2corgv6-demo.gis.internal:6555' is therefore not allowed access. The response had HTTP status code 400.

By the way: I agree that we should avoid that anyone may write using the API, but that's probably wanted that anyone may read the documents (at least non user-related docs). For instance if a hut website would like to show the outings starting from the hut (not sure we want to add websites in the config in that case).

asaunier:
In addition shouldn't we set the cors-origins values in the config/* files rather than directly in common.ini.in?

asaunier:
At least when setting cors.origins to * in https://github.com/c2corg/v6_api/blob/cornice_cors/common.ini.in#L12 it seems to work :P

tsauerwein:
> By the way: I agree that we should avoid that anyone may write using the API, but that's probably wanted that anyone may read the documents (at least non user-related docs). For instance if a hut website would like to show the outings starting from the hut (not sure we want to add websites in the config in that case).

So, let's use "*" for uncritical GET requests and a specific origin for POST/PUT and critical GET requests (like for the user data)?

> In addition shouldn't we set the cors-origins values in the config/* files rather than directly in common.ini.in?

Oh yes, I originally wanted to that, but then forgot.

asaunier:
> So, let's use "*" for uncritical GET requests and a specific origin for POST/PUT and critical GET requests (like for the user data)?

Makes sense indeed

gberaudo:
We discussed with @asaunier and it appears we should have two policies:
- one for public data, restricted to `GET` method and allowed for `*`;
- the other for the rest (restricted data), only allowed for c2corg and requiring credentials (JWT).

c2corg_api/views/__init__.py
```diff
@@ -12,6 +12,12 @@
 import json
 
 
+cors_policy = dict(
```
gberaudo: Would be renamed to cors_public_policy.

c2corg_api/views/image.py
```diff
@@ -2,11 +2,12 @@
 
 from c2corg_api.models.image import Image, schema_image, schema_update_image
 from c2corg_api.views.document import DocumentRest
-from c2corg_api.views import json_view
+from c2corg_api.views import json_view, cors_policy
 from c2corg_api.views.validation import validate_id
 
 
-@resource(collection_path='/images', path='/images/{id}')
+@resource(collection_path='/images', path='/images/{id}',
```
gberaudo: The policy should be passed to each view (GET and PUT should be handled differently).

tsauerwein:
Yes, that's what I also hand in mind.

c2corg_api/views/image.py
gberaudo: We could use annotations like: `@public_json_view` and `@restricted_json_view` instead of `@json_view`.

tsauerwein: Good idea!

asaunier:
I have rebased this branch to get the last changes in master and fixed the conflicts.

``cors.origins = c2corgv6-demo.gis.internal`` does not work on my instance on the dev server (got cross origin errors). I had to set it to ``cors.origins = *``.

What about merging this PR (with ``cors.origins = *``) and creating an issue to keep track of the required fix/improvements still to apply? We need this PR to make the UI work.

asaunier:
With @gberaudo we have figured out that the cors.origins should contain the port number as well as the URL scheme (http://). We may use * to allow whatever port and scheme. For instance
```
foo,*c2corgv6-demo.gis.internal:*,fdsfsdfs
```
works. 
I will suggest changes to get the cors.origins from the makefile.

asaunier:
What the use of "internalurl" set in https://github.com/c2corg/v6_api/blob/cornice_cors/config/default#L11 and https://github.com/c2corg/v6_api/blob/cornice_cors/config/dev#L6 ?

It seems to be used nowhere else.
Should I drop it?

tsauerwein:
Looks good!

Will you open tickets for https://github.com/c2corg/v6_api/pull/48#issuecomment-150605457 ?

asaunier:
> Looks good!

Thanks. By the way I have dropped "internalurl", I hope you have seen it.

> Will you open tickets for #48 (comment) ?

Yes, see #54

PR #50, opened 2015-10-23
Use custom schema for each waypoint type

### LINKED ISSUE #37: Returning a document for a single culture
When displaying a document, in most cases it is no use retrieving all cultures, since only one is displayed.

As for now the API returns all available cultures for a given document call.
It would be useful to be able to specify the desired culture.

tsauerwein:
... so that only the fields specific for a waypoint type are returned.

Also adds the column `available_cultures`.

Closes #37

### DIFF
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -4,17 +4,19 @@ from sqlalchemy import (
     Boolean,
     String,
     ForeignKey,
-    Enum
+    Enum,
+    func
     )
 from sqlalchemy.ext.declarative import declared_attr
 from sqlalchemy.orm import relationship
+from sqlalchemy.dialects import postgresql
 from geoalchemy2 import Geometry
 from colander import MappingSchema, SchemaNode, String as ColanderString, null
 from itertools import ifilter
 import abc
 import enum
 
-from c2corg_api.models import Base, schema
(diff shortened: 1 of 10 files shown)


asaunier:
"available_cultures" => a field that is only available in the main doc table, not in the archives, right? The info is then available only in one place per document. Looks good.

tsauerwein:
> "available_cultures" => a field that is only available in the main doc table, not in the archives, right?

Yes exactly. This field is not stored in the database. It is simply set before creating the JSON:
https://github.com/c2corg/v6_api/commit/4dd7a6ae694f90d1bf06fd13507dd5680df6fdc2#diff-b906af94797c5db6c394971ac60bf7a4R31

asaunier:
OK. Would it be relevent to store it in the documents table in order to avoid the query used to fill this attribute?

tsauerwein:
When we would store it in the database, we would have to make sure that this field is updated every time a locale is added/removed.

Also note that only a single query is made when requesting a list of documents. I don't think this additional query is a performance issue. Especially once we are enabling the cache.

asaunier:
OK it makes sense.
From my perspective, it's OK to merge this PR.

tsauerwein:
Thanks!

PR #51, opened 2015-10-23
Use common project

tsauerwein:
Use the new [common project](https://github.com/c2corg/v6_common).

For development a local checkout of the project can be used by changing the following in `requirements.txt`:

    -e ../v6_common
    # git+https://github.com/c2corg/v6_common.git


asaunier:
Looks good

PR #52, opened 2015-10-23
Use a custom schema for listings

### LINKED ISSUE #41: List of attributes for listing
As for now a listing request (eg. waypoints/) returns all the attributes of every documents.

In most cases only a few attributes such as the title, elevation, etc. are actually needed.
http://www.camptocamp.org/outings

In some cases, having a more comprehensive set of attributes is useful. For instance the "conditions" page:
http://www.camptocamp.org/outings/conditions/orderby/date/order/desc

tsauerwein:
Closes #41

### DIFF
--- a/c2corg_api/views/waypoint.py
+++ b/c2corg_api/views/waypoint.py
@@ -26,22 +26,35 @@ def validate_waypoint_update(request):
 @lru_cache(maxsize=None)
 def adapt_schema_for_type(waypoint_type):
     """Get the schema for a waypoint type.
-    The schemas are cached.
+    All schemas are cached using memoization with @lru_cache.
     """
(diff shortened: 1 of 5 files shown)


c2corg_api/views/waypoint.py
```diff
@@ -32,16 +32,29 @@ def adapt_schema_for_type(waypoint_type):
     return restrict_schema(schema_waypoint, fields)
 
 
+@lru_cache(maxsize=None)
```
gberaudo: It is hard to understand at first glance what `@lru_cache` does.
I think you should add a comment each time explaining what it is used for.

tsauerwein: Ok.

gberaudo: Thanks.

PR #58, opened 2015-10-29
Add database migration script (for waypoints)

tsauerwein:
Currently only for summits (including archive versions and history).

The new [bulk insertion functions](http://docs.sqlalchemy.org/en/rel_1_0/orm/persistence_techniques.html#bulk-operations) of SQLAlchemy 1.0.0 are used to avoid the overhead of the ORM.

To do before merging this PR:

* [x] Migrate all other waypoint types
 - [x] summits
 - [x] sites
 - [x] huts
 - [x] parkings
 - [x] products
* [x] Make sure to adjust the sequence values because the the old ids are used.


c2corg_api/scripts/migration/documents/document.py
```diff
@@ -0,0 +1,143 @@
...
+            version = 1
+            archives = []
+            geometry_archives = []
+            batch = DocumentBatch(
+                self.session_target, self.batch_size,
+                self.get_model_document(locales),
+                self.get_model_archive_document(locales),
+                self.get_model_geometry(),
+                self.get_model_archive_geometry())
+            with batch:
```
elemoine: Couldn't you remove the nested `with` by using the following?

```py
with transaction.manager, batch:
    …
```

tsauerwein: Nice! I was wondering if you could do something like that.

tsauerwein:
Now ready to review.

asaunier:
Arf, are my changes into common the reason of the tests failing?

Else your work seems excellent to me. I don't have paid much attention to the exact conversion of the v5 fields to v6 tables. I guess we can run the script and check the results in the UI (even let the client have a look at the new migrated documents).

tsauerwein:
> Arf, are my changes into common the reason of the tests failing?

I think #60 broke some tests. I will fix them and add the commits to this branch.

asaunier:
One "little" thing: in v5 there are special summits called "raid". For instance: http://www.camptocamp.org/summits/115772/fr/haute-route-chamonix-zermatt
I think (not sure) they are migrated to the new type "virtual".
Have you taken them into account?

tsauerwein:
> I think (not sure) they are migrated to the new type "virtual".
> Have you taken them into account?

Ah ok, "raid" is now "virtual". Good to know. I will change it.

asaunier:
Looks good to me.

tsauerwein:
Thanks!

PR #59, opened 2015-10-30
Fix route type

### LINKED ISSUE #56: route_types values for route model are incorrect
Currently:
https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L55-L70

They should be:
* return_same_way
* loop
* traverse

https://github.com/c2corg/camptocamp.org/blob/602b46c9388436cbbd21f99dde856065d33d0fed/apps/frontend/modules/routes/config/module.yml#L23-L25

### LINKED ISSUE #57: Make sure that a route may have several activities at a time

tsauerwein:
The fields of a route depend on the assigned activities.

Closes #57 
Closes #56

### DIFF
--- a/c2corg_api/models/enums.py
+++ b/c2corg_api/models/enums.py
@@ -54,8 +54,8 @@ climbing_rating = enum(
     'climbing_rating', attributes.climbing_ratings)
 equipment_rating = enum(
     'equipment_rating', attributes.equipment_ratings)
-route_form = enum(
-    'route_route_form', attributes.route_forms)
+route_type = enum(
+    'route_type', attributes.route_types)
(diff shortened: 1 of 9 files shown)


asaunier:
Looks good.
If I understand https://github.com/c2corg/v6_api/blob/df5d4d40ec54b1c277301335ba513c63849c9755/c2corg_api/models/route.py#L57 well, route_types is now a multi-value field?

tsauerwein:
Yes. Otherwise we still can change it later.

asaunier:
It's OK. That's what the provided table says for v6. I'll let you know if the client changes his mind.

PR #61, opened 2015-11-04
Upgrade dependencies on `make install`

tsauerwein:
Make sure that packages are updated when running ` make install`. This is to make sure that the latest version of the common project is used.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ serve: install development.ini
 .build/venv/bin/nosetests: .build/dev-requirements.timestamp
 
 .build/dev-requirements.timestamp: .build/venv dev-requirements.txt
-	.build/venv/bin/pip install -r dev-requirements.txt
+	.build/venv/bin/pip install --upgrade -r dev-requirements.txt
 	touch $@
 
 .build/venv:
@@ -73,7 +73,7 @@ serve: install development.ini
 	virtualenv --no-site-packages $@
 
 $(SITE_PACKAGES)/c2corg_api.egg-link: .build/venv requirements.txt setup.py
-	.build/venv/bin/pip install -r requirements.txt
+	.build/venv/bin/pip install --upgrade -r requirements.txt
 
 development.ini production.ini: common.ini
 


asaunier:
Thanks! Doing the same for the UI would make sense as well. If you want I can do it.

gberaudo:
Does this work when requirements.txt is not changed?

tsauerwein:
> Does this work when requirements.txt is not changed?

Yes.

asaunier:
I have ported the change to the UI as well: https://github.com/c2corg/v6_ui/pull/16
Feel free to merge it as well when you merge this one.

gberaudo:
It is OK for me though I would prefer a new target `upgrade-requirements` for upgrading dev-requirements.txt and requirements.txt, which could be invoked manually.

gberaudo:
Currently we depend on the master version of the common project.
Would it be possible to specify a commit hash instead?

It would avoid any mismatch between the API project and the common project.

tsauerwein:
Yes, you can specify a commit hash or branch:
https://pip.pypa.io/en/latest/reference/pip_install/#git

gberaudo:
Great, thanks for the info.

asaunier:
Do we still have things to do before merging the PR?

asaunier:
Merging this PR would really help.

At the moment tests fails in my API instance (branch master) because some fields have changed in the common part:
```
AssertionError: orientation not in <class 'c2corg_api.models.route.Route'>
```
related to https://github.com/c2corg/v6_common/commit/84de68937ac86f2dad35bae94a7b1e71f8898b45

Ditto in Travis checks for https://github.com/c2corg/v6_api/pull/21

I'm +1 for merging this PR as soon as possible :)

gberaudo:
+1

tsauerwein:
> It is OK for me though I would prefer a new target upgrade-requirements for upgrading dev-requirements.txt and requirements.txt, which could be invoked manually.

I agree with you. Now, every time when running `make serve` the dependencies in `requirements.txt` are downloaded again. This is causing the ui tests to fail on Travis because the server takes too long to start. I think I will add an `upgrade` target.

PR #66, opened 2015-11-05
Manually set inheritance column in migration

### LINKED ISSUE #64: TypeError: to_archive() takes exactly 2 arguments (1 given)
When trying to update a waypoint in the UI, the API fails with the following error:
>  File "/home/alex/v6_api/c2corg_api/views/waypoint.py", line 57, in put
    return self._put(Waypoint, schema_waypoint)
  File "/home/alex/v6_api/c2corg_api/views/document.py", line 86, in _put
    changed_langs)
  File "/home/alex/v6_api/c2corg_api/views/document.py", line 154, in _update_version
    locale_archive = self._get_locale_archive(locale, changed_langs)
  File "/home/alex/v6_api/c2corg_api/views/document.py", line 215, in _get_locale_archive
    locale_archive = locale.to_archive()
TypeError: to_archive() takes exactly 2 arguments (1 given)

Line 215 is
```
        if locale.culture in changed_langs:
            # create new archive version for this locale
            locale_archive = locale.to_archive()
```
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document.py#L213-L215
whereas DocumentLocale.to_locale() indeed requires a "locale" argument:
```
    def to_archive(self, locale):
        copy_attributes(self, locale, DocumentLocale._ATTRIBUTES)
        return locale
```
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/document.py#L236-L238

tsauerwein:
The type column in the documents table is the [discriminator column](http://docs.sqlalchemy.org/en/rel_1_0/orm/inheritance.html#joined-table-inheritance) for the inheritance. Usually this column is set automatically by SQLAlchemy. But when doing bulk-insertions, it seems that you have to set it manually.

Closes #64

### DIFF
--- a/c2corg_api/scripts/migration/documents/document.py
+++ b/c2corg_api/scripts/migration/documents/document.py
@@ -4,6 +4,8 @@ import zope
 import abc
 import re
 
+from c2corg_api.models.document import DocumentGeometry, \
+    ArchiveDocumentGeometry
 from c2corg_api.scripts.migration.documents.batch_document import DocumentBatch
 from c2corg_api.scripts.migration.migrate_base import MigrateBase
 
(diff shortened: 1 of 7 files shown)


asaunier:
Thanks for the fix. Looks good.

asaunier:
Is there something we can easily do with the already imported data?

tsauerwein:
You could manually set the values in the database. But because the migration script takes less than 10 minutes to run, I'd just run it again.

asaunier:
Makes sense. Thanks for answering.

Issue #70, opened 2015-11-06
Auth TODO


gberaudo:
- [x] add user table and methods;
- [x] add stubbed user creation methods;
- [x] add constraint on user on history_metadata (set user when creating a document version)
- [x] set user in migration script
- [x] store and retrieve and delete active tokens from database;
- [x] add token refresh method;
- [x] add or modify unit tests to check methods permissions;
- [x] ~~read and write cookies? (see https://github.com/c2corg/v6_api/wiki/Authentication-Concept)~~
- [x] check password migration from v5, notably unicode passwords
- [x] add index to the token table on the value
- [x] discourse integration
- [x] clean expired tokens from database (see https://github.com/c2corg/v6_api/commit/bbdef2a1c687dd668a1318836cd4b3cf87733529)

asaunier:
@gberaudo Could you please update the above list status?

asaunier:
Can the "discourse integration" item be checked?

gberaudo:
Done

asaunier:
What remains to do with 
> ￼ clean expired tokens from database (see bbdef2a)

?

gberaudo:
Yes

asaunier:
I mean, what should be done with this task since there's a commit that seems to do the job? :P

gberaudo:
This commit was actually reverted later following @tsauerwein request.

PR #71, opened 2015-11-06
Disable auth check for dev instances

asaunier:
Else it is no longer possible to make POST/PUT requests to the API from the UI until the authentication tool is available.

### DIFF
--- a/common.ini.in
+++ b/common.ini.in
@@ -3,6 +3,9 @@
 [app:main]
 use = egg:c2corg_api
 
+# Disable the authorization checks if True
+noauthorization = {noauthorization}
+
 pyramid.default_locale_name = en
 
 version = {version}
--- a/config/default
+++ b/config/default
@@ -10,6 +10,7 @@ host = c2corgv6.demo-camptocamp.com
 url = http://$(host)
 
 export cors_origins = $(url)
+export noauthorization = False
 
 export db_host = localhost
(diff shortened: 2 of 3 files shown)


tsauerwein:
:+1:

asaunier:
Since it is not a big change (same than for the tests) and perhaps only temporary (until the login is implemented), I merge the PR.
@gberaudo I let you suggest a new PR if a better solution would be relevent.

gberaudo:
As discussed at the meeting, this PR is fine for me.
The alternative would be to use one of the available (hardcoded) user
accounts.

asaunier:
> The alternative would be to use one of the available (hardcoded) user
accounts.

This might be a good idea in order to actually have a user linked to the write actions. I have not looked at the DB but I guess that can be a problem in the document history if no user is provided? Isn't there a constraint in the history_metadata table about that?

Anyway I think this is really temporary since when the auth is ready it will be more realistic to use real users in dev instances.

gberaudo:
Updated https://github.com/c2corg/v6_api/issues/70

tsauerwein:
> This might be a good idea in order to actually have a user linked to the write actions. I have not looked at the DB but I guess that can be a problem in the document history if no user is provided? Isn't there a constraint in the history_metadata table about that?

At the moment the user is not set when creating a document version.

PR #75, opened 2015-11-12
Fix comparator warning

gberaudo:
This results in a contradiction, which nonetheless can be expensive
to evaluate.  Consider alternative strategies for improved performance.
'strategies for improved performance.' % expr)


tsauerwein:
Great, thanks!

PR #76, opened 2015-11-12
Register and get user

gberaudo:
Register a user.
Get user info by id.

### DIFF
--- a/c2corg_api/views/user.py
+++ b/c2corg_api/views/user.py
@@ -1,0 +1,86 @@
+from functools import partial
+from pyramid.httpexceptions import HTTPInternalServerError
+
+from cornice.resource import resource, view
+
+from c2corg_api.models.user import User, schema_user, schema_create_user
+from c2corg_api.views import (
+        cors_policy, json_view, to_json_dict)
+from c2corg_api.views.validation import validate_id
+
+from c2corg_api.models import DBSession
+
+ENCODING = 'UTF-8'
+
+
+def validate_json_password(request):
+    """Checks if the password was given and encodes it.
+       This is done here as the password is not an SQLAlchemy field.
+       In addition, we can ensure the password is not leaked in the
+       validation error messages.
+    """
+
+    if 'password' not in request.json:
+        request.errors.add('user', 'password', 'Required')
+
+    try:
+        # We receive a unicode string. The hashing function used
+        # later on requires plain string otherwise it raises
+        # the "Unicode-objects must be encoded before hashing" error.
+        password = request.json['password']
+        request.validated['password'] = password.encode(ENCODING)
+    except:
+        request.errors.add('user', 'password', 'Invalid')
+
+
+def validate_unique_attribute(attrname, request):
+    """Checks if the given attribute is unique.
+    """
+
+    if attrname in request.json:
+        value = request.json[attrname]
+        attr = getattr(User, attrname)
+        count = DBSession.query(User).filter(attr == value).count()
+        if count == 0:
+            request.validated[attrname] = value
+        else:
+            request.errors.add('user', attrname, 'already used ' + attrname)
+
+
+@resource(path='/users/{id}', cors_policy=cors_policy)
+class UserRest(object):
+    def __init__(self, request):
+        self.request = request
+
+    @view(validators=validate_id)
+    def get(self):
+        id = self.request.validated['id']
+        user = DBSession. \
+            query(User). \
+            filter(User.id == id). \
+            first()
+
+        return to_json_dict(user, schema_user)
+
+
+@resource(path='/users/register', cors_policy=cors_policy)
+class UserRegistrationRest(object):
+    def __init__(self, request):
+        self.request = request
+
+    @json_view(schema=schema_create_user, validators=[
+        validate_json_password,
+        partial(validate_unique_attribute, "email"),
+        partial(validate_unique_attribute, "username")])
+    def post(self):
+        user = schema_create_user.objectify(self.request.validated)
+        user.password = self.request.validated['password']
+
+        DBSession.add(user)
+        try:
+            DBSession.flush()
+        except:
+            # TODO: log the error for debugging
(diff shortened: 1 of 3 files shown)


gberaudo:
> Should we use a more expressive route? E.g. POST /users/register instead of `POST /users'
@tsauerwein, indeed it would be greater. Do I have to add a new `resource` for this? Or switch to the `service` API?

tsauerwein:
> Do I have to add a newresourcefor this? Or switch to theservice` API?

I think a 2nd resource with the path `collection_path='/users/register'` could do the trick.

gberaudo:
> Why is it necessary to get the password in utf-8? Isn't everything already in utf-8?
No idea, I get `Unicode-objects must be encoded before hashing` error otherwise.

gberaudo:
Ready for a new review.

gberaudo:
Almost. Password requirement does not work though it is in the schema. Any idea why?

tsauerwein:
> Password requirement does not work though it is in the schema. Any idea why?

Probably because `password` is a property, which might not be understood by ColanderAlchemy.

c2corg_api/views/user.py
```diff
@@ -0,0 +1,65 @@
...
+        # "Unicode-objects must be encoded before hashing" error
+        user.password = self.request.json['password'].encode('utf-8')
+
+        DBSession.add(user)
+        try:
+            DBSession.flush()
+        except:
+            # Catching the exception ensures a json is sent back to the user
+            # and a message is shown.
+            self.request.errors.add('user', 'error', 'an error occured')
```
tsauerwein: The three values for the error have a specific meaning, see:
http://cornice.readthedocs.org/en/latest/validation.html#dealing-with-errors

You could use:

    self.request.errors.add('body', 'user', 'error creating user')

tsauerwein: But I am wondering: If you simply add something to this `errors` list (outside the validation), does this return the error messages? Because one line below you are returning the user as if everything worked fine.
I think you would have to raise a `BadRequest` like here: https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document.py#L321-L322

tsauerwein:
You could explicitly add the password field to the schema:

    from colander import SchemaNode, String as ColanderString, required
    ...
    schema_create_user.children.append(SchemaNode(ColanderString(), name='password', missing=required))

(untested, but something in this direction)

c2corg_api/views/user.py
gberaudo: Yes, adding something to the list will return validation errors.
But you are right raising would make sense; it is client code duty to detect HTTP errors and update interface accordingly. I think in this case an `InternalError` would be great.

Shouldn't we log the error? I see we are not using loggers yet.
How do we deal with sensitive data?

gberaudo:
I fixed the validation issue with `password`. I tried something similar to what you propose which I found
in the colander tests code but it did not work, so I am simply using a validator.
In some way it makes sense because we only handle deserialization.

gberaudo:
@asaunier, do you know if users can use non-ascii characters in v5?
I currently accept unicode as password input but it could easily be restricted to ascii instead.

c2corg_api/views/user.py
tsauerwein: > Yes, adding something to the list will return validation errors.

Inside a validation function, yes. But outside I think no?

> Shouldn't we log the error? I see we are not using loggers yet.

Yes, I created a ticket: https://github.com/c2corg/v6_api/issues/78

> How do we deal with sensitive data?

What do you mean? (Never log a password or token, but I think that is clear.)

gberaudo: I mean that some stack traces may contain sensitive data.

gberaudo:
I think it is now ready for merging. Do you have more comments?

asaunier:
> @asaunier, do you know if users can use non-ascii characters in v5?

You mean for passwords?
I think that any UTF-8 character is accepted. Why restricting to ascii?

The code of v5 about password updating:
https://github.com/c2corg/camptocamp.org/blob/602b46c9388436cbbd21f99dde856065d33d0fed/apps/frontend/modules/users/actions/actions.class.php#L458-L470

When someone signs in, it seems a password is auto-generated and send by email (same tool than for get a new password when you don't remember your password).

Generated passwords actually use ascii characters:
https://github.com/c2corg/camptocamp.org/blob/602b46c9388436cbbd21f99dde856065d33d0fed/lib/model/doctrine/UserPrivateData.class.php#L64-L80
but I have seen no restriction (?) when the user updates his password.

If really need, restricting passwords to ascii characters could be possible. It means that users we non-ascii passwords would have to ask for a new password. That would be better to avoid this but that's a possibility.

c2corg_api/views/user.py
```diff
@@ -0,0 +1,86 @@
...
+        cors_policy, json_view, to_json_dict)
+from c2corg_api.views.validation import validate_id
+
+from c2corg_api.models import DBSession
+
+ENCODING = 'UTF-8'
+
+
+def validate_json_password(request):
+    """Checks if the password was given and encode it.
```
asaunier: encodeS (or remove the trailing s in Checks)

c2corg_api/views/user.py
```diff
@@ -0,0 +1,86 @@
...
+       This is done here as the password is not an SQLAlchemy field.
+       In addition, we can ensure the password is not leaked in the
+       validation error messages.
+    """
+
+    if 'password' not in request.json:
+        request.errors.add('user', 'password', 'Required')
+
+    try:
+        # We receives an unicode string. The hashing function used
```
asaunier: an unicode => a unicode

asaunier:
I have no other comment.

gberaudo:
Fixed the typos, thanks @asaunier.
Regarding passwords, notably utf8, I added an item in https://github.com/c2corg/v6_api/issues/70 so that it is correctly checked.

gberaudo:
Thanks for the reviews @tsauerwein and @asaunier.

PR #77, opened 2015-11-17
Add `upgrade` target in Makefile

tsauerwein:
This reverts #61 to avoid that the dependencies are updated every time when running `make serve`. This caused problems with the ui tests on Travis because the server took too long to start.

Instead the dependencies can be upgraded by manually running `make -f ... upgrade` or `make -f ... upgrade-dev`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,8 @@ help:
 	@echo "Secondary targets:"
 	@echo
 	@echo "- lint			Run flake8 checker on the Python code"
+	@echo "- upgrade		Upgrade the dependencies."
+	@echo "- upgrade-dev		Upgrade the dev. dependencies."
 
 .PHONY: check
 check: lint test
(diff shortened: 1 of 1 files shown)


asaunier:
Looks good to me. @gberaudo OK for you?

@tsauerwein Should we do the same for the UI? If yes, could you do it?

tsauerwein:
> @tsauerwein Should we do the same for the UI? If yes, could you do it?

Yes, it will do it.

Makefile
```diff
@@ -26,6 +26,8 @@ help:
 	@echo "Secondary targets:"
 	@echo
 	@echo "- lint			Run flake8 checker on the Python code"
+	@echo "- upgrade		Upgrade the dependencies."
+	@echo "- upgrade-dev	Upgrade the dev. dependencies."
```
asaunier: By the way, I think there are indentation problems in the UI makefile usage tips: https://github.com/c2corg/v6_ui/blob/master/Makefile#L26-L37
Do you how it should be formatted? Using a fixed number of tabs?

gberaudo:
+1

PR #79, opened 2015-11-20
Use database to log in and validate token

gberaudo:


gberaudo:
This PR is now ready for review.

tsauerwein:
Good work! :+1:

gberaudo:
I updated the PR. Now all tests must run with authentication/authorization enabled.
All post/put requests were updated to:
- first check that a 403 response is received if no credentials are sent;
- then that the test pass with the credentials of the `contributor` user.

Auth is still disabled by default in the ini file so that we have time to change the ui.

tsauerwein:
Thanks for updating the tests!

gberaudo:
Added index on token expire column.

gberaudo:
I just tested creating a waypoint using the UI and it works as expected (no authorization checks).
So I think that when Travis passes it will be ready for merging in master.

gberaudo:
Passed. Everyone OK for the merging?

asaunier:
No objection from my perspective.

tsauerwein:
Yes, please merge.

gberaudo:
Thanks @asaunier and @tsauerwein for the review.

Makefile
```diff
@@ -43,7 +43,8 @@ cleanall: clean
 
 .PHONY: test
 test: .build/venv/bin/nosetests test.ini
-	.build/venv/bin/nosetests
+	# All tests are now run with authentication/authorization enabled
```
tsauerwein: Now that all tests are run with authentication enabled, do we still need the flag `FORCE_AUTHORIZATION_ENABLED`?

gberaudo: Right, see https://github.com/c2corg/v6_api/pull/84

PR #80, opened 2015-11-23
Obey convention for cornice errors

gberaudo:
See
http://cornice.readthedocs.org/en/latest/validation.html#dealing-with-errors

With errors being a JSON dictionary with the keys “location”, “name” and “description”.
- location is the location of the error. It can be “querystring”, “header” or “body”
- name is the eventual name of the value that caused problems
- description is a description of the problem encountered.

### DIFF
--- a/c2corg_api/views/user.py
+++ b/c2corg_api/views/user.py
@@ -21,7 +21,7 @@ def validate_json_password(request):
     """
 
     if 'password' not in request.json:
-        request.errors.add('user', 'password', 'Required')
+        request.errors.add('body', 'password', 'Required')
 
     try:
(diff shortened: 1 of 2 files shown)


tsauerwein:
Thanks! :)

gberaudo:
@tsauerwein this PR also fixes other places where we did not obeyed the convention.
@asaunier, this change may require changes in the ui.

asaunier:
> this change may require changes in the ui.

OK thanks, please merge, we'll then see what we have to change in the UI.

gberaudo:
Thanks for the reviews.

PR #86, opened 2015-11-27
Improve try_login function

asaunier:
As for now if no user matches the username/password, ``user`` is None and we get an error when trying to run ``user.validate_password(password, DBSession)``.

By the way shouldn't we test sooner (at the beginning of the function, before making the user query?) that ``username`` and ``password`` are indeed set?

I have also added a ``return None`` statement because I think it is more consistent to return something anyway (maybe that's implicit but...).

### DIFF
--- a/c2corg_api/security/roles.py
+++ b/c2corg_api/security/roles.py
@@ -57,7 +57,7 @@ def try_login(username, password, request):
     user = DBSession.query(User). \
         filter(User.username == username).first()
 
-    if username and password and user.validate_password(password, DBSession):
+    if user and user.validate_password(password, DBSession):
         policy = request.registry.queryUtility(IAuthenticationPolicy)
         now = datetime.datetime.utcnow()
         exp = now + datetime.timedelta(weeks=CONST_EXPIRE_AFTER_DAYS)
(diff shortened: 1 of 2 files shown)


tsauerwein:
Good catch! 

> By the way shouldn't we test sooner (at the beginning of the function, before making the user query?) that username and password are indeed set?

That would indeed make sense.

asaunier:
I have just discussed those changes with @gberaudo 
I will make additional changes according to his directions.
Username and password are already tested earlier while validating the login route request.

asaunier:
PR updated. I have also removed some code @gberaudo has forgotten to remove in auth.py.
@gberaudo ``username`` was already OK in https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/user.py#L108-L110

tsauerwein:
> Username and password are already tested earlier while validating the login route request.

But the schema accepts empty user names. Or am I missing something?

gberaudo:
Thanks @asaunier.
@tsauerwein, see the validation schema at https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/user.py#L93.

tsauerwein:
@gberaudo The default is `missing=required`, ok.

PR #91, opened 2015-12-01
Add renew token view

gberaudo:
If the same token as an existing one is recreated, reuse it.

### DIFF
--- a/c2corg_api/security/roles.py
+++ b/c2corg_api/security/roles.py
@@ -30,16 +30,20 @@ def groupfinder(userid, request):
     return ['group:moderators'] if is_moderator else [Authenticated]
 
 
-def validate_token(token):
+def is_valid_token(token):
     now = datetime.datetime.utcnow()
     return DBSession.query(Token). \
         filter(Token.value == token and Token.expire > now).count() == 1
 
 
-def add_token(value, expire, userid):
-    token = Token(value=value, expire=expire, userid=userid)
-    DBSession.add(token)
-    DBSession.flush()
+def add_or_retrieve_token(value, expire, userid):
+    token = DBSession.query(Token). \
+        filter(Token.value == value and User.id == userid).first()
+    if not token:
+        token = Token(value=value, expire=expire, userid=userid)
+        DBSession.add(token)
+        DBSession.flush()
+
     return token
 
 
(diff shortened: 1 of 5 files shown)


c2corg_api/security/roles.py
```diff
@@ -36,10 +36,14 @@ def validate_token(token):
         filter(Token.value == token and Token.expire > now).count() == 1
 
 
-def add_token(value, expire, userid):
-    token = Token(value=value, expire=expire, userid=userid)
-    DBSession.add(token)
-    DBSession.flush()
+def add_or_retrieve_token(value, expire, userid):
+    token = token = DBSession.query(Token). \
```
asaunier: ``token = token = ...``
???

c2corg_api/security/roles.py
```diff
@@ -67,6 +71,22 @@ def try_login(user, password, request):
...
+        return add_or_retrieve_token(token, exp, user.id)
+
+    return None
+
+
+def renew_token(user, request):
+    old_token = extract_token(request)
+    valid = validate_token(old_token)
+
+    if valid:
```
asaunier: Why not
```
if validate_token(old_token):
```
instead?

c2corg_api/security/roles.py
```diff
@@ -67,6 +71,22 @@ def try_login(user, password, request):
...
+        policy = request.registry.queryUtility(IAuthenticationPolicy)
+        now = datetime.datetime.utcnow()
+        exp = now + datetime.timedelta(weeks=CONST_EXPIRE_AFTER_DAYS)
+        claims = create_claims(user, exp)
+        token_value = policy.encode_jwt(request, claims=claims)
+        token = add_or_retrieve_token(token_value, exp, user.id)
+        return token
```
asaunier:
```
return add_or_retrieve_token(token_value, exp, user.id)
```
???

asaunier:
A few comments else no big objection. Travis fails with some permission problem!?

c2corg_api/security/roles.py
gberaudo: Thanks, strange that the linter does not catch this.

gberaudo:
Thanks for the review @asaunier.
I removed the local variables and reworked the test.

asaunier:
OK for me

gberaudo:
Is it fine with you @tsauerwein?

tsauerwein:
When renewing a token I would have expected that the old token is removed from the database to avoid that there are unused but still valid token in the database.

But could this be a problem? For example from the UI you make a request to renew the token. Before you get the response, you make another authenticated request still with the old token. But in the mean time the old token was removed, so the request would fail.

So, it seems to be good to keep the old tokens. We just have to make sure that we don't renew to often.

tsauerwein:
And yes, this looks good to me!

gberaudo:
Fixed the computed expiration and added a test.
@tsauerwein, I agree with your analyze.

gberaudo:
Thanks for the reviews @asaunier and @tsauerwein.

PR #95, opened 2015-12-04
Add basic user migration from v5

gberaudo:
25 users are duplicated in v5 (same login_name).
A suffix is added to the oldest user.


tsauerwein:
That's looking good! Can you also update the sequence for the user id, like here:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/sequences.py

gberaudo:
Updated.
@tsauerwein, do you know why Travis fails?

tsauerwein:
Mhm no, tests pass locally.

gberaudo:
Found the issue and updated the PR.

gberaudo:

PR #96, opened 2015-12-08
Add waypoint/route history views

gberaudo:


c2corg_api/views/document.py
```diff
@@ -7,11 +7,14 @@
     UpdateType, DocumentLocale, ArchiveDocumentLocale, ArchiveDocument,
     ArchiveDocumentGeometry, set_available_cultures)
 from c2corg_api.models.document_history import HistoryMetaData, DocumentVersion
+from c2corg_api.models.user import User
 from c2corg_api.search.sync import sync_search_index
 from c2corg_api.views import to_json_dict
 from c2corg_api.views.validation import check_required_fields, \
     check_duplicate_locales
 
+from __init__ import serialize
```
asaunier:
```
from c2corg_api.views import serialize
```
would be more consistent and explicit. By the way you could use the existing import:
```
from c2corg_api.views import to_json_dict
```

c2corg_api/views/route.py
```diff
@@ -52,3 +51,13 @@ def collection_post(self):
                           validators=[validate_id, validate_route_update])
     def put(self):
         return self._put(Route, schema_route)
+
+
+# FIXME: handle /routes/history/{id} ?
+@resource(path='/history/route/{id}', cors_policy=cors_policy)
```
asaunier: Don't forget to take the culture into account!
We only need the versions for a given culture.

Wouldn't it be more relevant to use ``/routes/history/{id}/{lang}`` instead? "history" is an action about the document type "routes", I'm not sure it should be the first element of the route => first comes the resources name, then the action and finally the parameters. Or the document reference (doc type > id > culture) and finally the action.

c2corg_api/views/document.py
gberaudo: I initially tried serializing with a pure colander schema and using `to_json_dict` but it was not working. Then I tried calling `serialize` on the colander schema but it was not good.
I updated the import.

asaunier: Isn't it possible to group all imports from c2corg_api.views?
For instance
```
from c2corg_api.views import to_json_dict, serialize
```

gberaudo:
The history resource is now unique and available on the document:
`/document/{id}/history/{lang}`.

This change simplifies the API and avoid writting wrappers.
One type of document will require specific care: `Outing` will have access permission checked.

c2corg_api/views/document.py
```diff
@@ -406,3 +411,42 @@ def get_all_fields(fields, activities, field_list_type):
...
+            HistoryMetaData.written_at) \
+            .filter(DocumentVersion.document_id == id) \
+            .filter(
+                HistoryMetaData.id == DocumentVersion.history_metadata_id) \
+            .filter(User.id == HistoryMetaData.user_id) \
+            .filter(DocumentVersion.culture == lang) \
+            .order_by(DocumentVersion.id) \
+            .all()
+
+        return [serialize({
```
tsauerwein: * Please do not return an array as top-level element, see https://github.com/c2corg/v6_api/issues/49
* In this case you don't manually have to call `serialize`. Because you only have simple data types, the default Pyramid JSON serializer will automatically convert the list to JSON.

tsauerwein:
Beside the one comment, it looks good!

c2corg_api/views/document.py
gberaudo: I now serialize the datetime myself so I removed the call to `serialize()` and wrapped the array.
I added some questions in #49.

tsauerwein:
Please merge when you think it is ready.

gberaudo:
Thank you @tsauerwein and @asaunier for the review.

PR #100, opened 2015-12-10
Return last document title together with history

gberaudo:


c2corg_api/views/document.py
```diff
@@ -433,6 +433,12 @@ def get(self):
         #    # validate permission (authenticated + associated)
         #    # return 403 if not correct
 
+        title = DBSession.query(DocumentLocale.title) \
+            .filter(DocumentLocale.document_id == id) \
+            .filter(DocumentLocale.culture == lang) \
```
asaunier: Arf https://github.com/c2corg/v6_api/issues/101 :P

c2corg_api/views/document.py
```diff
@@ -433,6 +433,12 @@ def get(self):
...
+            .order_by(DocumentLocale.id.desc()) \
```
asaunier: Do you really need this? Isn't there only one version of the document in DocumentLocale?

asaunier:
Small comments else looks good to me.

c2corg_api/views/document.py
tsauerwein: Right. `DocumentLocale` only contains the latest version.

c2corg_api/views/document.py
```diff
@@ -433,6 +433,12 @@ def get(self):
         #    # validate permission (authenticated + associated)
         #    # return 403 if not correct
 
+        title = DBSession.query(DocumentLocale.title) \
```
tsauerwein: This query might fail when you are requesting the history for a language without entry.

asaunier: Good catch! The view should indeed raise an error if the requested language is not available.

gberaudo: Currently I think it raises an internal error. Do we want a 404?

asaunier: Yes

c2corg_api/views/document.py
gberaudo: Right, fixed.

c2corg_api/views/document.py
gberaudo: Done

gberaudo:
Thanks for the review @asaunier and @tsauerwein. I think I have addressed all comments.

asaunier:
:+1:

Issue #98, opened 2015-12-09
Login errors on the main dev instance


asaunier:
Logging in with an existing account (for instance "alex/c2c") fails with the following server error:
```
[Wed Dec 09 16:19:01.060912 2015] [wsgi:error] [pid 23485] [remote 10.27.71.4:5099]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_api/c2corg_api/views/user.py", line 127, in post
[Wed Dec 09 16:19:01.060973 2015] [wsgi:error] [pid 23485] [remote 10.27.71.4:5099]     token = try_login(user, password, request) if user else None
[Wed Dec 09 16:19:01.061123 2015] [wsgi:error] [pid 23485] [remote 10.27.71.4:5099]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_api/c2corg_api/security/roles.py", line 68, in try_login
[Wed Dec 09 16:19:01.061209 2015] [wsgi:error] [pid 23485] [remote 10.27.71.4:5099]     if user.validate_password(password, DBSession):
[Wed Dec 09 16:19:01.061246 2015] [wsgi:error] [pid 23485] [remote 10.27.71.4:5099]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_api/c2corg_api/models/user.py", line 69, in validate_password
[Wed Dec 09 16:19:01.061317 2015] [wsgi:error] [pid 23485] [remote 10.27.71.4:5099]     if PasswordUtil.is_password_valid(plain_password, self._password):
[Wed Dec 09 16:19:01.061347 2015] [wsgi:error] [pid 23485] [remote 10.27.71.4:5099]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_api/c2corg_api/models/user.py", line 29, in is_password_valid
[Wed Dec 09 16:19:01.061377 2015] [wsgi:error] [pid 23485] [remote 10.27.71.4:5099]     return bcrypt.hashpw(plain, encrypted) == encrypted
[Wed Dec 09 16:19:01.061411 2015] [wsgi:error] [pid 23485] [remote 10.27.71.4:5099]   File "/var/www/vhosts/c2cassoc_c2corgv6/private/v6_api/.build/venv/lib/python2.7/site-packages/bcrypt/__init__.py", line 66, in hashpw
[...]

PR #105, opened 2015-12-11
Main instance config directly include the makefile, not the dev config

asaunier:
The main instance is not a dev instance. It uses apache, the standard URL, etc.

This PR fixes the cors_origins problem mentioned in #98


gberaudo:
+1

PR #107, opened 2015-12-11
Add service to get document in a specific version

tsauerwein:
Request: "/waypoints/{id}/{lang}/{version_id}"
for example: http://localhost:6544/waypoints/642179/fr/1020298

Example response:

    {
        "document": {...},
        "version": {
            "username": "lazare",
            "comment": "Edit in fr",
            "user_id": 6206,
            "written_at": 1434325861,
            "version_id": 1020298
        }
    }

### DIFF
--- a/c2corg_api/views/document.py
+++ b/c2corg_api/views/document.py
@@ -9,13 +9,12 @@ from c2corg_api.models.document import (
 from c2corg_api.models.document_history import HistoryMetaData, DocumentVersion
 from c2corg_api.models.user import User
 from c2corg_api.search.sync import sync_search_index
-from c2corg_api.views import to_json_dict
+from c2corg_api.views import to_json_dict, to_seconds
 from c2corg_api.views.validation import check_required_fields, \
     check_duplicate_locales, validate_id, validate_lang
 
 from cornice.resource import resource, view
 from c2corg_api.views import cors_policy
-import datetime
 
 # the maximum number of documents that can be returned in a request
 LIMIT_MAX = 100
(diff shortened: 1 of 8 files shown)


asaunier:
Thanks! Looks good to me!

asaunier:
@gberaudo any objection?
If not, feel free to merge soon @tsauerwein this PR is needed for viewing an archived doc version as well.

asaunier:
> Request: "/waypoints/{id}/{lang}/{version_id}"

By the way I notice a little inconsistency: for getting an archived version we have a dedicated route with ``{lang}`` in matchdict whereas in the standard view, "lang" should be passed as a GET ``l`` param.

gberaudo:
postmerge +1

tsauerwein:
> By the way I notice a little inconsistency: for getting an archived version we have a dedicated route with {lang} in matchdict whereas in the standard view, "lang" should be passed as a GET l param.

Yes, good point.

c2corg_api/views/document.py
```diff
@@ -336,6 +338,40 @@ def _check_update_type(self, document, old_versions):
                 'trying do update the document with the same content')
         return (update_types, changed_langs)
 
+    def _get_version(self, clazz, schema, adapt_schema=None):
+        id = self.request.validated['id']
+        lang = self.request.validated['lang']
+        version_id = self.request.validated['version_id']
+        document = self._get_in_lang(
```
gberaudo: @tsauerwein are you sure the return document is the one form the requested version?
Shouldn't it be looked in the archives?

tsauerwein: You are right!

PR #108, opened 2015-12-14
Fix web-service to get version of a document

tsauerwein:
Now for real.

### DIFF
--- a/c2corg_api/models/document_history.py
+++ b/c2corg_api/models/document_history.py
@@ -1,3 +1,4 @@
+from c2corg_api.models.user import User
 from sqlalchemy import (
     Column,
     Integer,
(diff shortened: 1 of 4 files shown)


asaunier:
:+1:

gberaudo:
@tsauerwein, isn't it better to request only the data we will actually send to the client? I am wondering what is the performance implication of using the ORM.
Retrieving the archived documents takes three requests to the database. Wouldn't it work with a join?

tsauerwein:
I changed the the four queries into a huge query. I am wondering if that is better performance-wise.

Anyway, I am not particularly worried about the performance for this web-service, because I don't think it will be called that often.

> isn't it better to request only the data we will actually send to the client?

To load only specific columns we could use [load_only](http://docs.sqlalchemy.org/en/latest/orm/loading_columns.html#load-only-cols).

gberaudo:
Looks great to me, thanks.

tsauerwein:
Thanks for the review.

PR #111, opened 2015-12-15
Allow * in CORS

gberaudo:
CORS * allows any web client to send requests to the API.
The API will check for the token so:
- browsers sending an authorization token will get access to protected
  content;
- browsers not sending an authorization token will only get access to
  public content.

There is no security reason to maintain and restrict access to some
whitelisted domains.

When using cookies together with a CSRF token, the situation is
identical and using * is also secure.

### DIFF
--- a/c2corg_api/__init__.py
+++ b/c2corg_api/__init__.py
@@ -9,7 +9,6 @@ from c2corg_api.models import (
     Base,
     )
 from c2corg_api.search import configure_es_from_config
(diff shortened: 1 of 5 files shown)


gberaudo:
See discussion in #111.

tsauerwein:
Sounds good for me.

asaunier:
OK then. Thanks for cleaning the code.

gberaudo:
Thanks for the review.

PR #112, opened 2015-12-18
Add previous and next version id to version web-service

tsauerwein:
This is needed to generate the links "← previous difference" and "Next difference →" in the diff page.


asaunier:
Looks good to me.

gberaudo:
@tsauerwein, the queries are missing an `orderby` so that the returned versions are actually the next and previous ones.

tsauerwein:
I noticed it, I will do a follow-up PR. But sharp eyes! :+1:

Issue #38, opened 2015-10-09
Returning all attributes when creating a document?


asaunier:
As for now when creating a document, all attributes for all cultures are returned in the JSON response. Ditto when updating?

It might be useful for the mobile app (?) in order to display the full document after creating it.
In the UI, after creating, visitors are redirected to the document "view" page. For instance:
 http://www.camptocamp.org/routes/53123/fr/col-du-bonhomme-depuis-notre-dame-de-la-gorge

This page anyway makes an API call to get the document attributes.
In that case it is no use to return all attributes in the doc creation response (the doc id is enough).

Should we be able to add a parameter in the API call to tell if doc attributes must be returned in the creation response?

gberaudo:
+1 for not returning the attributes when not required. For example it should not be necessary for PUT methods since the client already has the latest version.

asaunier:
Is this issue still relevant?

tsauerwein:
> Is this issue still relevant?

Yes.

tsauerwein:
> Ditto when updating?

When you update a document, you always return to the view page?

asaunier:
> When you update a document, you always return to the view page?

Indeed as far as I know.

tsauerwein:
Ok, so it will be ok with the changes I made in #150.

PR #114, opened 2015-12-22
Allow updates without geometry

gberaudo:


gberaudo:
See https://github.com/c2corg/v6_ui/issues/24

gberaudo:
PR ready.

gberaudo:
@tsauerwein, @asaunier why do we return the document on PUT?
We could simplify the code and gain some performance by returning only the status code.

tsauerwein:
See also https://github.com/c2corg/v6_api/issues/38

I think we don't need to return the attributes again.

gberaudo:
I tried returning `{}` but it breaks the tests. If we agree it would be better to return nothing then I will create an issue so that we address this next year.

gberaudo:
(in a dedicated PR)

c2corg_api/models/document.py
```diff
@@ -304,6 +308,29 @@ def to_archive(self):
...
+        g2 = None
+        if isinstance(other.geom, geoalchemy2.WKBElement):
+            g2 = geoalchemy2.shape.to_shape(other.geom)
+        else:
+            # WKT are used in the tests.
+            str2 = string.split(other.geom, ';')[1]
+            g2 = wkt.loads(str2)
+
+        # FIXME: using almost_equals with +- 0.8m
+        # OK for EPSG:3857 but what about EPSG:4326?
```
tsauerwein: At this point we only should have 3857 OR 4326. Or is this not the case?

asaunier:
> I tried returning {}

Is it relevent?
If we decide that write actions do not need to return the document attributes (would make sense to me), the response should simply be something like:
```
{success: true}
```
or even nothing (just a 200 or other HTTP code)?

OK for creating the issue.

c2corg_api/models/document.py
asaunier: We should only have 3857 I think.
4326 or 21781 are only used on the UI side.

c2corg_api/tests/models/test_waypoint.py
```diff
@@ -104,6 +104,39 @@ def test_version_concurrent_edit(self):
         waypoint1.elevation = 2345
         self.assertRaises(StaleDataError, self.session.merge, waypoint1)
 
+    def test_geometry_update_optmimizations(self):
+        geom1 = 'SRID=3857;POINT(1.000000001 2.000000001)'
```
asaunier: Is it relevent 3857 values?

c2corg_api/tests/models/test_waypoint.py
```diff
@@ -104,6 +104,39 @@ def test_version_concurrent_edit(self):
         waypoint1.elevation = 2345
         self.assertRaises(StaleDataError, self.session.merge, waypoint1)
 
+    def test_geometry_update_optmimizations(self):
```
asaunier: optmimizations?

c2corg_api/models/document.py
gberaudo: I think the waypoints are in EPSG:3857 in the database / API and EPSG:4326 in the UI.
Will it be always like this? For all types of documents? If yes then I will change the FIXME by a comment.

asaunier: 4326 is only used in the UI forms (but then converted back to 3857 when writing the request, as done in your last PR) or in the view pages. Only 3857 is passed to the API.

asaunier: But you may be right: when we will upload GPX files, the provided coords will be in 4326.
But when drawing a line on the map, it will be 857.

Will we convert GPX tracks on the client side before pushing the geometry to the API?

gberaudo: OK, added a commit for computing the delta according to projection.

gberaudo:
I think I addressed all the comments and that the PR is now ready for merging.

tsauerwein:
Minor typo, otherwise it looks good.

gberaudo:
Fixed. Thanks for the reviews, will merge the PR after the holidays.

Issue #101, opened 2015-12-10
Lang/culture consistency


asaunier:
Sometimes we have "culture" in the source code or in the DB cols names. Sometimes "lang".
I feel a bit uncomfortable with this inconsistency. That's a bit confusing.

Shouldn't we grep/replace "culture" by "lang" or "language" everywhere in the API and the UI?

gberaudo:
+1 with a preference to `lang`.

tsauerwein:
+1 for "lang" as well.

asaunier:
OK, I suggest to wait after the steering committee before doing so :P

tsauerwein:
At the same time we can also change the `l=...` parameter to `lang=...`.

PR #115, opened 2015-12-22
Add language parameter to document list service

tsauerwein:
The preferred language can be given with `/waypoints?l=fr`.

### DIFF
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -343,26 +343,32 @@ def get_update_schema(document_schema):
     return UpdateSchema()
 
 
-def set_available_cultures(documents):
+def set_available_cultures(documents, loaded=False):
     """Load and set the available cultures for the given documents.
     """
     if len(documents) == 0:
         return
 
-    document_ids = [doc.document_id for doc in documents]
-    documents_for_id = {doc.document_id: doc for doc in documents}
+    if loaded:
+        # all locales are already loaded, so simply set the attribute
+        for document in documents:
+            document.available_cultures = [
+                locale.culture for locale in document.locales]
+    else:
+        document_ids = [doc.document_id for doc in documents]
+        documents_for_id = {doc.document_id: doc for doc in documents}
 
-    # aggregate the cultures per document into an array
-    culture_agg = func.array_agg(
-        DocumentLocale.culture,
-        type_=postgresql.ARRAY(String)).label('cultures')
+        # aggregate the cultures per document into an array
+        culture_agg = func.array_agg(
+            DocumentLocale.culture,
+            type_=postgresql.ARRAY(String)).label('cultures')
 
-    cultures_per_doc = DBSession.query(
-        DocumentLocale.document_id, culture_agg). \
-        filter(DocumentLocale.document_id.in_(document_ids)). \
-        group_by(DocumentLocale.document_id). \
-        all()
+        cultures_per_doc = DBSession.query(
+            DocumentLocale.document_id, culture_agg). \
+            filter(DocumentLocale.document_id.in_(document_ids)). \
+            group_by(DocumentLocale.document_id). \
+            all()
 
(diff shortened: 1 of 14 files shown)


tsauerwein:
I will also do the same for the search service.

asaunier:
> The preferred language can be given with /waypoints?l=fr

Just wondering if it would not be more consistent to use ``lang=fr`` instead of ``l=fr``?

asaunier:
Else looks good to me.

gberaudo:
Is there a possibility to return an array of the available langs so that the client knows which ones he can request?

tsauerwein:
> Just wondering if it would not be more consistent to use lang=fr instead of l=fr?

Currently the URL to get a waypoint is `/waypoints/1?l=en`. It would be nicer to use `/waypoints/1/fr` and `/waypoints/fr`, but we would have to define the Cornice resources differently. I can open an issue for that.

tsauerwein:
> Is there a possibility to return an array of the available langs so that the client knows which ones he can request?

There is this property "available_cultures" which should be set all (?) views.

asaunier:
> Currently the URL to get a waypoint is /waypoints/1?l=en. It would be nicer to use /waypoints/1/fr and /waypoints/fr, but we would have to define the Cornice resources differently.

Indeed. In the meantime I was just wondering if we should not used "lang" instead of "l" as lang parameter in API requests? Is there a reason for using "l"?

tsauerwein:
> Is there a reason for using "l"?

It's short. ;)

asaunier:
OK :)
Just asking because we use ``culture/lang/l" whereas we could simply use "lang" everywhere without having to wonder what is the correct parameter name :P

tsauerwein:
I added a note to https://github.com/c2corg/v6_api/issues/101.

asaunier:
> The preferred language can be given with /waypoints?l=fr.

Just to make sure I get it well: on the UI side, you send a ``/waypoints?l=<lang>`` request with ``<lang>``being the user's interface language. The API sends back a list of waypoint preferably with the ``<lang>`` locale, if the latter is not available it uses the next available locale in the lang priority list. Right?

tsauerwein:
Yes, right. Each document returned by `/waypoints?l=...` contains exactly one locale.

gberaudo:
It would be better to use something else than  `l` or `lang`, since the returned locale is not the one in parameter but the `best` one. It will help avoid confusion. What about something like `bestlang`?

asaunier:
@gberaudo 's objection makes sense. For the sake of simplicity and shortness, ``l`` (or later ``lang``) is OK for me but we could use ``pref_lang`` (since "best" is confusing :P).

tsauerwein:
> OK for me but we could use pref_lang (since "best" is confusing :P).

I am not really fan of these heavy names for GET parameters. I don't think that this is a problem if we document it in our not yet existing API documentation.

tsauerwein:
I pushed one more commit doing the same for the search web-service. Same for https://github.com/c2corg/v6_ui/pull/73

gberaudo:
Even if documented, the parameter is ambigous`and will cause confusion.

It also conflicts with a potential parameter for retrieving the document `in` a specific language.
It would be wise to keep `l/lang` for this later use case.

I am fine with `pref_lang` as proposed by @asaunier; it is "heavy" but it is an API-only parameter.
All that being said, I don't want to block the PR so feel free to do as you think best.

tsauerwein:
> Even if documented, the parameter is ambigous`and will cause confusion.
> 
> It also conflicts with a potential parameter for retrieving the document in a specific language.
> It would be wise to keep l/lang for this later use case.
> 
> I am fine with pref_lang as proposed by @asaunier; it is "heavy" but it is an API-only parameter.

I changed it to `?pl=..`, short but still different to `?l=..`.

gberaudo:
Great thanks @tsauerwein.

PR #118, opened 2016-01-07
Provide sso for discourse

gberaudo:
Try to read a discourse sso and sig from authentication request.
If present, validate and return the response in a redirect attribute.


tsauerwein:
Looks good to me!

gberaudo:
Thanks for the review.

PR #119, opened 2016-01-07
Return empty locales list instead of 404

tsauerwein:
... when requesting a not-existing locale of a document.

Closes https://github.com/c2corg/v6_ui/issues/10


asaunier:
Thanks, looks good!

Discussed with @tsauerwein : 
when translating an existing document using a new culture, we need to fill the editing form with at least the "figures" data even though the locales are not available yet.

tsauerwein:
Thanks.

PR #121, opened 2016-01-09
make migration datasource configurable

mfournier:


asaunier:
Thanks @mfournier, looks good to me!

tsauerwein:
Good idea!

PR #122, opened 2016-01-09
pull in gunicorn

mfournier:
The idea is to avoid using a different WSGI server between dev & prod,
and gunicorn has a few production-related features which lack from
waitress. Also, I'd like to drop apache in favor of haproxy.

First step: evaluate if gunicorn offers a valid alternative for both
use-cases by running
  `.build/venv/bin/gunicorn --paste development.ini -b :6543`
instead of
  `make serve`

Once we've validated the use of this tool, we can update the
`[server:main]` option and get rid of the legacy bits (waitress, apache,
etc).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ template: $(TEMPLATE_FILES)
 .PHONY: serve
 serve: install development.ini
 	echo "#\n# Also remember to start the ElasticSearch syncer script with:\n# make -f ... run-syncer\n#"
-	.build/venv/bin/gunicorn --paste --reload development.ini
+	.build/venv/bin/gunicorn --paste development.ini --reload
 
 .PHONY: run-syncer
 run-syncer: install development.ini
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ help:
 	@echo "- clean			Remove generated files"
 	@echo "- cleanall		Remove all the build artefacts"
 	@echo "- install		Install and build the project"
-	@echo "- serve			Run the development server (pserve)"
+	@echo "- serve			Run the development server"
 	@echo "- run-syncer		Run the ElasticSearch syncer script."
 	@echo "- template		Replace the config vars in the .in templates"
 	@echo
@@ -59,7 +59,7 @@ template: $(TEMPLATE_FILES)
 .PHONY: serve
 serve: install development.ini
 	echo "#\n# Also remember to start the ElasticSearch syncer script with:\n# make -f ... run-syncer\n#"
-	.build/venv/bin/pserve --reload development.ini
+	.build/venv/bin/gunicorn --paste --reload development.ini
 
 .PHONY: run-syncer
 run-syncer: install development.ini
--- a/development.ini.in
+++ b/development.ini.in
@@ -30,7 +30,7 @@ debugtoolbar.hosts = 0.0.0.0/0
 ###
 
 [server:main]
-use = egg:gunicorn#main
+use = egg:waitress#main
 host = 0.0.0.0
 port = {debug_port}
 
--- a/development.ini.in
+++ b/development.ini.in
@@ -30,7 +30,7 @@ debugtoolbar.hosts = 0.0.0.0/0
 ###
 
 [server:main]
-use = egg:waitress#main
+use = egg:gunicorn#main
 host = 0.0.0.0
 port = {debug_port}
 
--- a/production.ini.in
+++ b/production.ini.in
@@ -18,7 +18,7 @@ pyramid.includes =
 sqlalchemy.url = postgresql://{db_user}:{db_password}@{db_host}:{db_port}/{db_name}
 
 [server:main]
-use = egg:waitress#main
+use = egg:gunicorn#main
 host = 0.0.0.0
 port = 6543
 
--- a/production.ini.in
+++ b/production.ini.in
@@ -18,7 +18,7 @@ pyramid.includes =
 sqlalchemy.url = postgresql://{db_user}:{db_password}@{db_host}:{db_port}/{db_name}
 
 [server:main]
-use = egg:gunicorn#main
+use = egg:waitress#main
 host = 0.0.0.0
 port = 6543
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,7 +24,7 @@ kombu==3.0.35
 redis==2.10.5
 apscheduler==3.1.0
 requests==2.9.1
-gunicorn==19.4.5
+gunicorn==19.6.0
 
 # ColanderAlchemy>=0.3.2
 # needed for ColanderAlchemy: https://github.com/stefanofontanelli/ColanderAlchemy/pull/90
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ pyramid_tm==0.12.1
 SQLAlchemy==1.0.11
 transaction==1.4.4
 zope.sqlalchemy==0.7.6
-gunicorn==19.4.5
+waitress==0.8.10
 psycopg2==2.6.1
 colander==1.2
 GeoAlchemy2==0.2.6
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,6 +24,7 @@ kombu==3.0.35
 redis==2.10.5
 apscheduler==3.1.0
 requests==2.9.1
+gunicorn==19.4.5
 
 # ColanderAlchemy>=0.3.2
 # needed for ColanderAlchemy: https://github.com/stefanofontanelli/ColanderAlchemy/pull/90
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,7 +7,7 @@ pyramid_tm==0.12.1
 SQLAlchemy==1.0.11
 transaction==1.4.4
 zope.sqlalchemy==0.7.6
-waitress==0.8.10
+gunicorn==19.4.5
 psycopg2==2.6.1
 colander==1.2
 GeoAlchemy2==0.2.6


asaunier:
Thanks @mfournier that looks promising!

I have used your branch in my instance for both the UI and the API. I have played around and can see no problem with Gunicorn. It seems to work the same than with ``paster serve`` in development mode, changes in python files are directly taken into account.

Using Gunicorn in dev mode looks reasonable. It's not too risky to use it "instead" of paster. By the way doesn't it use/embed paster anyway? 
And it would be cool to use the same tools for dev and prod modes (but I guess it would not work exactly the same..).

I have not tested yet in prod mode. The choice seems to be apache+mod_wsgi vs haproxy+gunicorn. Has there known issues with haproxy+gunicorn? What about the perfs? features? others?

Do we need nginx as well? Or maybe not needed if haproxy is used?
Gunicorn seems to be able to serve static files as well but I guess that's not the best choice (nginx or haproxy are better for this I guess?).

Here are a few info about Gunicorn I have found:
https://www.digitalocean.com/community/tutorials/how-to-deploy-python-wsgi-apps-using-gunicorn-http-server-behind-nginx
https://www.digitalocean.com/community/tutorials/django-server-comparison-the-development-server-mod_wsgi-uwsgi-and-gunicorn

The latter article concludes that
> If you like Apache and a simple setup and integration then by all means use mod_wsgi. If you don't use Apache then I would recommend either uWSGI or Gunicorn. Both have their merits and will work great for any Django application. 

PS: if you push your branches to standard project origin, it's easier to test your PR since it's not needed to add a remote. :P

mfournier:
On Sat, Jan 09, 2016 at 08:41:05AM -0800, Alex Saunier wrote:
> 
> I have used your branch in my instance for both the UI and the API. I
> have played around and can see no problem with Gunicorn. It seems to work
> the same than with ``paster serve`` in development mode, changes in
> python files are directly taken into account.

Great, thanks !

> Using Gunicorn in dev mode looks reasonable. It's not too risky to use it
> "instead" of paster. By the way doesn't it use/embed paster anyway? 
> And it would be cool to use the same tools for dev and prod modes (but I
> guess it would not work exactly the same..).

My understanding is that we can continue using paster/pserve by simply
referring to gunicorn instead of waitress in `{development,production}.ini`
(cf http://docs.gunicorn.org/en/latest/configure.html#framework-settings):

```
[server:main]
use = egg:waitress#main
```

BTW, paster is some sort of development helper, right ? In production, I
assume we should call the gunicorn binary directly, as described here:
http://docs.gunicorn.org/en/latest/run.html

> I have not tested yet in prod mode. The choice seems to be
> apache+mod_wsgi vs haproxy+gunicorn. Has there known issues with
> haproxy+gunicorn? What about the perfs? features? others?

What drives me towards this setup is mainly the logging & monitoring
features haproxy offers & the configurability gunicorn has. Waitress also
seems a nice tool, but looking at the docs it doesn't seem to be as
ops-focused as gunicorn. Anyways, there are quite a few options out there,
so we can always reconsider if gunicorn doesn't fulfil the needs.

I'm not worried about performance regarding gunicorn vs mod_wsgi, more
about the apps running under their control, slow SQL queries, etc ;-)

> Do we need nginx as well? Or maybe not needed if haproxy is used?
> Gunicorn seems to be able to serve static files as well but I guess
> that's not the best choice (nginx or haproxy are better for this I
> guess?).

haproxy can't serve static files (and will live in an independent container
in any case). So yes, we'll need some sort of http server to take care of
this.

Which files are we talking about exactly ? Would it be possible to serve
them under a separate domain (s.camptocamp.org ?)


> PS: if you push your branches to standard project origin, it's easier to
> test your PR since it's not needed to add a remote. :P

Ah, I was wondering why I was the first one to fork the project ;-)

I have https://gist.github.com/piscisaureus/3342247 in my default git
settings and I warmly recommend everyone to do the same.

asaunier:
> BTW, paster is some sort of development helper, right ?

Yes. Not used in prod as far as I know.

> I'm not worried about performance regarding gunicorn vs mod_wsgi, more
about the apps running under their control, slow SQL queries, etc ;-)

So it seems there is no reason not using haproxy+gunicorn, is there?
Will this set up imply many changes in the application code? I guess we have to give it a try...

> haproxy can't serve static files (and will live in an independent container
in any case). So yes, we'll need some sort of http server to take care of
this.

What do you recommend?

> Which files are we talking about exactly ? Would it be possible to serve
them under a separate domain (s.camptocamp.org ?)

Pictos, photos, built CSS and JS files for instance (all in the UI, not in the API I think). Using a separate domain: maybe but at least it would imply changing parts of the current code. For instance: https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/templates/base.html#L74-L80
I don't know if it is possible to make ``request.static_url()`` use a given domain. We also need to make sure it is still easy to use the dev mode (without relying on complicated vhosts settings...).

Uploaded photos might already be in a dedicated vhost/server/bucket/???

tsauerwein:
I like the idea of having the same server for dev and prod. This hopefully avoids issues like https://github.com/c2corg/v6_ui/issues/20.


> My understanding is that we can continue using paster/pserve by simply referring to gunicorn instead of waitress in `{development,production}.ini` (cf http://docs.gunicorn.org/en/latest/configure.html#framework-settings): ``` [server:main] use = egg:waitress#main

Do you already want to do that change to make it more convenient to use gunicorn?

mfournier:
>> My understanding is that we can continue using paster/pserve by simply referring to gunicorn instead of waitress in {development,production}.ini (cf http://docs.gunicorn.org/en/latest/configure.html#framework-settings): ``` [server:main] use = egg:waitress#main

> Do you already want to do that change to make it more convenient to use gunicorn?

Done in 90fe769, let's see how this works !

tsauerwein:
> Done in 90fe769, let's see how this works !

I tested it with `make -f ... serve` but the "reload" makes the server crash: I make a change to a file, the server reloads. I make a 2nd change, and the server crashes. Also I am seeing a fat deprecated warning which tells me to use `gunicorn --paste development.ini`.

And when I do the following change to the Makefile, it works fine:

    .PHONY: serve
    serve: install development.ini
    	.build/venv/bin/gunicorn --paste development.ini --reload

On a first impression it seems like the reloading is faster than with pserve/waitress.

tsauerwein:
> I have https://gist.github.com/piscisaureus/3342247 in my default git settings and I warmly recommend everyone to do the same.

[hub](https://github.com/github/hub) is also quite nice.

asaunier:
Is this PR ready to be merged?
What about the crash mentioned at https://github.com/c2corg/v6_api/pull/122#issuecomment-170840119 ?

mfournier:
@asaunier I just updated the PR as @tsauerwein suggested, can you take a look ?

asaunier:
> @asaunier I just updated the PR as @tsauerwein suggested, can you take a look ?

@mfournier Could you rebase your branch? As for now it's too old and seems to not support python3 etc.

> I have https://gist.github.com/piscisaureus/3342247 in my default git settings and I warmly recommend everyone to do the same.

> hub is also quite nice.

Maybe, but why using all that stuff when it's so simple to push your branch to the main repo? For instance I cannot make changes to your branch in your fork or rebase it. Should I fork your fork in the project's repo? :P

mfournier:
Rebased on master.

> Maybe, but why using all that stuff when it's so simple to push your branch to the main repo?

External contributors typically don't have RW access to the reference repo, so I usually work this way by default, regardless of if I have privileges or not. I also like to avoid having a load of stall branches on the reference repo, because it leads to confusion.
But I'll try to do as you suggest for future contributions here if you prefer, no problem :)

tsauerwein:
Feel free to merge, Marc!

Makefile
```diff
@@ -61,7 +61,7 @@ install-dev-egg: $(SITE_PACKAGES)/c2corg_api.egg-link
 
 .PHONY: serve
 serve: install development.ini
-	.build/venv/bin/pserve --reload development.ini
+	.build/venv/bin/gunicorn --paste --reload development.ini
```
asaunier: This command is incorrect and cause the following error:
```
alex@c2corgv6-demo:~/v6_api$ .build/venv/bin/gunicorn --paste --reload development.ini
usage: gunicorn [OPTIONS] [APP_MODULE]
gunicorn: error: argument --paste/--paster: expected one argument
```
Using the following parameters order seems to work:
```
.build/venv/bin/gunicorn --paste development.ini --reload
```

asaunier:
Tested in my instance, there's a problem in the Makefile.
What about the demo instance (using apache as of today)? Will it work out of the box or do we need to install additional tools?

mfournier:
Not sure about apache & mod_wsgi. Let's keep this in hold a few more days @asaunier, and reconsider once we're ready to replace apache.

asaunier:
OK, it's important that the "apache" ("webserver"?) mode actually works, else the demo instance will be broken.

mfournier:
@asaunier I just rebased this branch against master & fixed the issue you mentioned. Can you give an update about the demo instance blocker ? I would like to have this merged soon to simplify setting up the project inside a docker container. Thanks !

asaunier:
@mfournier what do you mean? I could temporarly upgrade the demo instance with your branches to make sure it works in "apache" (sic) mode. Would that fit for you?

asaunier:
See my comments at https://github.com/c2corg/v6_ui/pull/83#issuecomment-228694384
It seems OK.

asaunier:
OK @mfournier please merge!

mfournier:
Thanks to all of you for the feedback & testing !

PR #124, opened 2016-01-10
Switch to Python 3.4

tsauerwein:
Python 2.7 will at least be supported until 2020, but we have no reason not to use Python 3.

@gberaudo, I had to do some changes to the authentication and SSO code. The tests pass, but I would appreciate if you could test if the integration with Discourse still works without problem. Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 language: python
 
 python:
-- 2.7
+- 3.4
 
 addons:
   postgresql: "9.4"
--- a/Makefile
+++ b/Makefile
@@ -81,7 +81,7 @@ upgrade-dev:
 
 .build/venv:
 	mkdir -p $(dir $@)
-	virtualenv --no-site-packages $@
+	virtualenv --no-site-packages -p python3 $@
 
 $(SITE_PACKAGES)/c2corg_api.egg-link: .build/venv requirements.txt setup.py
 	.build/venv/bin/pip install -r requirements.txt
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -14,14 +14,13 @@ from geoalchemy2 import Geometry
 import geoalchemy2
 from shapely import wkt
 from colander import MappingSchema, SchemaNode, String as ColanderString, null
-from itertools import ifilter
+
 import abc
 import enum
-import string
 
 from c2corg_api.models import Base, schema, DBSession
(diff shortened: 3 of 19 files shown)


asaunier:
Thanks! It really makes sense using Python3 instead of 2.7. The project should use up-to-date technologies as much as possible in order to make sure it is possible to make it evolve as easily as possible.

What are the main differences to expect when using Python3? According to your changes it seems to be:
* use of absolute module paths
* more transparent use of UTF8
* ?

tsauerwein:
Here is an overview of the key changes:
http://sebastianraschka.com/Articles/2014_python_2_3_key_diff.html

The change that made the most work for the conversion, is that a string `'abc'` is always a Unicode string in Python 3 (in Python 2 you had to prefix Unicode strings with `u'abc'`).

asaunier:
Thanks for the explanations.

By the way the last 3.x version is 3.5.1. What about using 3.5 instead of 3.4?

tsauerwein:
> By the way the last 3.x version is 3.5.1. What about using 3.5 instead of 3.4?

Python 3.5 was only released in September 2015, and right now there is no package in the official Ubuntu repositories. And I am not sure if we need one of the new features of 3.5: https://www.python.org/dev/peps/pep-0478/#id5

gberaudo:
@tsauerwein, I did a `make cleanall` then `make serve` and I get the following error
Successfully installed pyramid-closure c2corg-api
Cleaning up...
.build/venv/bin/pserve --reload development.ini
  File ".build/venv/bin/pserve", line 7, in <module>
    from pyramid.scripts.pserve import main
  File "/home/gberaudo/dev/c2cv6api/.build/venv/lib/python3.4/site-packages/pyramid/scripts/pserve.py", line 26, in <module>
    from paste.deploy import loadserver

Do you know about this error?

gberaudo:
I was able to go a bit further using pip install paste PasteDeploy Venusian. @tsauerwein I am doing something wrong?

tsauerwein:
> I was able to go a bit further using pip install paste PasteDeploy Venusian. @tsauerwein I am doing something wrong?

Doing `make cleanall`, `make -f ... install` and `make -f ... serve` worked for me.

gberaudo:
Finally, I cloned the repo from scratch and now it builds properly.
@tsauerwein, I tested the integration with discourse, it is working.
Great work, +1 for merging.

tsauerwein:
Nice, thanks for testing!

Issue #120, opened 2016-01-08
Special case of route names


asaunier:
Route names are a special case compared to other documents titles because they are closely related to the associated waypoints.

For instance http://www.camptocamp.org/routes/53781/fr/mont-blanc-arete-des-bosses
* displayed route name is "Mont Blanc : Arête des Bosses".
* the actual route name is "Arête des Bosses" though. The summit name is appended first in order to "discriminize" the route name, for instance when the route name itself is too common (eg. "Face N").

Several drawbacks with this approach:
* In v5 the displayed route name determination is really (too!) complicated: when several summits are associated to the route, the highest summit is used (based upon the summit elevation attribute). Then if the summit has several culture, the most appropriate one must be figured out depending on the interface language or user's preferences. This is done every time the route name must be displayed (even though some caching may be used in some cases).
* In the mapfile it is not possible to make that complicated computations and the route names displayed for vector features on the maps omit the summit name.
* In some cases, the highest summit is not the most relevent one (eg. a route to a hut that would go through some pass above the hut).
* In some cases omitting the summit name would make sense, eg. for raids.

With @tsauerwein we considered moving the summit name (if relevent) directly into the route title field. I have discussed it with the association and it would cause other problems. For instance when climbing sites are reorganised, if a bunch of routes are transfered from one site to another, one would have to change all the route names by hand. Ditto if a new culture is added to the summit with a different summit name (eg. Cervin/Cervino/Matterhorn). In addition contributors would most likely introduce inconsistencies in route names, for instance "Cervin : Face N", "Cervin Face N", "Cervin / Face N", "Face N du Cervin" or even "Face N" (summit name incorrectly omitted).

By the way having the summit name in the route name attribute would require to adapt the v5>v6 migration script.

Please note that in v6 using WP names (not only summit names) would make sense. For instance for hut-access routes, climbing route names including the name of a climbing site.

Suggested solution (to be precised and debated in the current issue):
* the route title attribute still contains only the actual route name (no WP name is embedded). As in v5.
* The displayed route name contains a WP name as well. As in v5. Except maybe for some route types (eg. raids).
* in the route detail page, it should be possible to specify the "main WP" among associated WPs. If only one WP, it is automatically considered as the main WP. If several WP are associated, the main WP would be the highest one by default.
[...]

Issue #90, opened 2015-11-27
Include associated waypoint title for routes in search index


tsauerwein:
To do once associations are implemented:

> Please note that "route" docs are special: we most likely need to get the routes associated to a given waypoint when searching the waypoint title. But the waypoint title is not embedded in the route name (especially when you have several WP associated to a route).
> 
> For instance "Mont Blanc":
> 
>     waypoint: http://www.camptocamp.org/summits/37355/fr/mont-blanc
>     routes:
>         http://www.camptocamp.org/routes/53781/fr/mont-blanc-arete-des-bosses => name in DB is "Arête des Bosses"
>         http://www.camptocamp.org/routes/45967/fr/mont-blanc-par-les-grands-mulets-itineraire-classique => name in DB is "Par les Grands Mulets - Itinéraire classique"
https://github.com/c2corg/v6_api/pull/82#issuecomment-160164212

stef74:
Hello,

Juste a comment ... on simple search title is sufficiant, but in advanced search we have probably to search an object depending an other. 
Ex : I want to search a route in Mont-Blanc area we have a WP type is summit and higher than 1000m

My idea is we have to manage some association in search (title for simple imple search other property for advanced search).

I don't know how you manage indexation (and i'm not specialist of ES). 

https://www.elastic.co/guide/en/elasticsearch/guide/current/relations.html
https://www.elastic.co/guide/en/elasticsearch/guide/current/parent-child.html

asaunier:
@stef74 I don't know if this is really related to the route title actually indexed by ES. I guess it is more an "advanced search" issue.

stef74:
@asaunier I'm agrre is not probably directly related to the issue ... but they can influence they to manage that.

PR #125, opened 2016-01-11
Add support for associations

tsauerwein:
* [x] Associations model
 * [x] Migrate associations
 * [x] Web-services to add/remove associations
 * [x] Return associations when requesting a document ([example resonse](http://pastebin.com/FZwrzDpc))
 * [x] Handle route names (see https://github.com/c2corg/v6_api/issues/120)
 * [ ] Handle geo-associations (see https://github.com/c2corg/v6_api/issues/128)
 * [ ] Show summit hierarchy
Example: http://www.camptocamp.org/summits/38641/fr/aiguille-du-peigne
c2corg-links

Closes https://github.com/c2corg/v6_api/issues/90
Closes https://github.com/c2corg/v6_api/issues/120

### DIFF
--- a/c2corg_api/models/__init__.py
+++ b/c2corg_api/models/__init__.py
@@ -26,3 +26,4 @@ from c2corg_api.models import route  # noqa
 from c2corg_api.models import document_history  # noqa
 from c2corg_api.models import image  # noqa
 from c2corg_api.models import user  # noqa
+from c2corg_api.models import association  # noqa
--- a/c2corg_api/models/association.py
+++ b/c2corg_api/models/association.py
@@ -1,4 +1,7 @@
+from c2corg_api.models.route import Route, RouteLocale
 from c2corg_api.models.user import User
+from c2corg_api.models.waypoint import Waypoint, WaypointLocale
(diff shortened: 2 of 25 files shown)


tsauerwein:
I think this would be ready for a review. I haven't started with geo-associations yet, I would prefer to do that in a separate PR, so that this one does not grow to big.

asaunier:
I will review the PR this afternoon I think.
Having a separated PR for geo associations makes sense.

asaunier:
Looks good to me.

I have posted a few comments about the fact that the PR assumes (?) that WP and routes will be possibly associated for any document type (which I think is incorrect) but perhaps the behaviour could be change when adding other document types?

Concerning ES, it seems that only main WPs may be used to search for a route. Not any associated WP, which is really useful when you don't know what the main WP is or want to find all routes associated to a given WP. To do so we could not rely only on adding the associated WP names in the route titles in the ES index. This could be discussed in a separated issue, not preventing the current PR from being merged.

tsauerwein:
Thanks for the review, Alex!

Issue #127, opened 2016-01-12
Document type is not set when inserting/updating a document in ElasticSearch


tsauerwein:
The document is inserted/updated in ES, but `doc_type` is not set.

Related to https://github.com/c2corg/v6_ui/issues/92

PR #129, opened 2016-01-12
Make sure the discriminator column is set

tsauerwein:
Due to a wrong SQLAlchemy mapping the values for the inheritance discriminator column were not set (e.g. `type = 'w'` for waypoints).

Closes https://github.com/c2corg/v6_ui/issues/92
Closes https://github.com/c2corg/v6_api/issues/127


asaunier:
Looks good to me.

PR #131, opened 2016-01-14
Fix the version of v6_common in requirements

tsauerwein:
Currently I can't push a change to the common project that I need for https://github.com/c2corg/v6_api/pull/125 because it would break master and all installations.


asaunier:
OK

PR #134, opened 2016-01-18
Login to discourse directly from backend

gberaudo:
When the discourse nonce is not present in the request, the backend will
send a request to the forum and generate an authentication URL.

This URL must be sent from the client side (in the browser) to
allow correct setting of discourse cookies.

See https://github.com/c2corg/v6_forum/issues/5


gberaudo:
Added logout from discourse when logging out of the (backend) API.

gberaudo:
Note about logout:
    For this to work, the user ids in discourse MUST be the same as in the
    main UI. We need to make sure v5 users are migrated in the right order
    and without gap and that new users are added _sequentially_.
Is it possible/reasonable to guarantee no two users are added at the same?

tsauerwein:
> Note about logout:
> For this to work, the user ids in discourse MUST be the same as in the
> main UI. We need to make sure v5 users are migrated in the right order
> and without gap and that new users are added sequentially.
> 
> Is it possible/reasonable to guarantee no two users are added at the same?

This would work for migrated users. But I think we can not guarantee the order when user accounts are created later on.

What's the idea about creating the user accounts for the forum (for accounts that are created after the migration)? Do we create the forum user account when the main account is created? Or do we delay the creation of the forum user account until the user visits the forum? For the first case, we could try to use a method like [DiscourseClient.createUser](https://github.com/gberaudo/pydiscourse/commit/aac961d6e22578ab2e2fa66876fd6178c827d2ae#diff-0b54ff8984472a0c8d41e7d4e8ec5695R26). I don't know if there an equivalent call for SSO users? And I assume this API call returns the used user id? We should store this user id and use it for the logout.

tsauerwein:
> Search users by external_id
> 
> User profile data can be accessed using the /users/by-external/{EXTERNAL_ID}.json endpoint. This will return a JSON payload that contains the user information, including the user_id which can be used with the log_out endpoint.
https://meta.discourse.org/t/official-single-sign-on-for-discourse/13045

We could use this to get the Discourse user id. Only once and then store it in our model.

asaunier:
Seems a nice job to me but don't forget to document (in a README or...) how login/logout works and why this strategy has been chosen.

c2corg_api/security/discourse_sso_provider.py
```diff
@@ -11,25 +11,18 @@
 import logging
 log = logging.getLogger(__name__)
 
+# 1 second timeout for getting sso information from discourse
+TIMEOUT = 1
```
tsauerwein: Do we need two different timeout constants? There is also `CLIENT_TIMEOUT`.

gberaudo: Right, passing the timeout as parameter.

gberaudo:
> Seems a nice job to me but don't forget to document (in a README or...) how login/logout works and why this strategy has been chosen.

Added a link to the README.

gberaudo:
@tsauerwein, I now supports the case when the user id and the discourse user id are different.
A memory cache is used to avoid doing extra requests.

@asaunier, I think the only point left before merging is to move the pydiscourse fork to
the c2corg organization. Could you please create a fork there?

asaunier:
Thanks @gberaudo this looks good to me.

> I think the only point left before merging is to move the pydiscourse fork to the c2corg organization. Could you please create a fork there?

Done: https://github.com/c2corg/pydiscourse

gberaudo:
Thanks @asaunier. Will merge when Travis build completes.

Issue #128, opened 2016-01-12
Geo-associations


asaunier:
I paste here info sent by email to @tsauerwein a few days ago.

Geoassociations are associations between usual documents (WP, routes, outings, images, etc.) with region documents (http://www.camptocamp.org/areas => countries, CH cantons / F departements and ranges/massifs) and in some cases "paper maps" (http://www.camptocamp.org/maps)

* It's not worth spending too much time about maps. In v5 it's a kind of document but it's not that important. I don't think they are often used. The goal was to automatically list the maps without having to type the map names in a text field, at least in France and Switzerland where we had the geometries.
Not sure we even need to have a "map" document type in v6: if we can provide the list of maps (even without linking toward a detail page for the map) intersecting the current page geometry I guess it's enough.
Please note that we don't show linked maps in a document lists => the associations could be computed on the fly when showing a document if it has at least a default geometry.

* regions are more important since they can be used to filter documents (eg. "show me outings in Mont-Blanc range"). Regions associations may be needed in many pages (detail page, list page) for almost all doc types.
In v5 even books may be linked to regions (based upon linked routes)! For instance: http://www.camptocamp.org/books/704165/fr/scialpinismo-in-svizzera
To be honest, I wonder if it is really needed. Of course it's useful to get the list of guidebooks for a given region. But you don't do that kind of search very often.

It is worth figuring out if we stick to table-based geo associations (as in v5) or if we can use geographical operations by having a associated-docs-based default geometry when the document has no own geometry. The latter solution would be more flexible I think and would allow to make proximity searching or box/polygon filtering for all documents. It would also be needed (required?) for map-extent based filtering in search pages (see wireframes about "topoguide"/"page de recherche").

I wonder if we would not need a "point" default geometry for all documents, including the routes and outings. It would be useful for instance to:
* make map filtering (advanced search)
* show outings on the map: even if we have line geometries, it is probably more usable/performant to show markers rather than a bunch of lines? especially if all features have no own geometry. That probaby can be challenged since in some cases (routes in a WP page for instance) it's interesting to show the actual lines.
* make region filtering (not only a map extent filtering) => I guess it's more performant to get a list of points rather than lines when intersectiong with a polygons or even bboxes. Do you think we could replace the current v5 geo-associations filtering by point-geom-based intersections? Would it be too performances-hurting?

asaunier:
[...]

PR #136, opened 2016-01-21
Geo-associations

tsauerwein:
This PR deals with geo-associations, the associations between maps, areas/regions and documents:

 * Adds the model classes for maps and areas, updates the migration script and adds web-services
 * The associations between maps and documents are not persisted. When requesting a document, the relevant maps are retrieved with a spatial query.
 * On the contrary, the associations between areas and documents are persisted in a table `area_associations`, because this information is needed also in listings. This table is updated when a new area or document is added, or the geometry of an area or document is updated.
 * Maps and associations are returned when requesting a single document.

TODO

 * [x] Include areas in listing views
 * [x] Everybody can edit title and description of an area, currently the geometry should not be editable. (-> only moderators can edit geometries)
 * [x] Currently no new areas can be created ( -> limited to moderators).

See https://github.com/c2corg/v6_api/issues/128

### DIFF
--- a/c2corg_api/models/__init__.py
+++ b/c2corg_api/models/__init__.py
@@ -29,3 +29,4 @@ from c2corg_api.models import user  # noqa
 from c2corg_api.models import association  # noqa
 from c2corg_api.models import topo_map  # noqa
 from c2corg_api.models import area  # noqa
+from c2corg_api.models import area_association  # noqa
--- a/c2corg_api/models/__init__.py
+++ b/c2corg_api/models/__init__.py
@@ -27,3 +27,5 @@ from c2corg_api.models import document_history  # noqa
 from c2corg_api.models import image  # noqa
 from c2corg_api.models import user  # noqa
 from c2corg_api.models import association  # noqa
+from c2corg_api.models import topo_map  # noqa
+from c2corg_api.models import area  # noqa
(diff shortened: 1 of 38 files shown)


tsauerwein:
I forgot to include the areas in listing views. But besides it's ready to review.

asaunier:
Thanks for the huge and nice work!
As far as I can see, it looks good.

tsauerwein:
Ok, this should be ready. Areas can now only be created by moderators. Everybody can update an area, but only moderators can change the geometry of an area. Even if we don't let moderators update an area in the UI, at least it is supported by the API.

asaunier:
Cool, thanks a lot!
Looks good to me.

What about geoassociations of documents lacking their own geometry (for instance routes with no GPS track)?
Do they inherit the geoassociations from the associated documents (like in v5)? Note sure I have seen such a behaviour in the PR.
Or will we compute the geoassociations using an "inherited" default geometry from associated docs (as suggested in https://github.com/c2corg/v6_api/issues/128)? Do you think this is realistic?

When migrating v5 geoassociations do you also migrate those of geom-lacking docs?

tsauerwein:
> What about geoassociations of documents lacking their own geometry (for instance routes with no GPS track)?

This PR does not deal with that. I opened a new issue for that: https://github.com/c2corg/v6_api/issues/144

tsauerwein:
Ok for merging?

asaunier:
Please do :+1:

tsauerwein:
Thanks!

PR #139, opened 2016-02-03
Simplify route title building

asaunier:
We can assume that the separator between ``title_prefix`` and ``title`` is always `` : `` whatever the language.

### DIFF
--- a/c2corg_api/scripts/es/fill_index.py
+++ b/c2corg_api/scripts/es/fill_index.py
@@ -91,7 +91,7 @@ def fill_index(db_session):
                 }
 
             search_document['title_' + culture] = get_title(
-                title, title_prefix, culture)
+                title, title_prefix)
             search_document['summary_' + culture] = strip_bbcodes(summary)
             search_document['description_' + culture] = \
                 strip_bbcodes(description)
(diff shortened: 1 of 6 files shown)


tsauerwein:
:+1:

tsauerwein:
I think you'll have to adapt some tests.

asaunier:
Yes :) In progress

gberaudo:
https://fr.wikipedia.org/wiki/Ponctuation#Traitement_des_espaces_pour_les_symboles_bipoints

asaunier:
> https://fr.wikipedia.org/wiki/Ponctuation#Traitement_des_espaces_pour_les_symboles_bipoints

Yes. The goal here is to have a language-independent separator since we want to simplify this part. ":" is not fully language-independent but that's the best candidate and that's not that a big deal if route titles have an extra space before the colon.

asaunier:
> I think you'll have to adapt some tests.

Done at 02f49f

Actually I have only replaced ":" by " :" in the "en" locales. Not sure if it was relevant to change some tests, for instance because we no longer have a difference in the route title whatever the culture.

asaunier:
If no one objects I will rebase the fixup and merge.

tsauerwein:
No objections from my side.

gberaudo:
+1

PR #146, opened 2016-02-05
Update python dependencies

gberaudo:
Except elasticsearch and elasticsearch_dsl.


gberaudo:
See https://github.com/c2corg/v6_ui/issues/139.

tsauerwein:
Thanks, that looks good! :+1:

asaunier:
:+1: as well
Would be cool to merge only next week, I would like to update the demo instance shortly (just in case we have side-effects)

PR #148, opened 2016-02-08
Rename API's 'l' parameter to 'lang'

asaunier:
See https://github.com/c2corg/v6_api/issues/101#issuecomment-166672389

Should we rename the "pl" parameter as well?


asaunier:
Not sure there's not something missing, I have not been able to test getting a given lang for a multi-lang document (ES is down...).

PR #150, opened 2016-02-10
Do not return full document on create/update

tsauerwein:
Create only returns the `document_id` (`{'document_id': ...}`), update an empty object (`{}`).

Closes https://github.com/c2corg/v6_api/issues/38

### DIFF
--- a/c2corg_api/views/area.py
+++ b/c2corg_api/views/area.py
@@ -32,7 +32,7 @@ class AreaRest(DocumentRest):
             permission='moderator')
     def collection_post(self):
         return self._collection_post(
-            Area, schema_area, after_add=insert_associations)
+            schema_area, after_add=insert_associations)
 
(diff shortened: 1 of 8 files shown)


c2corg_api/views/area.py
```diff
@@ -32,7 +32,7 @@ def get(self):
             permission='moderator')
     def collection_post(self):
         return self._collection_post(
-            Area, schema_area, after_add=insert_associations)
```
gberaudo: Is this line a fix to a different issue?

tsauerwein: Not a fix, a small clean-up. The class was not used in `_collection_post`, that's why I removed it.

gberaudo:
LGTM

tsauerwein:
Thanks for the review.

PR #153, opened 2016-02-19
Explicily define inherit_condition in models

tsauerwein:
This was required for a query using routes, but to avoid the same problem in the future, it makes sense to do it for all models.

See https://github.com/c2corg/v6_api/commit/03c5e9cc1ab505addf9f48e9d45901834a371ed7#commitcomment-16181282


asaunier:
Looks good thanks!

tsauerwein:
Thanks for the review.

PR #154, opened 2016-02-19
Add images support

### LINKED ISSUE #151: TODO images API
* [x] Make sure all attributes are available in the model See https://github.com/c2corg/v6_api/pull/154
* [x] Make sure activities attribute can contain several values (see https://github.com/c2corg/v6_api/issues/45)
* [x] Create v5>v6 migration script See https://github.com/c2corg/v6_api/pull/154
* [x] Support file upload 
* [x] Image storage See https://github.com/c2corg/v6_api/issues/167
* [ ] Process uploaded files (create thumbnails, etc.)  See #168
* [x] Index images with ES

asaunier:
TODO
* [x] Add all attributes to Image model
* [x] Migration script
* [x] Fix unittests
* [ ] ~~Add image upload support to view~~
* [ ] ~~Image handling (thumbnails etc.)~~
* [x] Associations
* [x] ES index

Related to #151

### DIFF
--- a/c2corg_api/models/image.py
+++ b/c2corg_api/models/image.py
@@ -1,33 +1,66 @@
 from sqlalchemy import (
+    Boolean,
     Column,
+    DateTime,
+    Float,
+    ForeignKey,
     Integer,
     SmallInteger,
-    ForeignKey,
-    Enum
+    String
     )
 
 from colanderalchemy import SQLAlchemySchemaNode
 
-from c2corg_api.models import schema
-from c2corg_api.models.utils import copy_attributes
+from c2corg_api.models import schema, enums
+from c2corg_api.models.utils import copy_attributes, ArrayOfEnum
 from c2corg_api.models.document import (
     ArchiveDocument, Document, get_update_schema, geometry_schema_overrides,
     schema_document_locale, schema_attributes)
-from c2corg_common.attributes import activities
 from c2corg_common import document_types
 
 IMAGE_TYPE = document_types.IMAGE_TYPE
 
 
 class _ImageMixin(object):
-    activities = Column(
-        Enum(name='activities', inherit_schema=True, *activities),
-        nullable=False)
+
+    activities = Column(ArrayOfEnum(enums.activity_type))
+
+    categories = Column(ArrayOfEnum(enums.image_category))
+
+    image_type = Column(enums.image_type)
+
+    author = Column(String(100))
+
+    has_svg = Column(Boolean)
+
+    elevation = Column(SmallInteger)
 
     height = Column(SmallInteger)
 
+    width = Column(SmallInteger)
 
-attributes = ['activities', 'height']
+    file_size = Column(Integer)
+
+    filename = Column(String(30))
+
+    date_time = Column(DateTime)
+
+    camera_name = Column(String(100))
+
+    exposure_time = Column(Float)
+
+    focal_length = Column(Float)
+
+    fnumber = Column(Float)
+
+    iso_speed = Column(SmallInteger)
+
+
+attributes = [
+    'activities', 'categories', 'image_type', 'author', 'has_svg', 'elevation',
+    'height', 'width', 'file_size', 'filename', 'camera_name', 'exposure_time',
+    'focal_length', 'fnumber', 'iso_speed', 'date_time'
+]
 
 
 class Image(_ImageMixin, Document):
--- a/c2corg_api/views/image.py
+++ b/c2corg_api/views/image.py
@@ -1,10 +1,15 @@
+from c2corg_api.models.image import Image, schema_image, schema_update_image
+from c2corg_common.fields_image import fields_image
 from cornice.resource import resource, view
 
-from c2corg_api.models.image import Image, schema_image, schema_update_image
-from c2corg_api.views.document import DocumentRest
+from c2corg_api.views.document import DocumentRest, make_validator_create, \
+    make_validator_update
 from c2corg_api.views import cors_policy, restricted_json_view
 from c2corg_api.views.validation import validate_id, validate_pagination, \
     validate_lang_param, validate_preferred_lang_param
+
+validate_image_create = make_validator_create(fields_image.get('required'))
+validate_image_update = make_validator_update(fields_image.get('required'))
 
 
 @resource(collection_path='/images', path='/images/{id}',
@@ -19,10 +24,15 @@ class ImageRest(DocumentRest):
     def get(self):
         return self._get(Image, schema_image)
 
-    @restricted_json_view(schema=schema_image)
+    @restricted_json_view(
+            schema=schema_image, validators=validate_image_create)
     def collection_post(self):
         return self._collection_post(schema_image)
 
-    @restricted_json_view(schema=schema_update_image, validators=validate_id)
+    @restricted_json_view(
+            schema=schema_update_image,
+            validators=[validate_id, validate_image_update])
     def put(self):
+        # FIXME: personal images should only be modifiable by
+        # their creator and moderators
         return self._put(Image, schema_image)
--- a/c2corg_api/views/image.py
+++ b/c2corg_api/views/image.py
@@ -1,3 +1,5 @@
+from c2corg_api.models import DBSession
+from c2corg_api.models.document_history import has_been_created_by
 from c2corg_api.models.image import Image, schema_image, schema_update_image
 from c2corg_common.fields_image import fields_image
 from cornice.resource import resource, view
@@ -7,6 +9,8 @@ from c2corg_api.views.document import DocumentRest, make_validator_create, \
 from c2corg_api.views import cors_policy, restricted_json_view
 from c2corg_api.views.validation import validate_id, validate_pagination, \
     validate_lang_param, validate_preferred_lang_param
+
+from pyramid.httpexceptions import HTTPForbidden, HTTPNotFound, HTTPBadRequest
 
 validate_image_create = make_validator_create(fields_image.get('required'))
 validate_image_update = make_validator_update(fields_image.get('required'))
@@ -33,6 +37,20 @@ class ImageRest(DocumentRest):
             schema=schema_update_image,
             validators=[validate_id, validate_image_update])
     def put(self):
-        # FIXME: personal images should only be modifiable by
-        # their creator and moderators
+        if not self.request.has_permission('moderator'):
+            image_id = self.request.validated['id']
+            image = DBSession.query(Image).get(image_id)
+            if image is None:
+                raise HTTPNotFound('No image found for id %d' % image_id)
+            if image.image_type == 'collaborative':
+                image_type = self.request.validated['document']['image_type']
+                if image_type != image.image_type:
+                    raise HTTPBadRequest(
+                        'Image type cannot be changed for collaborative images'
+                    )
+            # personal images should only be modifiable by
+            # their creator and moderators
+            elif not has_been_created_by(image_id,
+                                         self.request.authenticated_userid):
+                raise HTTPForbidden('No permission to change this image')
         return self._put(Image, schema_image)
--- a/c2corg_api/models/document_history.py
+++ b/c2corg_api/models/document_history.py
@@ -7,9 +7,10 @@ from sqlalchemy import (
     ForeignKey
     )
 from sqlalchemy.orm import relationship, backref
+from sqlalchemy.sql.expression import exists, and_
 import datetime
 
-from c2corg_api.models import Base, schema, users_schema
+from c2corg_api.models import Base, DBSession, schema, users_schema
 from c2corg_api.models.document import (
     Document, ArchiveDocument, ArchiveDocumentLocale, ArchiveDocumentGeometry)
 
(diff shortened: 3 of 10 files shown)


c2corg_api/models/image.py
```diff
@@ -1,32 +1,63 @@
...
+
+    activities = Column(ArrayOfEnum(enums.activity_type))
+
+    categories = Column(ArrayOfEnum(enums.image_category))
+
+    image_type = Column(enums.image_type)
+
+    author = Column(String(100))
+
+    # FIXME not sure still needed:
```
tsauerwein: why was it needed?

asaunier: Because in v5 it is possible to upload an SVG. For instance for schemas or images with drawn routes. See for instance: http://www.camptocamp.org/images/724568/fr/roletta-couloir-nw
In that case the SVG is automatically converted to a JPEG file on the server, the latter file being then displayed instead of the SVG. A link "SVG file" is then available in the image page to download the original file.

In v6, since we no longer support old browsers, I think we will be able to display the SVG file directly? What about the resized/cropped versions?

tsauerwein: > In v6, since we no longer support old browsers, I think we will be able to display the SVG file directly? 

Yes, we should be able to directly use the SVG files. 

> What about the resized/cropped versions?

You mean if we also should create thumbnail versions for SVG files? If I look at the example file that you have linked to, it would make sense.

asaunier: I guess we still need the "has_svg" flag then, at least to figure out when to display the original SVG and when to display the converted/resized version. Unless we can have resized/cropped SVG versions?

Would be cool but that would mean that all SVG files from v5 must be resized/cropped when migrating the images to v6.

asaunier:
I think I'm done with the migration script for images. I just don't understand if I have to do something to migrate the associations for images or if they should already be migrated by ``MigrateAssociations`` https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/documents/associations.py

I have picked a few ``document_id`` of the migrated images entries but cannot find them in the ``guidebook.associations`` table?

tsauerwein:
As you've added `app_images_archives` to the list of [tables](https://github.com/c2corg/v6_api/pull/154/files#diff-5c239002d1caf0de0db0bf2e73f9a04d) it *should* be fine. You are sure that the images that you checked do have associations?

asaunier:
> You are sure that the images that you checked do have associations?

Maybe you're right. At least I have tried with images who do have associations such as http://www.camptocamp.org/images/576289/fr/miaz-gcb-elle-pointe-son-nez-perce and the associated documents in v5 are actually linked in my v6 instance:
```
c2corg_alex=# select * from guidebook.associations where child_document_id = 576289;
 parent_document_id | child_document_id 
--------------------+-------------------
              40128 |            576289
              49572 |            576289
             177763 |            576289
             470801 |            576289
             541304 |            576289
(5 rows)
```
The outing is missing but that's normal since your PR importing outings is not merged yet :)

asaunier:
> TODO
* ES index

Do we need to do something explicitely to get the new documents indexed when running the ES index command or does ES already index all available documents?

Concerning the unittests I struggle figuring out what is going wrong:

> FAIL: Test updating a document by adding a new locale.
  File "/home/alex/v6_api/c2corg_api/tests/views/test_image.py", line 297, in test_put_success_new_lang
    (body, image) = self.put_success_new_lang(body, self.image)
  File "/home/alex/v6_api/c2corg_api/tests/views/__init__.py", line 786, in put_success_new_lang
    self.assertEquals(body.get('version'), document.version)
nose.proxy.AssertionError: 2 != 1

tsauerwein:
Can you try to rebase? https://github.com/c2corg/v6_api/commit/952cbe0a40ca3c1f42d424f1a391adf0f8bb71f2 should have fixed that issue.

asaunier:
I have just rebased it 10 min ago and pushed the branch back to github...

tsauerwein:
Ok, I see: The [test-case](https://github.com/c2corg/v6_api/blob/images/c2corg_api/tests/views/test_image.py#L280-L299) checks adding a new locale while the figures stay the same. You are also changing the figures (image_type): https://github.com/c2corg/v6_api/blob/images/c2corg_api/tests/views/test_image.py#L289

asaunier:
I have fixed some of the testing problems.
But tests still say that the version (ie. figures) of the document has changed for both following tests. I don't get where the figures have changed...

> FAIL: test_put_success_lang_only (c2corg_api.tests.views.test_image.TestImageRest)
  File "/home/travis/build/c2corg/v6_api/c2corg_api/tests/views/test_image.py", line 276, in test_put_success_lang_only
    (body, image) = self.put_success_lang_only(body, self.image)
  File "/home/travis/build/c2corg/v6_api/c2corg_api/tests/views/__init__.py", line 703, in put_success_lang_only
    self.assertEquals(body.get('version'), document.version)
nose.proxy.AssertionError: 2 != 1

> FAIL: Test updating a document by adding a new locale.
  File "/home/travis/build/c2corg/v6_api/c2corg_api/tests/views/test_image.py", line 298, in test_put_success_new_lang
    (body, image) = self.put_success_new_lang(body, self.image)
  File "/home/travis/build/c2corg/v6_api/c2corg_api/tests/views/__init__.py", line 786, in put_success_new_lang
    self.assertEquals(body.get('version'), document.version)
nose.proxy.AssertionError: 2 != 1

tsauerwein:
Your are updating [this image](https://github.com/c2corg/v6_api/blob/images/c2corg_api/tests/views/test_image.py#L315-L322):

        self.image = Image(
            activities=['paragliding'], height=1500)

In both tests that fail you are setting 'image_type' to 'collaborative', thus changing the figures:
https://github.com/c2corg/v6_api/blob/images/c2corg_api/tests/views/test_image.py#L267

asaunier:
OK, thanks for the explanations!
Now I get
>     body.get('version'), document.version)
nose.proxy.AssertionError: 1 == 1

:P

tsauerwein:
Which makes sense. Now that you set `image_type='collaborative'` on your base image, you will have to change the value for `test_put_success_figures_only` and `test_put_success_all` which expect that the figures change. You could for example set `'height': 1600,` in the request body of these two.

asaunier:
> you will have to change the value for test_put_success_figures_only and test_put_success_all which expect that the figures change.

Doh! I had not paid attention that the failed tests had changed, since I was used to have problems with ``test_put_success_lang_only`` etc. Sorry!

I have fixed the remaining tests, made sure the migration works (takes a while now with all the documents to import) and that v5 associations have been migrated to v6 as well. It seems the images are indexed by ES as well. The commits have been rebased.

Images uploading and manipulating should be addressed in some other PR, I have striked them from the TODO list.

Are there other tasks that must be performed before merging the PR?
Thanks!

gberaudo:
Added a comment about a FIXME. Looks like a great work @asaunier!

tsauerwein:
I am planning to do a review today.

tsauerwein:
There is still the open question about how the owner of an image is stored, but besides it's looking good.

asaunier:
> about how the owner of an image is stored

In v5 the owner ("creator") is the user bound to the initial version (whatever the language) of the document:
https://github.com/c2corg/camptocamp.org/blob/master/apps/frontend/modules/images/actions/actions.class.php#L22-L23
https://github.com/c2corg/camptocamp.org/blob/dacf3df6fc7d300df59478b857cecd5115616d79/lib/model/doctrine/generated/BaseDocument.class.php#L2454-L2478

Would it make sense to create a ``_get_creator`` method in the ``Document`` model (or in ``DocumentVersion``)?

asaunier:
I have added a permission check when updating an image: https://github.com/c2corg/v6_api/commit/9ac5180f9e6408da0de5b8605d93679bc72eb88b

It is still work in progress. Test ``test_put_wrong_document_id`` failed, so I have added an HTTPNotFound check. There is still a problem with ``test_put_wrong_ids``.
Some tests about editing personal images with or without the correct user (is it possible to test with several users?) might be relevant as well.

tsauerwein:
> There is still a problem with test_put_wrong_ids.

You can set the user:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/tests/views/test_outing.py#L430

> Some tests about editing personal images with or without the correct user (is it possible to test with several users?) might be relevant as well.

Check: https://github.com/c2corg/v6_api/blob/master/c2corg_api/tests/views/test_outing.py#L435-L489

asaunier:
Thanks @tsauerwein for your help!
I have added ``test_put_wrong_user`` and ``test_put_good_user`` tests.

I think we should also prevent users from switching a collaborative image to some other ``image_type``:
```
    @restricted_json_view(
            schema=schema_update_image,
            validators=[validate_id, validate_image_update])
    def put(self):
        if not self.request.has_permission('moderator'):
            image_id = self.request.validated['id']
            image = DBSession.query(Image).get(image_id)
            if image is None:
                raise HTTPNotFound('No image found for id %d' % image_id)
            if image.image_type == 'collaborative':
                print('IMAGE_TYPE', self.request.validated.get('image_type'))
                if self.request.validated.get('image_type') != 'collaborative':
                    raise HTTPBadRequest(
                        'Image type cannot be changed for collaborative images'
                    )   
            # personal images should only be modifiable by
            # their creator and moderators
            elif not self._is_creator(
                 self.request.authenticated_userid,
                 image_id):
                raise HTTPForbidden('No permission to change this image')
        return self._put(Image, schema_image)
```
but I don't find a way to get the updated ``image_type``. ``self.request.validated.get('image_type')`` seems to return ``None``.
According to http://cornice.readthedocs.org/en/latest/tutorial.html#users-management 
> Validators are filling the request.validated mapping

Is this related to https://github.com/c2corg/v6_api/blob/images/c2corg_api/views/image.py#L17 ?
If yes ``image_type`` is indeed a required field: https://github.com/c2corg/v6_common/blob/image-geom-not-required/c2corg_common/fields_image.py#L26

Should I do something like https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/outing.py#L231-L241 ?

tsauerwein:
> print('IMAGE_TYPE', self.request.validated.get('image_type'))

The new image type is in `self.request.validated['document']['image_type']`.

asaunier:
OK thanks. PR updated and ready to review.
Do you think it would make sense to move the newly created function ``_is_creator`` to some more generic place?
https://github.com/c2corg/v6_api/blob/images/c2corg_api/views/image.py#L60-L70

asaunier:
Arf :p 
Travis complains about the maps and areas since I have updated ``requirements.txt`` to take this change into account: https://github.com/c2corg/v6_common/commit/8849577220d77c8e1f7b66aa6681ee45aaf2fefa
but it seems related to a former change:
https://github.com/c2corg/v6_common/commit/d659052b1192f0975866548562eff8c546ae3bda

c2corg_api/views/image.py
```diff
@@ -19,10 +29,42 @@ def collection_get(self):
...
+                 self.request.authenticated_userid,
+                 image_id):
+                raise HTTPForbidden('No permission to change this image')
         return self._put(Image, schema_image)
+
+    def _is_creator(self, user_id, image_id):
+        """Check if the user with the given id has permission to change a
+        personal image (is the author).
+        """
+        initial_version = DBSession.query(DocumentVersion) \
```
tsauerwein: You could write this as:

    DBSession.query(
            exists().where(
                DocumentVersion.document_id == image_id,
                HistoryMetadata.user_id == user_id)).scalar()

By doing this you avoid downloading the data for `DocumentVersion`.

tsauerwein:
> Do you think it would make sense to move the newly created function _is_creator to some more generic place?

Maybe in in `models/document.py`?

tsauerwein:
Sorry, about the issue with the common project. I pushed without considering that this might break your build. I reverted my commit. If you update to https://github.com/c2corg/v6_common/commit/5bf3030ff56af6d096b78cdfa113b0c46bd10b8b, it should work now.

c2corg_api/views/image.py
asaunier: I don't think this can work: we need to get the older version whatever the language and then check that the user_ids match. Your query only checks that the user has created a version (which might not be the first one).

tsauerwein: Good point, you can add the check for `DocumentVersion.version == 1`.

asaunier: OK
I just wanted to avoid this, because in the future we might have to implement version deleting. Then we could have no version 1.

asaunier:
e1eac54 moves the document creator check into the ``Document`` model.
Please note I had to import the ``DocumentVersion`` and ``HistoryMetaData`` models directly in the method itself. If I try to import at the beginning of the file I get the following error (circular import?):
>   File "/home/alex/v6_api/c2corg_api/models/document.py", line 24, in <module>
    from c2corg_api.models.document_history import DocumentVersion, HistoryMetadata
  File "/home/alex/v6_api/c2corg_api/models/document_history.py", line 1, in <module>
    from c2corg_api.models.user import User
  File "/home/alex/v6_api/c2corg_api/models/user.py", line 3, in <module>
    from c2corg_api.models.user_profile import UserProfile
  File "/home/alex/v6_api/c2corg_api/models/user_profile.py", line 2, in <module>
    from c2corg_api.models.document import (
nose.proxy.ImportError: cannot import name 'ArchiveDocument'

I have used an ``exists()`` statement using ``ArchiveDocument.version == 1``. Does it make sense (``DocumentVersion`` has no ``version`` property).

tsauerwein:
That looks great! Thanks for your persistence.

asaunier:
Thanks for your help and your review @tsauerwein !
I have just merged the fixup commits and push the branch back.
I will merge when Travis gives a thumb up.

PR #156, opened 2016-02-23
Add user profile

tsauerwein:
TODO:

* [x] Create user profile when creating user
* [x] Adapt migration script

### DIFF
--- a/c2corg_api/models/__init__.py
+++ b/c2corg_api/models/__init__.py
@@ -30,3 +30,4 @@ from c2corg_api.models import association  # noqa
 from c2corg_api.models import topo_map  # noqa
 from c2corg_api.models import area  # noqa
 from c2corg_api.models import area_association  # noqa
+from c2corg_api.models import user_profile  # noqa
--- a/c2corg_api/models/enums.py
+++ b/c2corg_api/models/enums.py
@@ -94,3 +94,5 @@ map_scale = enum(
     'map_scale', attributes.map_scales)
 area_type = enum(
     'area_type', attributes.area_types)
+user_category = enum(
+    'user_category', attributes.user_categories)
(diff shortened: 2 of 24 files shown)


tsauerwein:
Ready to review

asaunier:
Looks good to me, thanks!

Concerning the migration of the user profiles, I will ask the association if the "pro" category is really relevant. I don't think so since pros could be mountain guides, mountain leaders ("accompagnateurs en montagne"), ski instructors, climbing instructors, etc. => category "pro" is too wide to be useful. In addition we probably need a category for hut wardens. But no need to bother about that now.

Concerning the "profileless" users, just curious what this means: do we really have accounts with no profile?

tsauerwein:
> Concerning the "profileless" users, just curious what this means: do we really have accounts with no profile?

Yes, these four: 110471, 110473, 112744, 112745

tsauerwein:
Thanks for the review!

asaunier:
Another question: I guess that as for other document types, the user profile "title" is edited directly in the document (creating a new archive version etc.), right?

Maybe it's OK, however please note that in v5 the "profile" title (aka "topoguide name") is updated using a different "user account settings" interface where the user may change his "topoguide name" as well as his "forum name", email address, password etc.

tsauerwein:
> Maybe it's OK, however please note that in v5 the "profile" title (aka "topoguide name") is updated using a different "user account settings" interface where the user may change his "topoguide name" as well as his "forum name", email address, password etc.

Ok, I see. I think it makes things too complicated to use the profile title as "guidebook name". We should add a new field to the user model for that.

asaunier:
Can't we ignore the userprofile's title (by simply not showing it in the profile page nor in the editing page) and display the user's "guidebook name" instead?

Issue #158, opened 2016-02-26
Main associated WP when migrating from v5 to v6 should be a summit


asaunier:
The migration script sets the "main waypoint" to the highest associated WP in v5. Actually it should be the highest associated summit (not other WP types).

For instance http://www.camptocamp.org/routes/642876/fr/pointe-adolphe-rey-super-lionel
=> the access is the upper cable car station of Aiguille du Midi (3776m) which is higher than the real summit (Pointe Adolphe Rey 3535m). The main waypoint should be "Pointe Adolphe Rey", not "Téléphérique de l'Aiguille du Midi" as currently in v6.

PR #161, opened 2016-02-29
Make sure main WP are set to v5-summits when migrating

asaunier:
Fixes https://github.com/c2corg/v6_api/issues/158


tsauerwein:
Doing it like that, does every route have a main waypoint?

asaunier:
I think so, because in v5 all routes have at least one summit and I have made sure to include all v5-summit-like WP in the list of "authorized" main WP.

tsauerwein:
:+1:

PR #169, opened 2016-03-02
Also extract [abstract] as summary in migration

### LINKED ISSUE #165: Migration script: abstract tag
> The current script to migrate the topoguide, detects the [abs] tags to put the content into the new abstract field.
> In the V5, there are 2 tags to define an abstract : [abs] and [abstract]. They are decoded in the same html code. "abs" is only a shortcut for "abstract".
> The content between [abstract] tags must be migrated as the [abs] tags.

from https://github.com/c2corg/v6_ui/issues/213

tsauerwein:
Fixes #165


asaunier:
Looks good to me.

Issue #164, opened 2016-03-01
Some routes return a 500 error


asaunier:
For instance:
http://c2corgv6-demo.gis.internal:6545/routes/45412?l=fr
http://c2corgv6-demo.gis.internal:6545/routes/45856?l=fr

In v5:
http://www.camptocamp.org/routes/45412/fr/dent-d-oche-couloir-du-pierrier
http://www.camptocamp.org/routes/45856/fr/dent-d-oche-couloir-du-pierrier-et-petit-tour
(routes associated to each other)

Perhaps sqlalchemy fails retrieving the associated routes when getting the associations of a route?

```
File "/home/alex/v6_api/c2corg_api/views/document.py", line 140, in _set_associations
￼￼linked_waypoints, linked_routes = get_associations(document, lang)
File "/home/alex/v6_api/c2corg_api/models/association.py", line 135, in get_associations
￼￼set_best_locale(child_routes, lang)
File "/home/alex/v6_api/c2corg_api/views/__init__.py", line 130, in set_best_locale
￼￼DBSession.expunge(document)
File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/sqlalchemy/orm/scoping.py", line 150, in do
￼￼return getattr(self.registry(), name)(*args, **kwargs)
File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/sqlalchemy/orm/session.py", line 1494, in expunge
￼￼state_str(state))
sqlalchemy.exc.InvalidRequestError: Instance <Route at 0x7f0c5a7c85f8> is not present in this Session
```

asaunier:
@tsauerwein You mean that the problem came from the routes being associated to each other twice? One with one route as parent and the other with the same route as child?

tsauerwein:
Yes, exactly.

Issue #157, opened 2016-02-25
Limit between which document types associations can be created


tsauerwein:
Currently the association web-service allows to create association between all kind of documents. The following combinations should only be allowed: https://github.com/c2corg/camptocamp.org/blob/master/data/sql/tables/db_associations.sql#L14-L73

To consider: Should two-way links be allowed (e.g. R1 -> R2 and R2 -> R1 between two routes)?

tsauerwein:
Or: Only allow associations between the following types: https://github.com/c2corg/camptocamp.org/blob/master/data/sql/tables/db_associations.sql#L14-L73

asaunier:
Shouldn't we use the other approach? Instead of blocking some kinds of associations we should only allow some. For instance:
* WP
  * WP
  * routes
  * images
  * ...
* routes
 * WP
 * routes
 * images
 * outings
 * ...
* outings
 * routes
 * images
 * users
 * ...

I don't list geoassociations (areas) and maps since they are computed automatically.

asaunier:
> Only allow associations between the following types:

Voila :)

PR #170, opened 2016-03-02
Handle two-way-links between routes

tsauerwein:
Handle the case where there are two associations R1 -> R2 and R2 -> R1 between two routes.

Closes https://github.com/c2corg/v6_api/issues/164


asaunier:
Looks good.
Should we make sure it is not possible to create an association between 2 documents already associated together (even though using the reverse parent>child relationship)?

tsauerwein:
Then we would have to remove these double links during the migration. I added a comment to https://github.com/c2corg/v6_api/issues/157, not totally sure about this yet.

Issue #163, opened 2016-03-01
Add "guidebook name" to user model


tsauerwein:
In v5 the title of an user profile was used as the "guidebook name" for an user. We should add this field to the user model instead.

See discussion here:
https://github.com/c2corg/v6_api/pull/156#issuecomment-188693242

PR #171, opened 2016-03-03
Add `name` field to user model

tsauerwein:
In v5 the "topo_name" of an user was stored on the user, but also as title of the user profile (updated by a trigger). For simplicity the name is now only stored on the user.

Closes https://github.com/c2corg/v6_api/issues/163


gberaudo:
LGTM

asaunier:
OK for me too.
Is the ``topo_name`` already migrated to the user table?

gberaudo:
Yes, as the `name` attribute.

tsauerwein:
See: https://github.com/c2corg/v6_api/pull/171/files#diff-4a2fd9f9c14d29caf41722ede1c20823

Thanks for the review!

PR #172, opened 2016-03-03
Check association types

tsauerwein:
Check if it is valid to create an association between two document types. The valid combinations are listed in the common project: https://github.com/c2corg/v6_common/blob/master/c2corg_common/associations.py

Closes https://github.com/c2corg/v6_api/issues/157


asaunier:
Looks good, thanks!

PR #173, opened 2016-03-04
Check for duplicate associations

tsauerwein:
Avoid that there are two associations D1 -> D2 and D2 -> D1.

Related to https://github.com/c2corg/v6_api/issues/157

### DIFF
--- a/c2corg_api/scripts/migration/documents/associations.py
+++ b/c2corg_api/scripts/migration/documents/associations.py
@@ -105,6 +105,18 @@ class MigrateAssociations(MigrateBase):
             zope.sqlalchemy.mark_changed(self.session_target)
         self.stop()
 
+    def _remove_duplicate_associations(self):
+        """In v5 there are some double associations between documents, for
+        example two associations R1 -> R2 and R2 -> R1. The following two
+        queries remove these duplicates, so that there is only one.
+        """
+        print('Removing duplicate associations')
+        with transaction.manager:
+            self.session_target.execute(SQL_DELETE_DUPLICATE_ASSOCIATONS_LOG)
+            self.session_target.execute(SQL_DELETE_DUPLICATE_ASSOCIATONS)
+            zope.sqlalchemy.mark_changed(self.session_target)
+        print('Done')
+
     def _set_route_main_waypoint(self):
         """Set the field `main_waypoint_id` to the highest associated waypoint.
         """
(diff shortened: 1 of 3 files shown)


gberaudo:
Why is it a problem?

tsauerwein:
It is not really a problem, but if you have the association R1 -> R2 you implicitly also have the association R2 -> R1. This change simplifies things.

asaunier:
Another nice piece of SQL :P 
Not sure to understand the "delete" query. When a duplicate is detected, how do you decide what association to remove? The first one that is found?

tsauerwein:
The query could also be written as follows (if you would use an extension to sort arrays):

    delete from guidebook.association_log l
      using (
        select
          duplicate_link[1] as document_id_1,
          duplicate_link[2] as document_id_2
        from (
          select (sort(ARRAY[a.parent_document_id, a.child_document_id]) duplicate_link
          from guidebook.associations a join guidebook.associations b
            on a.parent_document_id = b.child_document_id and
               a.child_document_id = b.parent_document_id
          group by duplicate_link) t) v
      where l.parent_document_id = v.document_id_1 and
            l.child_document_id = v.document_id_2;

I am removing the association of the two whose parent id is the smaller document id. But this is arbitrary.

asaunier:
> I am removing the association of the two whose parent id is the smaller document id. But this is arbitrary.

For route-route association, dropping whatever duplicate association should be OK, since those are "flat" associations. For summit-summit and access-access, since the association is actually "vertical" (a hierarchy primary summit > secondary summit), that could probably be a problem to drop the wrong association. But:
* do we really have summit-summit or access-access duplicates?
* if yes, how could we automatically figure out what association is the good one? I guess we need a human decision (most likely done upward in v5).

I have tested in a v5 test instance: it's no longer possible to make a duplicate route-route (or summit-summit or access-access) association, even a reverse one (inversion the parent/child).

tsauerwein:
The only duplicate with summits is between the following two:
http://www.camptocamp.org/summits/43404/fr/courchevel-1650
http://www.camptocamp.org/summits/112312/fr/cascades-de-glace-tarentaise

asaunier:
The main summit is obviously http://www.camptocamp.org/summits/112312/fr/cascades-de-glace-tarentaise

tsauerwein:
For reference, the query I used to get the duplicates on the v5 database is:

     select a.main_id, a.linked_id, a.type
     from app_documents_associations a join app_documents_associations b
     on a.main_id = b.linked_id and a.linked_id = b.main_id
     order by a.type;

asaunier:
OK I will try to run it on the real prod DB when I can access the server.

asaunier:
I have run the query in the prod DB and get the following results:
>  main_id | linked_id | type 
---------+-----------+------
  112382 |    106920 | cc
  151860 |    108393 | cc
  138536 |    107005 | cc
  121570 |    121567 | cc
  121567 |    121570 | cc
  119931 |    119928 | cc
  106920 |    112382 | cc
  107005 |    138536 | cc
  108393 |    151860 | cc
  119928 |    119931 | cc
   14777 |    136248 | ii
  136248 |     14777 | ii
  189334 |    189336 | ii
  189336 |    189334 | ii
  189469 |    189471 | ii
  189471 |    189469 | ii
  289513 |     47974 | rr
   45856 |     45412 | rr
   45412 |     45856 | rr
  176334 |     50828 | rr
   50828 |    176334 | rr
   47974 |    289513 | rr
   43404 |    112312 | ss
  112312 |     43404 | ss
(24 rows)

=> 12 duplicated associations?!

tsauerwein:
Yes, right. And still only one for summits.

tsauerwein:
Can this one be merged, or are there still open points?

asaunier:
For the summit-summit duplicate association, I think we have to fix it by hand in v5, right?
All the other duplicates will be handled automatically by the migration script with this change?

tsauerwein:
> For the summit-summit duplicate association, I think we have to fix it by hand in v5, right?

Yes, a human has to decide which association should be kept in this case.

> All the other duplicates will be handled automatically by the migration script with this change?

"handled" aka one will be removed.

asaunier:
OK, looks good to me, :+1: for merging the PR.
I will report the summit-summit duplicate to the association.

tsauerwein:
Thanks for the review and input.

asaunier:
The summit-summit duplicate has been dropped in the prod DB.

PR #174, opened 2016-03-08
Only refresh requirements.txt when needed

gberaudo:
Currently, each time we run the Makefile with the serve target,
the egg target is run and the requirements.txt are refreshed.
This is due to the egg-link never being created.

Remove broken egg target.
Depend on pip file instead of a directory.
Only trigger a requirements.txt refresh when needed.

This commit is equivalent to what was done for the UI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ cleanall: clean
 	rm -rf .build
 
 .PHONY: test
-test: .build/venv/bin/nosetests template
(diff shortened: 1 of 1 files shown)


tsauerwein:
That looks good, thanks!

gberaudo:
Rebased on top of master. Thanks for the review will merge when Travis completes.

PR #175, opened 2016-03-08
Always regenerate template files

gberaudo:
The selection of which template to generate is done based
on the -f config/xx parameter to make. However this is not
trackable by Make. As a consequence this commit always
regenerate the templates, ensuring they are always up-to-date.

This is equivalent to what was done on the UI.


tsauerwein:
:+1:

gberaudo:

PR #176, opened 2016-03-08
Retrieve list of forum private messages

gberaudo:
- split non sso code related to discourse in a separate file;
- add a POST method for retrieving list of all or unread  messages.

### DIFF
--- a/c2corg_api/views/forum.py
+++ b/c2corg_api/views/forum.py
@@ -1,0 +1,35 @@
+from c2corg_api.security.discourse_client import (
+    discourse_get_username_by_userid, get_discourse_client)
+
+from cornice.resource import resource
+
+from c2corg_api.views import cors_policy, restricted_view
+
+import colander
+import logging
+log = logging.getLogger(__name__)
+
+
+class PrivateMessageRequestMapping(colander.MappingSchema):
+    kind = colander.SchemaNode(
+            colander.String(),
+            name='kind',
+            default='all',
+            validator=colander.OneOf(['all', 'unread']))
+
+
+@resource(path='/forum/private-messages', cors_policy=cors_policy)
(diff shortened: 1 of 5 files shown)


asaunier:
Thanks @gberaudo 
Actually we don't need the detailed list of messages, only the number of unread messages (will be shown in the UI near the user name or avatar, as a "new unread messages" indicator). We also need a way to redirect to the MP inbox when the user clicks on some button.

gberaudo:
@asaunier, updated the PR to count and return the number of unread messages.
The link to the forum is returned together so that it is straightforward to integrate in the UI.

asaunier:
OK, thanks for the change.

c2corg_api/views/forum.py
```diff
@@ -0,0 +1,30 @@
...
+
+        client = get_discourse_client(settings)
+        d_username = discourse_get_username_by_userid(client, userid)
+        messages = client.private_messages_unread(d_username)
+
+        count = len(messages['topic_list']['topics'])
+        link = '%s/users/%s/messages' % (
+            get_discourse_public_url(settings), d_username)
+
+        return dict(
```
tsauerwein: Minor: elsewhere we are using `{link: link, count: count}`

asaunier:
Should we advertise somewhere this new ``/forum/private-messages/unread-count`` service or do you think having a look at the ``view_config`` in the ``forum`` view is enough?

tsauerwein:
LGTM

c2corg_api/views/forum.py
gberaudo: OK, updated.

gberaudo:
Thanks for the review @asaunier and @tsauerwein.

Issue #144, opened 2016-02-04
Have a default geometry for every document


tsauerwein:
From https://github.com/c2corg/v6_api/issues/128

> I wonder if we would not need a "point" default geometry for all documents, including the routes and outings. It would be useful for instance to:
> 
>  * make map filtering (advanced search)
>  * show outings on the map: even if we have line geometries, it is probably more usable/performant to show markers rather than a bunch of lines? especially if all features have no own geometry. That probaby can be challenged since in some cases (routes in a WP page for instance) it's interesting to show the actual lines.
>  * make region filtering (not only a map extent filtering) => I guess it's more performant to get a list of points rather than lines when intersectiong with a polygons or even bboxes. Do you think we could replace the current v5 geo-associations filtering by point-geom-based intersections? Would it be too performances-hurting?

> Having a default geometry for all documents would also help to show other documents in the neighbourhood. For instance in an outing (or route or WP) page to get links towards other outings in the area. Or when trying to associate documents together, close documents could be suggested (as it's done in v5 for WP by the way). Etc.
> Doing so we could replace associated-regions based links (not that relevent since the area could be too big). Regions association would then only be used for direct linking between regions and other documents.

tsauerwein:
Notes/ideas from discussions with @asaunier: 

- On the filter map we would only show the default points (also for routes and outings that might have linestrings). When selecting a route, the full linestring will be shown.
- On the detail page for a waypoint, the associated routes are shown as lines in the map.

tsauerwein:
To summarize, the motivation is

 * to have a location for routes and outings that do not have a geometry, so that they are taken into account when doing a spatial search.
 * to have a point geometry for routes and outings which can be used to display them on the map in list/filter views (the full track will be used for the detail view).

We would have to do the following changes:

- [ ] Field `geom` of model `Geometry` is always a point, tracks for routes and outings are stored in `geom_detail`.
- [ ] Adapt the migration script
  - [ ] Set the default geom for outings and routes that already have a track (e.g. using [ST_Line_Interpolate_Point](http://postgis.net/docs/manual-1.5/ST_Line_Interpolate_Point.html)).
  - [ ] If a route does not have a track, take the center of the convex hull of all associated waypoints.
  - [ ] For outings without track, take the default geometry of the associated route(s).
- [ ] Adapt the web-services
 - [ ] When creating a route, and a track is given, use the track to set the default geometry. If not use the point of the main waypoint.
[...]

PR #179, opened 2016-03-10
Use a default geometry

tsauerwein:
See https://github.com/c2corg/v6_api/issues/144#issuecomment-192289223

> To summarize, the motivation is
> 
> * to have a location for routes and outings that do not have a geometry, so that they are taken into account when doing a spatial search.
> * to have a point geometry for routes and outings which can be used to display them on the map in list/filter views (the full track will be used for the detail view).

To do:

- [x] Field `geom` of model `Geometry` is always a point, tracks for routes and outings are stored in `geom_detail`.
- [x] Adapt the migration script
  - [x] Set the default geom for outings and routes that already have a track (e.g. using [ST_Line_Interpolate_Point](http://postgis.net/docs/manual-1.5/ST_Line_Interpolate_Point.html)).
  - [x] If a route does not have a track, take the center of the convex hull of all associated waypoints.
  - [x] For outings without track, take the default geometry of the associated route(s).
  - [x] For outings and routes, set the new default geometry on all geometry versions.
- [x] Adapt the web-services
 - [x] When creating a route, and a track is given, use the track to set the default geometry. If not use the point of the main waypoint.
 - [x] When creating an outing, take the default geometry of the associated route.
- [x] The default geometry can be edited manually by changing `geometry.geom`.

Closes https://github.com/c2corg/v6_api/issues/144


tsauerwein:
Ready to review.

asaunier:
A minor comment else it looks excellent to me!

asaunier:
By the way doing something similar for the images (later...) would be useful too.
And a bit challenging. For instance we could use the gps track of an associated outing to interpolate the coordinates of images using their datetime...

asaunier:
> Makes sense. I moved it to models.utils.

OK, for me you can feel free to merge.

tsauerwein:
Thanks for the review!

PR #177, opened 2016-03-11
Create association to main waypoint when creating/updating route

tsauerwein:
The `main_waypoint_id` field can be set when creating or updating a route. But currently no explicit association is created between the route and this main waypoint.

### DIFF
--- a/c2corg_api/views/area.py
+++ b/c2corg_api/views/area.py
@@ -47,13 +47,13 @@ class AreaRest(DocumentRest):
         return self._put(Area, schema_area, after_update=update_associations)
 
 
-def insert_associations(area):
+def insert_associations(area, user_id):
     """Create links between this new area and documents.
     """
     update_area(area, reset=False)
 
 
-def update_associations(area, update_types):
+def update_associations(area, update_types, user_id):
     """Update the links between this area and documents when the geometry
     has changed.
     """
--- a/c2corg_api/models/association.py
+++ b/c2corg_api/models/association.py
@@ -46,6 +46,14 @@ class Association(Base):
         Base.__table_args__
     )
 
+    def get_log(self, user_id, is_creation=True):
+        return AssociationLog(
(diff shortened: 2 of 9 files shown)


tsauerwein:
Pull-request attached.

tsauerwein:
This PR is still open to review

c2corg_api/views/area.py
```diff
@@ -47,13 +47,13 @@ def put(self):
         return self._put(Area, schema_area, after_update=update_associations)
 
 
-def insert_associations(area):
+def insert_associations(area, user_id):
```
asaunier: Why do we need a user_id here?

tsauerwein: `insert_associations` is called down here:
https://github.com/c2corg/v6_api/pull/177/files/8a2af2cbd823d03305c69d4e3c9611027caabf58..e447928a757b40dde6a4c763044d7cb109b4993c#diff-b906af94797c5db6c394971ac60bf7a4R218

Some methods that use the `after_add` hook require the user id.

asaunier: OK, so you mean that ``user_id`` is passed anyway when calling the ``after_add`` hook and that the hooked methods should have it in their signature, even though not actually using it?

asaunier:
Well, I would say that the main waypoint should first be associated to a route but indeed it's worth checking it is actually associated when setting a new main waypoint.

c2corg_api/views/area.py
tsauerwein: Exactly.

asaunier:
Looks good to me!

tsauerwein:
> Well, I would say that the main waypoint should first be associated to a route but indeed it's worth checking it is actually associated when setting a new main waypoint.

When updating a route, yes. But when creating a new route, it has to be done at the same time.

tsauerwein:
Thanks for the review.

Issue #72, opened 2015-11-10
Documenting the API


asaunier:
The main README already gives some tips on how to use the API:
https://github.com/c2corg/v6_api/blob/master/README.md#requests-examples

but it might be relevent to create a dedicated doc and advertise more comprehensively the possible actions.

For instance to document the pagination parameters.

asaunier:
See this PR that documents how to install the API app (still some work to do):
https://github.com/c2corg/v6_api/pull/261

asaunier:
The API main README has many information. There are also READMEs for several sections giving explanations. And the services parameters are documented as function comments in the code.

Closing the issue, please reopen if more info (tutorial?) are eventually needed.

PR #149, opened 2016-03-11
Two different views to get a document for editing or detail page

tsauerwein:
The view to get a single document now returns associated areas, maps and other documents, because this information is needed in the detail page. It might make sense to have a separate view which is used for the edit form which does only contain the data needed to edit the document.

### DIFF
--- a/c2corg_api/tests/views/test_waypoint.py
+++ b/c2corg_api/tests/views/test_waypoint.py
@@ -105,6 +105,16 @@ class TestWaypointRest(BaseDocumentTestRest):
             self.outing.document_id,
             recent_outings['outings'][0].get('document_id'))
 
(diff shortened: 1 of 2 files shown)


tsauerwein:
Pull-request attached.

When you now request `/waypoints/38575?a=0`, you will not get associations, maps and areas.

asaunier:
Looks good!
Should we document this somewhere?

tsauerwein:
I opened a ticket on the ui project: https://github.com/c2corg/v6_ui/issues/233

>  Should we document this somewhere?

Right now, we have no place where we are documenting the input for our web-services (see also https://github.com/c2corg/v6_api/issues/72). I will see what's the best way to do this.

PR #188, opened 2016-03-16
Fix generation of area associations

tsauerwein:
The area geometry is now in `geom_detail`.

Follow up on https://github.com/c2corg/v6_api/pull/179.


gberaudo:
LGTM

tsauerwein:
Thanks for the review.

Issue #185, opened 2016-03-14
Manage merged documents


asaunier:
For instance:
http://www.camptocamp.org/routes/640047/fr/depuis-les-baudes-boucle-par-le-plan-des-eaux-et-le-col-de-bounavalette
appears normally in v6 (with no associated WP):
http://c2corgv6-demo.gis.internal/routes/640047/fr

How does the API returns merged documents?

Perhaps migrating to v6 would be a good opportunity to get rid of the old merged documents during the migration (but we will still need to support documents merged in v6).

tsauerwein:
Merged documents are migrated as such but not yet handled differently.

tsauerwein:
> Perhaps migrating to v6 would be a good opportunity to get rid of the old merged documents during the migration (but we will still need to support documents merged in v6).

We could ignore merged documents during the migration. But wouldn't you somehow lose the history of a document? I mean we could also simply migrate only the last version, but still we get all versions.

> How does the API returns merged documents?

How exactly should merged documents be handled?

- Not indexed in ElasticSearch.
- Not editable.
- What should be returned? Only `{merged: 12345}` or all the data?

asaunier:
> But wouldn't you somehow lose the history of a document?

Merged documents are already neither searchable nor listed. If you don't have the link you cannot find such documents. Merged documents are generally poor quality documents or not very useful. It's no use having them editable.
It's not a big deal if they are removed. For me it is unnecessary complexity to have those merged documents remaining in the DB. In addition they cause a search engine issue: instead of keeping them (even with a link toward the remaining document), it would probably be better to make a 301 redirection.

> How exactly should merged documents be handled?

Of course not indexed and not editable.
Returning only the id of the target document would make sense so that the UI makes a 301 redirection.

PR #189, opened 2016-03-17
Handle merged documents

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/185

Still to do:

* [x] Do not index in ES
* [x] Ignore when returned as ES search result
* [x] Do not migrate merged documents

### DIFF
--- a/c2corg_api/models/area_association.py
+++ b/c2corg_api/models/area_association.py
@@ -61,6 +61,10 @@ def update_area(area, reset=False):
                 AreaAssociation.area_id == area.document_id)
         )
 
+    if area.redirects_to:
+        # ignore forwarded areas
+        return
+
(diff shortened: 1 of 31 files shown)


asaunier:
The changes about the areas sure make sense. Please note though that since it is not possible (at least in v5) to create areas, we never have to merge areas.

tsauerwein:
> Please note though that since it is not possible (at least in v5) to create areas, we never have to merge areas.

I did it for all document types to be consistent.

asaunier:
Yep, you did well. That was just to precise what actually happens with the areas.

tsauerwein:
Ready to review

asaunier:
Looks good to me, thanks!

Issue #73, opened 2015-11-10
Do not return protected or redirected documents


tsauerwein:
Check the properties `protected` and `redirects_to` (which also should be a FK)

asaunier:
Actually they should be returned by the API but clealy identified as protected or merged/redirected so that the UI adapts the presentation. For merged/redirected we probably need to make a 301 redirection? in v5, no redirection is done, the page is simply marked as "merged with doc ..." and all associations are skipped/removed.

tsauerwein:
How should protected documents be handled?

- GET: return the document with all data?
- PUT: block unless moderator?
- show in list views?
- show in associations?
- index in ES?

asaunier:
Protected documents are only documents that cannot be edited unless deprotected first by a moderator. Only moderators may protect/deprotect a document.

> GET: return the document with all data?

yes

> PUT: block unless moderator?

In v5, no one can edit unless a moderator deprotect the document. We could say that moderators are allowed to edit protected documents without deprotecting them first.

It would probably be a good idea to show an error when opening the editing form of a protected document.

> show in list views?

yes

> show in associations?

yes

> index in ES?

yes

PR #191, opened 2016-03-18
Handle protected documents

tsauerwein:
* Include `protected` field in output
* Only moderators can edit protected documents (changing the protected status will come later with the moderator tools)

PR based on https://github.com/c2corg/v6_api/pull/189, only the last 2 commits are new.
Closes https://github.com/c2corg/v6_api/issues/73

### DIFF
--- a/c2corg_api/models/area_association.py
+++ b/c2corg_api/models/area_association.py
@@ -141,7 +141,7 @@ def get_areas(document, lang):
             AreaAssociation,
             Area.document_id == AreaAssociation.area_id). \
         options(load_only(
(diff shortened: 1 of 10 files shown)


asaunier:
Just one question: can any authenticated user set a document to "protected"?
I think only moderators should be allowed to do so.

tsauerwein:
Right now the protected status can not be changed yet. This will be part of the moderator tools, see https://trello.com/c/kFFBhgAA

asaunier:
OK for merging then.

Issue #155, opened 2016-02-23
Elastic Search indexing must be fine-tuned


asaunier:
- [ ] Don't index documents marked as "merged"
- [ ] Should index the document id as well (in v5 for associating a document it is possible to provide the doc id)
- [ ] ???

tsauerwein:
> Don't index documents marked as "merged"

See also: https://github.com/c2corg/v6_api/issues/73

PR #193, opened 2016-03-18
Include remaining document types in simple search

tsauerwein:
PR based on https://github.com/c2corg/v6_api/pull/191

Still to do:

* [x] Search for document id
* [x] Index user profile
* [x] Search for user profile (only when authenticated)
* [x] Add option to only search for specific document types. E.g. `?q=...&t=w,r` only searches for waypoints and routes.

Closes https://github.com/c2corg/v6_api/issues/155

### DIFF
--- a/c2corg_api/search/sync.py
+++ b/c2corg_api/search/sync.py
@@ -3,6 +3,8 @@ import logging
 
 from c2corg_api.models import DBSession
 from c2corg_api.models.route import Route
+from c2corg_api.models.user import User
+from c2corg_api.models.user_profile import UserProfile
 from c2corg_api.search import elasticsearch_config
 from c2corg_api.search.mapping import SearchDocument
 from c2corg_api.search.utils import strip_bbcodes, get_title
@@ -62,13 +64,27 @@ def sync_search_index(document):
             # https://github.com/c2corg/v6_api/issues/89
             DBSession.refresh(document)
 
+        is_user_profile = isinstance(document, UserProfile)
+        if is_user_profile:
+            # set user login + full-name as document title so that it can
+            # be searched
+            user_login, user_name = DBSession. \
+                query(User.name, User.username). \
+                filter(User.id == document.document_id). \
+                first()
+            user_title = '{0} {1}'.format(user_name or '', user_login or '')
+
         has_title_prefix = isinstance(document, Route)
         for locale in document.locales:
             lang = locale.lang
 
-            # set the title prefix (name of the main waypoint) for routes
-            title_prefix = locale.title_prefix if has_title_prefix else None
-            title = get_title(locale.title, title_prefix)
+            if is_user_profile:
+                title = user_title
+            else:
+                # set the title prefix (name of the main waypoint) for routes
+                title_prefix = locale.title_prefix if has_title_prefix \
+                    else None
+                title = get_title(locale.title, title_prefix)
 
             doc['title_' + lang] = title
             doc['summary_' + lang] = strip_bbcodes(locale.summary)
--- a/c2corg_api/views/search.py
+++ b/c2corg_api/views/search.py
@@ -53,7 +53,7 @@ class SearchRest(object):
 
             `limit=...` (optional)
             How many results should be returned per document type
-            (default: 10). The maximum ist 50.
+            (default: 10). The maximum is 50.
 
             `t=...` (optional)
             Which document types should be included in the search. If not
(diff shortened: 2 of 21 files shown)


tsauerwein:
Ready to review

asaunier:
> Add option to only search for specific document types. E.g. ?q=...&t=w,r only searches for waypoints and routes.

Could we use this to restrict the association tool to a subset of document types?

tsauerwein:
> Could we use this to restrict the association tool to a subset of document types?

Exactly.

c2corg_api/search/sync.py
```diff
@@ -62,13 +64,27 @@ def remove_doc():
...
 
+        is_user_profile = isinstance(document, UserProfile)
+        if is_user_profile:
+            # set user login + full-name as document title so that it can
+            # be searched
+            user_login, user_name = DBSession. \
+                query(User.name, User.username). \
+                filter(User.id == document.document_id). \
+                first()
+            user_title = '{0} {1}'.format(user_name or '', user_login or '')
```
asaunier: Makes sense. @gberaudo do you know what are the relevant usernames you have discussed with the client. It seems that some fields must be kept secret (login?)? Perhaps we should only index public usernames such as topo_name and forum_name (I don't remember the names).

tsauerwein: > It seems that some fields must be kept secret (login?)?

What's the problem with the login name being public? That's the case for basically any community website.

c2corg_api/views/search.py
```diff
@@ -38,34 +38,83 @@ def __init__(self, request):
...
+            `q=...`
+            The search word.
+
+            `lang=...` (optional)
+            When set only the given locale will be included (if available).
+            Otherwise all locales will be returned.
+
+            `limit=...` (optional)
+            How many results should be returned per document type
+            (default: 10). The maximum ist 50.
```
asaunier: I think you meant "is" :P

c2corg_api/views/search.py
```diff
@@ -38,34 +38,83 @@ def __init__(self, request):
...
+                schema_listing_topo_map, None, limit, lang)
+
+        if self._include_type(IMAGE_TYPE, types_to_include):
+            results['images'] = search.search_for_type(
                 search_term, IMAGE_TYPE, Image, DocumentLocale,
-                schema_listing_image, None, limit, lang),
-            'users': search.search_for_type(
+                schema_listing_image, None, limit, lang)
+
+        if self._include_type(IMAGE_TYPE, types_to_include) and \
```
asaunier: IMAGE_TYPE => USERPROFILE_TYPE ?

c2corg_api/search/sync.py
asaunier: c2corg is not just any community website! :P 
I know that that kind of issue was reported for the forum migration. Because some people use their real name as login and that it's not possible to change a login (unlike a forum pseudo or the guidebook name).
@gberaudo probably knows more about this, since he had a meeting with the client yesterday.

asaunier:
Except maybe the login stuff (?!), looks good to me, thanks!

c2corg_api/search/sync.py
gberaudo: Which pseudo fields are secret or not is a bit complicated and seems to have changed over time.
Is the pseudo visible in v6? If no, then why should it be indexed in Elasticsearch?

For Discourse we may use the more or less secret pseudo and allow user to change it afterwards.
See https://github.com/c2corg/v6_forum/issues/8.

c2corg_api/views/search.py
tsauerwein: Ehm, yes. :)

c2corg_api/views/search.py
tsauerwein: Good catch!

c2corg_api/search/sync.py
asaunier: > Is the pseudo visible in v6?

You mean the forum pseudo? Isn't it shown anyway in the posts?
By the way I guess it is not available in the API tables, only in Discourse database? Thus not indexable by ES?

I suggest we keep the user indexing (User.name, User.username) as is and change later if this is a problem for the client.

By the way, could we use ES indexing of the users to associated users to an outing?

tsauerwein: In the api the [user](https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/user.py#L59-L60) has a field `name` and `username`. `name` is filled with the old `topo_name` and `username` with `login_name` (see https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/users.py#L58-L66).

> By the way, could we use ES indexing of the users to associated users to an outing?

We would have to store the usernames on the outing in ES.

tsauerwein:
Thanks for the reviews.

c2corg_api/search/sync.py
asaunier: > We would have to store the usernames on the outing in ES.

I didn't mean that we needed to search outings by the associated users (maybe useful I don't know yet). I meant that we would probably have to use ES to pick the users to associate when editing an outing.

tsauerwein: Yes, the search could be used with `..&t=u`.

Issue #198, opened 2016-03-23
Maps should only be associated to WP and routes


asaunier:
Not to outings for instance.
Then it is no use returning a list of geo-associated maps when requesting the attributes of an outing.

PR #201, opened 2016-03-23
Only include maps for waypoints and routes

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/198


asaunier:
Thanks for the really fast fix! Looks good to me.

Issue #200, opened 2016-03-23
Convert PHP serialized object to JSON


ginold:
Outing property "conditions_levels" is a serialized PHP array, it would be great if during the next DB migration it were converted into a JSON object so i can manipulate it with angular. For now it looks like :
`a:6:{
  i:0;
  a:4:{
    s:11:"level_place";
    s:64:"Départ du virage 1380 entre Liddes et Chandonne (torrent), 8h30";
    s:22:"level_snow_height_soft";
    s:1:"0";
    s:23:"level_snow_height_total";
    s:0:"";
    s:13:"level_comment";
    s:149:"Neige dure, 10m de déchaussage en passant par le clou mais sinon nickel. La neige est un peu plus meuble dans les parties ombragées dans la forêt.";
  }
}`

tsauerwein:
A serialized PHP object?! :O

ginold:
*array :P

asaunier:
Perhaps we should use the python package "phpserialize" https://pypi.python.org/pypi/phpserialize in the migration script in order to unserialize those PHP objects and then ``json.dumps``?

tsauerwein:
`phpserialize` looks promising!

PR #202, opened 2016-03-23
Convert `conditions_levels` from serialized PHP to JSON

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/200


asaunier:
Looks good!

tsauerwein:
I tested with a full import, seems to work.

Issue #194, opened 2016-03-22
Handle i18n when sending emails


gberaudo:
See #190, notably https://github.com/c2corg/v6_api/pull/190#issuecomment-199720299.
TODO:
- [x] register email;
- [x] change password email.

> For the translations for the email: I would propose to start simple without a translation framework and have an email template for each language, e.g.
> 
> /email_templates
>     /registration
>         en.mako
>         fr.mako
>         ...
>     /password_forgotten
>         ...

tsauerwein:
How well does gettext work for longer texts like email templates?

asaunier:
I guess those gettext (or whatever) translations be done on the API server. Then it means it is completely separated from the translations on the UI, thus being a separated gettext project on transifex. Anyway we have to make sure those server translations are not provided in the JSON i18n files downloaded by the browsers.

By the way, if we had translations on the API side, should we translate the error/info messages returned by the API directly on the API side and no longer in the UI?

asaunier:
> How well does gettext work for longer texts like email templates?

Why would it be a problem? gettext msgids may contain "variables" that would then be replaced by the username and validation links for instance.

Please note we will probably have to send emails not only for user subscription or password renewal. For instance in v5, notification emails are sent when a user is associated to an outing.
The corresponding email translation in v5 PO files:
https://github.com/c2corg/camptocamp.org/blob/master/apps/frontend/i18n/messages.fr.po#L3015-L3023

tsauerwein:
> Why would it be a problem? gettext msgids may contain "variables" that would then be replaced by the username and validation links for instance.

Sure it's possible. But is it convenient? Because the email template would be a multi-line string. Would you create a separate translation string for each paragraph? Or have one translation string for the whole text? What if you want to use html for the email?

asaunier:
Just have a look at my example above :P 
A msgstr may contain several lines and HTML tags.

Anyway we don't need to send long emails. Only a few sentences/lines.

tsauerwein:
> Just have a look at my example above :P 

Sorry, I didn't click on the link. :)

tsauerwein:
@gberaudo Can be closed? For gettext there is a separate issue: #238

PR #204, opened 2016-03-24
Handle forgot password

gberaudo:


gberaudo:
Related to https://github.com/c2corg/v6_ui/pull/249.

c2corg_api/emails/email_service.py
```diff
@@ -38,6 +38,13 @@ def send_registration_confirmation(self, lang, user, link):
                 subject='Inscription sur Camptocamp.org',
                 body='Pour activer votre compte cliquez sur %s' % link)
 
+    def send_request_change_password(self, lang, user, link):
+        # TODO: handle i18n
+        self._send_email(
+                user.email,
+                subject='Changement de mot de passe sur Camptocamp.org',
```
asaunier: Shouldn't you use english sentences instead?
This way we will know that the translations are still missing.

c2corg_api/tests/views/test_user.py
```diff
@@ -19,6 +19,29 @@ def setUp(self):  # noqa
     def extract_urls(self, data):
         return re.findall('http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+', data)  # noqa
 
+    def extract_nonce(self, key):
+        validation_url = self.extract_urls(self.get_last_email().body)[0]
+        from urllib.parse import urlparse, parse_qs
```
asaunier: Why not moving this import at the beginning of the file?

asaunier:
2 minor comments else looks good to me

c2corg_api/emails/email_service.py
gberaudo: We have an hardcoded prefered language to 'fr' so I think it is best to hardcode sending emails in French. It is a temporary solution until we get https://github.com/c2corg/v6_api/issues/194, hopefully next week.

c2corg_api/tests/views/test_user.py
gberaudo: OK, done.

c2corg_api/emails/email_service.py
asaunier: I disagree. All interface strings are written in english in the code, why using french for emails? Especially if it is a temporary solution: it's not a problem for testers if they receive messages in english.

c2corg_api/emails/email_service.py
```diff
@@ -38,6 +38,13 @@ def send_registration_confirmation(self, lang, user, link):
...
+                body='Pour changer votre mot de passe cliquez sur %s' % link)
```
tsauerwein: Is the email sent as HTML? Because otherwise you might not be able to click on the link.

c2corg_api/views/user.py
```diff
@@ -53,6 +54,8 @@ def validate_json_password(request):
         # later on requires plain string otherwise it raises
         # the "Unicode-objects must be encoded before hashing" error.
         password = request.json['password']
+        if len(password) < 3:
+            raise 'Password too short'
```
tsauerwein: When you `raise ...`, does this also add the error message as with `request.errors.add('body', 'password', 'Required')`?

c2corg_api/views/user.py
```diff
@@ -198,6 +192,36 @@ def post(self):
...
+            raise HTTPInternalServerError('Error persisting user')
+
+        # FIXME use attribute stored in the user object
+        lang = 'fr'
+
+        email_service = get_email_service(request)
+        nonce = user.validation_nonce
+        settings = request.registry.settings
+        link = settings['mail.request_password_change_url_template'] % nonce
+        email_service.send_request_change_password(lang, user, link)
```
tsauerwein: When sending the mail fails, can the use simply request the password change a 2nd time?

c2corg_api/views/user.py
```diff
@@ -198,6 +192,36 @@ def post(self):
...
+class UserRequestChangePasswordRest(object):
+    def __init__(self, request):
+        self.request = request
+
+    @json_view(validators=[partial(validate_required_json_string, "email")])
+    def post(self):
+        request = self.request
+        email = request.validated['email']
+        user = DBSession.query(User).filter(User.email == email).first()
+        user.update_validation_nonce(VALIDATION_EXPIRE_DAYS)
```
tsauerwein: What does this do? Does this set the user back to unconfirmed?

tsauerwein:
Could you document somewhere how the 'password forgotten' procedure works? Because I didn't understand it...

c2corg_api/views/user.py
gberaudo: No, it adds a password invalid error, see the except block below.

c2corg_api/views/user.py
gberaudo: No, this is handled in the dedicated `email_validated` attribute.

c2corg_api/views/user.py
gberaudo: Sure, the user can request a new password change any time, notably if sending the email fails.
In addition, an error is reported to the user if the email was not sent.

c2corg_api/views/user.py
tsauerwein: Why are you doing this check in the `try except` block? It would maybe cleaner to write:
```
    if 'password' not in request.json:
        request.errors.add('body', 'password', 'Required')
        return # note the return!

    password = request.json['password']
    if len(password) < 3:
        request.errors.add('body', 'password', 'Password too short')
        return
    try:
        # We receive a unicode string. The hashing function used
        # later on requires plain string otherwise it raises
        # the "Unicode-objects must be encoded before hashing" error.
        request.validated['password'] = password.encode(ENCODING)
    except:
        request.errors.add('body', 'password', 'Invalid')
```

That way it is clearer what raises an exception.

c2corg_api/emails/email_service.py
gberaudo: Though it is sent as plain text, clicking the link works in modern email clients like gmail (and mutt! ^^).
A message like `click the following link or copy it to your browser` would be great (to do later, together with i18n).

gberaudo:
Sure, I was planing to add some doc. Here it is: https://github.com/c2corg/v6_api/wiki/Email-based-authentication.

c2corg_api/emails/email_service.py
gberaudo: OK changing these temporary strings to English.

c2corg_api/views/user.py
gberaudo: Done

gberaudo:
Thanks for the review. I think it is now ready for merging.

asaunier:
OK for me.
@tsauerwein OK for you?

tsauerwein:
:+1: for my side. Thanks for adding the documentation!

Issue #116, opened 2015-12-22
Store interface language for users


tsauerwein:
A user should be able to set their preferred interface language which is stored in the database and used every time they access the site.

gberaudo:
This could be done this way:
- during registration, read the current language of the UI and send it together with other info;
- store it in user table;
- use it for sending emails...
- each time the user change the language a request is sent to the API to update the state.

tsauerwein:
And:

 * when logging in, the language is sent to the UI so that the interface language can be set accordingly.

asaunier:
Could this "user prefered language" attribute be used to serve documents when the language is not specified?
For instance in v5, there are links such as http://www.camptocamp.org/summits/37355 that will redirect to http://www.camptocamp.org/summits/37355/fr/mont-blanc depending on the available locales, the UI interface language or some priority list of languages.

Of course for anonymous users we would not have such a "user prefered language" attribute and should use a default lang.

tsauerwein:
On the ui server side you usually do not have the authentication token (only when doing xhr requests where the token is explicitly set in the headers).

But the current interface language is stored in a cookie, which is already used for list views, see https://github.com/c2corg/v6_ui/pull/73.

asaunier:
Can we close this issue @gberaudo ?

PR #208, opened 2016-03-29
Preferred language

gberaudo:


gberaudo:
See #116.

c2corg_api/models/user.py
```diff
@@ -66,6 +67,8 @@ class User(Base):
     validation_nonce_expire = Column(DateTime, nullable=True, unique=False)
     _password = Column('password', String(255), nullable=False)
 
+    lang = Column(String(2), nullable=False, default='fr')
```
tsauerwein: This should be a foreign-key:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/document.py#L223

tsauerwein:
Looking good! Only one small comment.

c2corg_api/models/user.py
gberaudo: Thanks, done.

gberaudo:
Thanks for the review.

PR #209, opened 2016-03-29
Skip 'other_users' user category when migrating from v5

asaunier:
Else the profile migration tool fails because of an unknown "5" value introduced recently in v5:
https://github.com/c2corg/camptocamp.org/blob/03e9666faf87f5c667bf7c68b47727f11f58a2d3/apps/frontend/modules/users/config/module.yml#L29

### DIFF
--- a/c2corg_api/scripts/migration/documents/user_profiles.py
+++ b/c2corg_api/scripts/migration/documents/user_profiles.py
@@ -131,7 +131,7 @@ class MigrateUserProfiles(MigrateDocuments):
                 document_in.activities, MigrateRoutes.activities),
             categories=self.convert_types(
                 categories, MigrateUserProfiles.user_categories,
-                skip_values=[0, 2]),
+                skip_values=[0, 2, 5]),
         )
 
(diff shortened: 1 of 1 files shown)


tsauerwein:
You don't want to map this category to a new category?

Otherwise feel free to merge.

asaunier:
I don't see the point of an "other_users" category, it makes more sense to choose no category.

fbunoz:
The "other_users" category must be keeped as a category option.

2016-03-29 15:39 GMT+02:00 Alex Saunier <notifications@github.com>:

asaunier:
Why?

fbunoz:
It avoids that users choose any category when no one corresponds to their
status, instead keeping the field empty.

2016-03-29 17:03 GMT+02:00 Alex Saunier <notifications@github.com>:

asaunier:
We don't need this in v6: this attribute is a multi-checkboxes field, it's easy to uncheck all categories if the user finally realizes no one is relevant.

PR #211, opened 2016-03-30
Get and update user account

gberaudo:
While logged in, allow:
- changing password;
- changing toponame and forum username;
- changing email (with validation).

TODO:
- [x] update migration script to set forum name;
- [x] require name to be not null, add name to registration form;
- [x] rollback on discourse sync failure;
- [x] do not sync to discourse when only password change;
- [x] mock sync_sso part of discourse client and implement more tests.


gberaudo:
To discuss: do we want to rollback the change when there is a failure in propagating the change to discourse? Or do we accept inconsistency for security reason (a user should never be prevented to change her password or email)?

c2corg_api/views/user.py
```diff
@@ -67,15 +67,18 @@ def validate_json_password(request):
         request.errors.add('body', 'password', 'Invalid')
 
 
+def available_user_attribute(attrname, value):
+    attr = getattr(User, attrname)
+    return DBSession.query(User).filter(attr == value).count() == 0
```
tsauerwein: You could also use `exists()`:
```
    attr_exists = DBSession.query(User.id).filter(attr == value)
    return DBSession.query(attr_exists).scalar()
```

c2corg_api/views/user.py
```diff
@@ -108,8 +111,9 @@ def __init__(self, request):
         partial(validate_unique_attribute, "username")])
     def post(self):
         user = schema_create_user.objectify(self.request.validated)
+        user.forum_username = user.username  # same usernames by default
         user.password = self.request.validated['password']
-        user.update_validation_nonce(VALIDATION_EXPIRE_DAYS)
+        user.update_validation_nonce('regemail', VALIDATION_EXPIRE_DAYS)
```
tsauerwein: Should 'regemail' and 'newpass' maybe be constants?

c2corg_api/views/user.py
```diff
@@ -67,15 +67,18 @@ def validate_json_password(request):
...
 def validate_unique_attribute(attrname, request):
     """Checks if the given attribute is unique.
     """
 
     if attrname in request.json:
         value = request.json[attrname]
-        attr = getattr(User, attrname)
-        count = DBSession.query(User).filter(attr == value).count()
-        if count == 0:
+        if available_user_attribute(attrname, value):
```
tsauerwein: What about renaming `available_user_attribute` to `is_unique_user_attribute`?

```python
if is_unique_user_attribute(attrname, value):
```

c2corg_api/views/user.py
```diff
@@ -162,7 +168,9 @@ class UserValidateNewPasswordRest(object):
     def __init__(self, request):
         self.request = request
 
-    @json_view(validators=[validate_user_from_nonce, validate_json_password])
+    @json_view(validators=[
```
tsauerwein: Shouldn't all these views that required authentication use `@restricted_view`?

c2corg_api/views/user.py
```diff
@@ -288,6 +296,146 @@ def post(self):
...
+            colander.String(),
+            missing=colander.drop,
+            validator=colander.All(
+                colander.Email(),
+                colander.Function(
+                    partial(available_user_attribute, 'email'),
+                    'Already used email'
+                )
+            ))
+    toponame = colander.SchemaNode(
```
tsauerwein: Can we use consistent names for these fields? `name` instead of `toponame` and `forum_name` instead of `forumname`.

c2corg_api/views/user.py
```diff
@@ -288,6 +296,146 @@ def post(self):
...
+        self.request = request
+
+    def get_user(self):
+        userid = self.request.authenticated_userid
+        return DBSession.query(User).get(userid)
+
+    @restricted_view(renderer='json', http_cache=0)
+    def get(self):
+        user = self.get_user()
+        return {
```
tsauerwein: `user_name` is not required/should not be returned?

c2corg_api/views/user.py
```diff
@@ -288,6 +296,146 @@ def post(self):
...
+            }
+
+    @restricted_json_view(renderer='json', schema=updateschema, http_cache=0)
+    def post(self):
+        user = self.get_user()
+        request = self.request
+
+        result = {}
+
+        # Before all, check whether the user knows current password
```
tsauerwein: "know **the** current ..."

c2corg_api/views/user.py
```diff
@@ -288,6 +296,146 @@ def post(self):
...
+    @restricted_view(renderer='json', http_cache=0)
+    def get(self):
+        user = self.get_user()
+        return {
+            'email': user.email,
+            'toponame': user.name,
+            'forumname': user.forum_username
+            }
+
+    @restricted_json_view(renderer='json', schema=updateschema, http_cache=0)
```
tsauerwein: Do you also need to set `http_cache` for `POST`?

c2corg_api/security/discourse_client.py
```diff
@@ -53,10 +53,16 @@ def discourse_get_username_by_userid(client, userid):
 def discourse_sync_sso(user, settings):
     key = str(settings.get('discourse.sso_secret'))  # must not be unicode
     client = get_discourse_client(settings)
+    if user.forum_username is None:
```
tsauerwein: When you register a new account, would you have to provide at least two different user names?

c2corg_api/views/user.py
tsauerwein: `user.name` is not a required field. So, without `user.user_name` you might not have a name to display in the guidebook.

c2corg_api/tests/views/test_user.py
```diff
@@ -285,3 +285,40 @@ def test_restricted_request_unauthenticated(self):
 
     def _add_test_data(self):
         pass
+
+    def test_read_and_update_account_info(self):
```
tsauerwein: I would appreciate some more tests (e.g. failures) because changing the email and password is critical.

c2corg_api/security/discourse_client.py
tsauerwein: `forum_username` is `nullable=False`. So how can this variable ever be `None`?

asaunier:
> To discuss: do we want to rollback the change when there is a failure in propagating the change to discourse? Or do we accept inconsistency for security reason (a user should never be prevented to change her password or email)?

Are the password and email stored in Discourse really used?
If yes, we can't accept that the propagation fails.

tsauerwein:
> To discuss: do we want to rollback the change when there is a failure in propagating the change to discourse? Or do we accept inconsistency for security reason (a user should never be prevented to change her password or email)?

I tend to say that the operation should fail when the changes can not be synced with Discourse. Otherwise it will create unexpected effects (e.g. the email was changed, but Discourse is still sending to the old email).

tsauerwein:
Despite the number of remarks that I made, this looks like good work! Thanks! :+1:

c2corg_api/views/user.py
gberaudo: Using `exists()` would not be significantly faster and is harder to read.
So I would favor sticking with `count()`.

c2corg_api/views/user.py
gberaudo: I think it is a good idea in general but here it would be overkill ; we only use those keys twice in this file.

c2corg_api/views/user.py
tsauerwein: I haven't done any performance tests. But `exists` stops as soon as the first result is found. `count` checks all rows/the whole index.

c2corg_api/views/user.py
tsauerwein: Since when is a constant overkill?! :P

c2corg_api/views/user.py
gberaudo: When calling the function we are actually intersted in knowing if this value is available (unused yet).
I considered a lower level name like `would_violate_unique_constraint_for_user_attribute` though I felt it a bit too long.
What about `is_available_user_attribute` or `is_unused_user_attribute`?

tsauerwein: `is_unused_user_attribute` would be fine for me.

c2corg_api/views/user.py
gberaudo: Resources for validating a nonce received by email are public thus should not use `@restricted_view`.

tsauerwein: Ok, right. But this is the link that you click on in the confirmation email? So it shouldn't be `POST`?

c2corg_api/views/user.py
gberaudo: These are the names displayed in v5 (account parameters).
I tried to keep the same naming in the UI and the submitted form.
In addition, angular sometines does not like `_` so I favored a safe one word for the forum name.

c2corg_api/views/user.py
gberaudo: This resource is for modifying the account parameters.
The usename should be set during registration.

c2corg_api/views/user.py
gberaudo: OK

c2corg_api/security/discourse_client.py
gberaudo: Right, it is a leftover.

c2corg_api/views/user.py
gberaudo: I did a quick test:
```
c2corg_gberaudo=# explain select count(*) from users.user where username = 'guillaume';
                                        QUERY PLAN                                         
-------------------------------------------------------------------------------------------
 Aggregate  (cost=4.44..4.45 rows=1 width=0)
   ->  Index Only Scan using user_username_key on "user"  (cost=0.41..4.43 rows=1 width=0)
         Index Cond: (username = 'guillaume'::text)
(3 rows)


c2corg_gberaudo=# explain select exists (select 1 from users.user where username = 'guillaume');
                                         QUERY PLAN                                          
---------------------------------------------------------------------------------------------
 Result  (cost=4.43..4.44 rows=1 width=0)
   InitPlan 1 (returns $0)
     ->  Index Only Scan using user_username_key on "user"  (cost=0.41..4.43 rows=1 width=0)
           Index Cond: (username = 'guillaume'::text)
(4 rows)
```

c2corg_api/views/user.py
gberaudo: OK

c2corg_api/views/user.py
gberaudo: We use non constant everyhere in the code. Notably see a few lines before, each time we access the `request.validated` map.

c2corg_api/views/user.py
gberaudo: It is handled in the UI and coming from angular.

c2corg_api/views/user.py
gberaudo: OK

c2corg_api/views/user.py
asaunier: I don't find "check if a user exists with a given attr" harder to read than "count how many users have a given attr".

c2corg_api/views/user.py
tsauerwein: We also use constants everywhere in the code. Notably in the very same line. ;)

In this case it would be helpful to see at one place which values can be used as `purpose`.

gberaudo:
Thanks for the review; I created a TODO with the (many) tasks to be addressed before merging the PR.

c2corg_api/tests/views/test_user.py
gberaudo: Added to TODO list.

c2corg_api/views/user.py
tsauerwein: Ok, makes sense.

c2corg_api/views/user.py
tsauerwein: For all other fields (100+) we use the same name in the database model, in the API and in the UI. Let's please do the same for these two. (And there are also fields with an underscore that apparently haven't been a problem so far with Angular.)

c2corg_api/views/user.py
tsauerwein: `GET /users/account` is for modifying account parameters?

But this a more general question. @asaunier, what was decided for `User.user_name`? Can this be shown? The problem is that `name` (the old `topo_name`) is not a required field, thus it can be empty. If we can not display `user_name` we might not have a name to display for a user?!

c2corg_api/views/user.py
tsauerwein: Thanks for testing! So, you were right, it doesn't make a difference. Maybe Postgres takes into account that the column is unique?

c2corg_api/views/user.py
gberaudo: I see no such cases:
```
c2corg=# select count(*) from app_users_private_data where topo_name is null;
 count 
-------
     0
(1 row)                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                       
c2corg=# select count(*) from app_users_private_data where topo_name = '';                                                                                                                                                                                      
 count                                                                                                                                                                                                                                                                 
-------                                                                                                                                                                                                                                                                
     0                                                                                                                                                                                                                                                                 
(1 row)                                                                                                                                          ```

Comment (asaunier) [inline c2corg_api/views/user.py]: > @asaunier, what was decided for User.user_name? Can this be shown? 

I don't know. @gberaudo has attended a meeting with the client about the forum. I haven't. I remember they complained that the login appeared in the forum instead of the "pseudo" (forumname?). What has been decided with the client about this?

> The problem is that name (the old topo_name) is not a required field, thus it can be empty.

??
As far as I know the name/topo_name should be required since we need it as a title in the user profiles, in all guidebook pages where users are mentioned (history, outing participants, etc.).
Should I ask the client for a confirmation about how login/forum name/topo name should be used?

Comment (gberaudo) [inline c2corg_api/views/user.py]: They said that as long as the application allows changing these names they are fine to use.

Comment (asaunier) [inline c2corg_api/views/user.py]: > I see no such cases (last dump):

Yes, in v5 topo_name is required.

Comment (tsauerwein) [inline c2corg_api/views/user.py]: Ok, so it's an error that `name` is not required. How does it work during the registration? For example do we expect that the user enters three different user names? Or only require `user_name` and use this name for the other two if no value was given?

Comment (asaunier) [inline c2corg_api/views/user.py]: OK, then to summarize we need a login (not changeable?), ``name`` (= full name = guidebook name = user profile title), ``forum_name``. All 3 are required. The 2 last ones at least should be changeable.

Comment (asaunier) [inline c2corg_api/views/user.py]: > Or only require user_name and use this name for the other two if no value was given?

Sounds reasonable.

File: c2corg_api/views/user.py (hunk 2 above)
Comment (gberaudo) [inline c2corg_api/views/user.py]: I clearly saind constants are the way to go in the general case.
Typically, the expiration time should be a constant and its value easily modifiable in one place.

The values used for `purpose` need to match, in the same way the attribute name need to match the one used during validation. Not sure using constants everywhere, in all cases, would be very pythonic.

File: c2corg_api/views/user.py (hunk 6 above)
Comment (gberaudo) [inline c2corg_api/views/user.py]: Updating the TODO with the change to the registration form and the user model.

**Comment by asaunier** [comment]
According to the remaining TODO liste, I guess this PR is not ready yet? :P

**Comment by gberaudo** [comment]
Indeed @asaunier, this PR is still WIP.

**Comment by gberaudo** [comment]
It is almost ready so you can already have a look if you want.

**Comment by gberaudo** [comment]
It **is** ready. @tsauerwein, I have added mocks for the discourse client.

File: c2corg_api/scripts/v5/create_unique_normalized_usernames.sql [hunk 11]
```diff
@@ -0,0 +1,71 @@
...
+ALTER TABLE punbb_users RENAME COLUMN username TO old_username;
+
+
+-- Add a username column with normalized old_username for Discourse and constraints
+ALTER TABLE punbb_users ADD COLUMN  username varchar(15);
+
+-- Initially populate with unaccented alphanumeric old_username truncated to 15 characters
+update punbb_users set username = substring(regexp_replace(unaccent(old_username), '[^a-zA-Z0-9]', '', 'g'), 0, 16);
+
+-- Create an call a normalizing function which will update the username column for conflicting rows
```
tsauerwein: Create an**d**

c2corg_api/security/discourse_client.py
```diff
@@ -1,71 +1,162 @@
...
+        # FIXME: are we guaranteed usernames can never change? -> no!
+        self.discourse_username_cache = {}
+
+        self.client = DiscourseClient(
+                self.discourse_base_url,
+                api_username='system',  # the built-in Discourse user
+                api_key=self.api_key,
+                timeout=TIMEOUT)
+
+    def get_userid_by_userid(self, userid):
```
tsauerwein: Maybe `get_discourse_userid_for_userid`?

tsauerwein: Or simply: `get_userid` as the function will be called like `discouse_client.get_userid(someid)`.

c2corg_api/security/discourse_client.py
```diff
@@ -1,71 +1,162 @@
...
+        discourse_userid = self.discourse_userid_cache.get(userid)
+        if not discourse_userid:
+            discourse_user = self.client.by_external_id(userid)
+            discourse_userid = discourse_user['id']
+            self.discourse_userid_cache[userid] = discourse_userid
+            self.discourse_username_cache[userid] = discourse_user['username']
+        return discourse_userid
+
+    def get_username_by_userid(self, userid):
```
tsauerwein: only `get_username`?

c2corg_api/tests/__init__.py
```diff
@@ -186,3 +191,20 @@ def tearDown(self):  # noqa
         DBSession.remove()
         self.session.close()
         self.connection.close()
+
+    def app_post_json(self, *args, **kwargs):
```
tsauerwein: Good idea!

tsauerwein:
I added a couple of minor comments, but it looks very good to me! Thanks for extending the tests.

gberaudo:
Thanks for the review, I updated with your comments and rebased on top of master. I will merge with Travis green light. 11 commits and 60 comments, that became a huge PR!

PR #214, opened 2016-03-31
Use custom ES mapping type for each document type

tsauerwein:
Right now there is only one ElasticSearch [mapping type](https://github.com/c2corg/v6_api/blob/master/c2corg_api/search/mapping.py) for all document types. For the advanced search we will have to store more fields in ES so that it makes sense to have a separate mapping type for each document types.

After this PR is merged, the ES index has to be rebuilt:

```bash
curl -XDELETE 'http://localhost:9200/c2corg_tsauerwein/'
.build/venv/bin/initialize_c2corg_api_es development.ini
.build/venv/bin/fill_es_index development.ini
```

### DIFF
--- a/c2corg_api/scripts/es/fill_index.py
+++ b/c2corg_api/scripts/es/fill_index.py
@@ -14,7 +14,6 @@ from datetime import datetime, timedelta
 
 from sqlalchemy import engine_from_config
 
-from c2corg_api.search.mapping import SearchDocument
 from c2corg_api.models import DBSession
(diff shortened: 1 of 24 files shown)


gberaudo:
Is it still possible to search without knowing in advance the type of the document?

tsauerwein:
It still is, though even now we always know the type in advance.

gberaudo:
Thanks, LGTM.

tsauerwein:
Thanks for the review.

PR #219, opened 2016-04-06
Move quality_types to c2corg_common

### LINKED ISSUE #213: Quality values should be retrieved from c2corg_common
instead of behing hard coded in https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/document.py#L28-L34

asaunier:
Fixes #213 
Requires https://github.com/c2corg/v6_common/pull/17

### DIFF
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -5,7 +5,6 @@ from sqlalchemy import (
     Boolean,
     String,
     ForeignKey,
-    Enum,
     func
     )
 from sqlalchemy.ext.declarative import declared_attr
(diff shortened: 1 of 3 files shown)


tsauerwein:
:+1:

c2corg_api/models/document.py
```diff
@@ -62,8 +53,7 @@ def redirects_to(self):
             Integer, ForeignKey(schema + '.documents.document_id'),
             nullable=True)
 
-    quality = Column(
-        Enum(name='quality_type', inherit_schema=True, *quality_types))
+    quality = Column(enums.quality_type)
```
asaunier: Is this change OK @tsauerwein ?
Do we need to have ``inherit_schema=True`` somewhere?

tsauerwein: This was to make sure that the enum is created in the `guidebook` schema. Now this is ensured with:
https://github.com/c2corg/v6_api/pull/219/files/dabe5ff4aef6c7a9b8d09e75cef5acb99a2a63fb#diff-058a5619d0451ccd20f1bd7f3f2b6358L9

asaunier: OK thanks

tsauerwein:
Looks good!

Issue #220, opened 2016-04-07
Quality of migrated documents


asaunier:
- [x] All migrated documents should have a quality value = "correct" (3rd level in https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L610-L616).
- [ ] Nice to have: set quality to "medium" (2rd level) if "[ebauche]" or "[ébauche]" or "(ébauche)" is detected in the document title: http://www.camptocamp.org/routes/list/srnam/ebauche/orderby/rnam/order/asc
- [ ] Nice to have 2: remove this tag from title if detected.

stef74:
@asaunier propose to close ... nice to have will be manual task

asaunier:
OK, closing then

PR #222, opened 2016-04-08
Quality field

asaunier:
Related to https://github.com/c2corg/v6_api/issues/220 and https://github.com/c2corg/v6_common/pull/19

Currently the tests fail because
> sqlalchemy.engine.base.Engine: INFO: UPDATE guidebook.documents SET version=%(version)s, quality=%(quality)s WHERE guidebook.documents.document_id = %(guidebook_documents_document_id)s AND guidebook.documents.version = %(guidebook_documents_version)s
sqlalchemy.engine.base.Engine: INFO: {'quality': None, 'version': 2, 'guidebook_documents_version': 1, 'guidebook_documents_document_id': 1531}
sqlalchemy.engine.base.Engine: INFO: ROLLBACK
txn.139844722906880: DEBUG: abort
c2corg_api: DEBUG: debug_authorization of url http://localhost/waypoints/1531 (view name '' against context IntegrityError('(psycopg2.IntegrityError) null value in column "quality" violates not-null constraint\nDETAIL: Failing row contains (2, null, null, w, 1531, null).\n',)): Allowed (NO_PERMISSION_REQUIRED)

see https://github.com/c2corg/v6_common/pull/19#issuecomment-207486296

### DIFF
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -23,6 +23,7 @@ from c2corg_api.ext import colander_ext
 from c2corg_api.models.utils import copy_attributes, extend_dict
 from pyramid.httpexceptions import HTTPInternalServerError
 from c2corg_common import document_types
+from c2corg_common.attributes import quality_types
 
 UpdateType = enum.Enum(
     'UpdateType', 'FIGURES LANG GEOM')
@@ -53,7 +54,8 @@ class _DocumentMixin(object):
             Integer, ForeignKey(schema + '.documents.document_id'),
             nullable=True)
 
-    quality = Column(enums.quality_type)
+    quality = Column(
+        enums.quality_type, nullable=False, server_default=quality_types[1])
 
     type = Column(String(1))
     __mapper_args__ = {
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -381,7 +381,7 @@ class ArchiveDocumentGeometry(Base, _DocumentGeometryMixin):
 
 
 schema_attributes = [
-    'document_id', 'version', 'locales', 'geometry'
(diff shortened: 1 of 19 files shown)


tsauerwein:
> Do you mean I should make sure the quality attribute is actually set in https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/document.py#L111 ?

When the field `quality` is not sent in an update request, it is like sending `"quality": null` which means that you explicitly want to set the value to null (which conflicts with the not-null constraint). If you don't want to change the value, you have to send the current value, like for all other fields.

I fixed the tests for you.

c2corg_api/models/document.py
```diff
@@ -53,7 +54,8 @@ def redirects_to(self):
             Integer, ForeignKey(schema + '.documents.document_id'),
             nullable=True)
 
-    quality = Column(enums.quality_type)
+    quality = Column(
+        enums.quality_type, nullable=False, server_default=quality_types[1])
```
tsauerwein: The default value should be 'stub'?

tsauerwein:
Feel free to merge when you agree with my changes.

c2corg_api/models/document.py
asaunier: Yes, the client told me that the default value for migrated documents should be "medium" and "stub" for new documents (unless the contributor changes the quality value when saving the document - which is the encouraged behaviour).

asaunier:
> I fixed the tests for you.

Wow, cool, thanks!

Then the quality field must be provided when saving the document, which does make sense since the field is available in the editing form (it is not possible to choose an empty value). And if it is missing for some reason, the API returns an error (because of the ``nullable=False`` constraint).

Issue #207, opened 2016-03-28
Waypoint model: wrong type for "custodianship"


asaunier:
WP attribute is currently defined as a boolean:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/waypoint.py#L142

The specs on google docs say it is an enum attribute (but single selected value) whose values are:
>   oui - fermé hors gardiennage
  oui - ouvert hors gardiennage
  clé à récupérer
  non gardienné

ginold:
I can't find those values in attributes.py, should I create it, use a <select> tag with those explicit <option> values  or is it somewhere else?

asaunier:
Yep, we need to create the values (in english - can be simplified) in https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py 
Then the enum must be added in https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/enums.py and finally used in https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/waypoint.py#L142

PR #224, opened 2016-04-11
custodianship is a multi-value attribute, not a boolean

asaunier:
Fixes https://github.com/c2corg/v6_api/issues/207

### DIFF
--- a/c2corg_api/models/waypoint.py
+++ b/c2corg_api/models/waypoint.py
@@ -139,7 +139,7 @@ class _WaypointMixin(object):
     phone_custodian = Column(String(50))
 
     # gardiennage (gite, camp_site, hut, abri, bivouac, base_camp)
-    custodianship = Column(Boolean)
+    custodianship = Column(enums.custodianship_type)
 
     # matelas hors gardiennage (hut, abri, bivouac, base_camp)
     matress_unstaffed = Column(Boolean)
--- a/c2corg_api/scripts/migration/documents/waypoints/huts.py
+++ b/c2corg_api/scripts/migration/documents/waypoints/huts.py
@@ -2,6 +2,7 @@ from c2corg_api.models.waypoint import WAYPOINT_TYPE
 from c2corg_api.scripts.migration.documents.document import DEFAULT_QUALITY
 from c2corg_api.scripts.migration.documents.waypoints.waypoint import \
     MigrateWaypoints
+from c2corg_common.attributes import custodianship_types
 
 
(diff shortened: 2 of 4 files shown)


c2corg_api/scripts/migration/documents/waypoints/huts.py
```diff
@@ -94,6 +95,16 @@ def get_document_locale(self, document_in, version):
             access_period=document_in.staffed_period
         )
 
+    def set_custodianship(self, document_in):
```
tsauerwein: `get_custodianship`?

c2corg_api/models/waypoint.py
```diff
@@ -139,7 +139,7 @@ class _WaypointMixin(object):
     phone_custodian = Column(String(50))
 
     # gardiennage (gite, camp_site, hut, abri, bivouac, base_camp)
-    custodianship = Column(Boolean)
+    custodianship = Column(enums.custodianship_type)
```
tsauerwein: With 'multi-value' you don't mean you can select multiple values? You would have to use `Column(ArrayOfEnum(enums.custodianship_type))` in that case.

asaunier: I meant it is a field with more value than yes/no. But only one value at a time can be selected. Do you think I should use another word than "multi-value"?

c2corg_api/scripts/migration/documents/waypoints/huts.py
asaunier: The goal of the function is to set the value of the ``custodianship`` attribute, isn't it?

c2corg_api/models/waypoint.py
tsauerwein: Ok. No, you don't have to change the commit message for me. I just wanted to make sure that we are talking about the same thing.

c2corg_api/scripts/migration/documents/waypoints/huts.py
tsauerwein: You are using the return value of this function to set the attribute. But this function does not set an attribute.

asaunier: Right. I have pushed a fixup.

asaunier:
The spaces within the custodianship values seem to cause no problem as far as I can see.
But to be consistent with the other attributes I have added them. By the way I have replaced "stub" by "draft" in the quality types since the client prefers this term.

OK for merging the PR?

tsauerwein:
:+1:

Issue #230, opened 2016-04-12
Indoor/outdoor types maybe not imported correctly during the migration


asaunier:
```
c2corg_alex=# select climbing_outdoor_types, count(*) from guidebook.waypoints where waypoint_type = 'climbing_outdoor' group by climbing_outdoor_types;
 climbing_outdoor_types | count 
------------------------+-------
                        |  1036
 {}                     |  2925
(2 rows)

c2corg_alex=# select climbing_indoor_types, count(*) from guidebook.waypoints where waypoint_type = 'climbing_outdoor' group by climbing_indoor_types;
 climbing_indoor_types | count 
-----------------------+-------
                       |  1036
 {}                    |  2925
(2 rows)
```
=> only empty values and the same count for both attributes?

tsauerwein:
And there are no `climbing_indoor` documents. Something is wrong with this function:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/documents/waypoints/sites.py#L111-L139

PR #233, opened 2016-04-13
Fix migration of indoor/outdoor types for sites

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/230


asaunier:
Not tested the migration but it looks good, thanks!

tsauerwein:
I've run a full migration and it worked.

Issue #89, opened 2015-11-27
Make search index sync more fail-safe


tsauerwein:
With PR #82 the ElasticSearch index is updated when a new document is added to the database or when a document is modified. But when there is a problem during this sync operation (for example ElastisSearch is down, does not respond, ...) the update is lost and the ElasticSearch index will still have the old version of a document.

To make it fail-safe, we should use a persisted queue:

> A queue could be used to do the updates asynchronously. Two strategies are possible:
> * When a document is updated, the api app adds an entry (document_id, local) to the queue. A worker task will consume the queue and update the document in the ElasticSearch index.
> * (Erics idea) The api app adds a message to the queue without specific information. When the worker task receives a message, it retrieves all documents from the database that have changed since the last update and makes the changes to the ElasticSearch index.

PR #234, opened 2016-04-13
Decouple synchronization with ElasticSearch

tsauerwein:
To make the synchronization with ElasticSearch more fail-safe (see https://github.com/c2corg/v6_api/issues/89), a syncer script is now taking care of the synchronization. When a document changes, a message is pushed into a Redis queue. The syncer script listens to this queue and upon a message it gets the documents that have changed since the last update and sends these documents to ElasticSearch.

To run the API, besides doing `make -f ... serve` the syncer script has to be started with `make -f ... run-syncer`.

Closes https://github.com/c2corg/v6_api/issues/89


asaunier:
Wow, looks like a really fine work (even though I'm not sure to understand everything so I cannot really make a precise review :P).
Please merge if no else objects shortly!

gberaudo:
I will lack time to do a proper review this week, please merge if you want.

tsauerwein:
> I will lack time to do a proper review this week, please merge if you want.

I am not in a hurry to merge. If you could do a review next week, it would be much appreciated.

gberaudo:
Thanks @tsauerwein, will do the review next Tuesday.

c2corg_api/scripts/es/syncer.py
```diff
@@ -0,0 +1,90 @@
...
+        self.session = session
+        self.session_factory = session_factory
+
+    def get_consumers(self, consumer_factory, channel):
+        return [consumer_factory(
+            queues=[self.queue], callbacks=[self.process_task])]
+
+    def process_task(self, body, message):
+        log.info('Sync requested')
+        message.ack()
```
gberaudo: Do you want failed messages to be removed from the queue (no retry)?
If yes, then I think you should add a comment saying so.

c2corg_api/scripts/es/sync.py
```diff
@@ -0,0 +1,113 @@
...
+
+def sync_es(session):
+    last_update, date_now = es_sync.get_status(session)
+
+    if not last_update:
+        raise Exception('No last update time, run `fill_index` to do an '
+                        'initial import into ElasticSearch')
+
+    # get all documents that have changed since the last update
+    # TODO also check changes to associations and user names
```
gberaudo: Is this TODO for a subsequent PR and already tracked in an issue?

c2corg_api/search/notify_sync.py
```diff
@@ -0,0 +1,40 @@
...
+log = logging.getLogger(__name__)
+
+
+def notify_es_syncer(queue_config):
+    """Notify the syncer script that a document has changed by pushing a
+    friendly message into the Redis queue.
+    """
+    def push_notification():
+        with producers[queue_config.connection].acquire(
+                block=True, timeout=3) as producer:
```
gberaudo: What happens when the timeout is reached?

config/default
```diff
@@ -19,6 +19,10 @@ export elasticsearch_host = localhost
 export elasticsearch_port = 9200
 export elasticsearch_index = c2corg
 
+export redis_url = redis://localhost:6379/
```
gberaudo: If running on the same server, using the redis socket might be more performant.

gberaudo:
Good job. I added a minor comment, +1 for merging when it is addressed.

As a side note, have you considered a simple client/server solution instead of using kombu + redis?

c2corg_api/search/notify_sync.py
tsauerwein: An Empty exception will be [thrown](https://docs.python.org/3/library/queue.html#queue.Queue.get) and an error message is logged. I wonder if that is enough.

config/default
tsauerwein: Thanks to Kombu, you can simply change the url in that case:

http://docs.celeryproject.org/projects/kombu/en/latest/userguide/connections.html#urls

c2corg_api/scripts/es/sync.py
tsauerwein: I created a new issue for the user names: https://github.com/c2corg/v6_api/issues/244

c2corg_api/search/notify_sync.py
gberaudo: When can we get a timeout?
- no more connections in pool (all 200 connections active; do we need so many of them?);
- dead/unresponsive syncher (will it restart automatically? I think there is a sync on start)
Are there other cases?

tsauerwein: > When can we get a timeout?
> 
>     no more connections in pool (all 200 connections active; do we need so many of them?);

Good point, I think we don't need that many.

>     dead/unresponsive syncher

At this point you know nothing about the syncer. You only talk to Redis.

>  (will it restart automatically? I think there is a sync on start)

Right now it does not restart automatically in case it died somehow. This is a point I have to check with Mark.

> Are there other cases?

When Redis is down.

tsauerwein:
> As a side note, have you considered a simple client/server solution instead of using kombu + redis?

Because we already have a Redis running for Discourse, it seemed simpler for me to push a message to a Redis queue and read from that queue, instead of re-inventing some sort of queue or syncer server.

Thanks for the review!

PR #235, opened 2016-04-13
Make sure user is loaded together with profile (search)

tsauerwein:
Currently a separate request had to be made for each user profile to get the associated user. Now it is really done with one request.


asaunier:
Simply by adding ``documents_query =``? Wow!

tsauerwein:
I didn't do it properly before..  :o

PR #237, opened 2016-04-13
Add i18n to registration and change password emails

gberaudo:
Trello: https://trello.com/c/CH4dhxcO/31-outils-d-inscription-7-5j

### DIFF
--- a/c2corg_api/emails/email_service.py
+++ b/c2corg_api/emails/email_service.py
@@ -1,9 +1,35 @@
 from pyramid_mailer import get_mailer
 from pyramid_mailer.message import Message
+from functools import lru_cache
+from c2corg_common.attributes import default_langs
 
 import logging
+import os
 
 log = logging.getLogger(__name__)
+
+
+class EmailLocalizator(object):
+    def __init__(self):
+        self.default_lang = 'fr'
+
+    @lru_cache(maxsize=None)
+    def _get_file_content(self, lang, key):
+        filepath = os.path.dirname(__file__) + '/i18n/%s/%s' % (lang, key)
+        if not os.path.isfile(filepath):
+            filepath = os.path.dirname(__file__) + '/i18n/%s/%s' % ('fr', key)
+        f = open(filepath, 'r')
+        return f.read()
+
+    def get_translation(self, lang, key):
+        if lang not in default_langs:
+            raise Exception('Bad language' + lang)
+        try:
+            return self._get_file_content(lang, key)
+        except:
+            log.exception('The %s translation for %s could not be read' % (
+                lang, key))
+            return self._get_file_content('fr', key)
 
 
 # See https://docs.python.org/3/library/smtplib.html#smtplib.SMTP.sendmail
@@ -19,6 +45,8 @@ class EmailService:
         self.mail_server = '%s:%s' % (host, port)
         self.mailer = mailer
         self.settings = settings
+        localizator = EmailLocalizator()
+        self._ = lambda lang, key: localizator.get_translation(lang, key)
 
     def _send_email(self, to_address, subject=None, body=None):
         """Send an email. This method may throw."""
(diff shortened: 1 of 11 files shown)


c2corg_api/emails/email_service.py
```diff
@@ -1,11 +1,33 @@
...
+import os
 
 log = logging.getLogger(__name__)
 
 
+class EmailLocalizator(object):
+    def __init__(self):
+        self.default_lang = 'fr'
+
+    @lru_cache(maxsize=None)
```
tsauerwein: :+1:

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/194?

c2corg_api/emails/email_service.py
```diff
@@ -1,11 +1,33 @@
...
+    def _get_file_content(self, lang, key):
+        filepath = os.path.dirname(__file__) + '/i18n/%s/%s' % (lang, key)
```
tsauerwein: Do you maybe want to use [os.path.isfile(path)](https://docs.python.org/2/library/os.path.html#os.path.isfile) to check first if the file exists?

tsauerwein:
@asaunier convinced me to use gettext/Transifex also for the API translations, but your approach would be fine for me. @asaunier, also ok for you?

c2corg_api/emails/email_service.py
gberaudo: Even if I checked for the existence of the file, there may still be an exception thrown during the open or read operations.
So I simply access the content and any error is handled by the calling code.
Does it make sense for you?

gberaudo:
I am perfectly fine using gettext/Transifex also for the API translations.

This solution is both simple and working so let's merge it so that we can move forward and when we have gettext integrated in the API we can easily switch backend.

Is it fine for you @asaunier?

c2corg_api/emails/email_service.py
tsauerwein: > Even if I checked for the existence of the file, there may still be an exception thrown during the open or read operations.

True, but when the file does not exist, there is no translation for the requested language so it's ok to use the default language. But when the file exists and it can not be read, there is a problem (e.g. wrong permissions) that has to be resolved.

gberaudo: OK, done and added a log.

tsauerwein: Thanks!

gberaudo:
Restarted the Travis build, will merge this afternoon if there is no objection.

asaunier:
> This solution is both simple and working so let's merge it so that we can move forward and when we have gettext integrated in the API we can easily switch backend.

So you mean this approach is a temporary workaround until gettext is actually available on the API side?
Why not "simply" directly setting gettext up on the API side?

This files-based approach seems not very handy to make the client create/update the translation (using Transifex for instance) .

OK for merging as is but we should open an issue about API-side localisation.

gberaudo:
Certainly not a workaround but a working solution which can easily be updated if/when we have gettext/transifex available in the API.

Having plain files was both very fast to implement and is also very easy for the client who may simply update the email translations directly from the github website.

As said before it fine for me to switch later to a global gettext/transifex solution; @asaunier could you please create the issue with some basic instructions on how gettext/transifex could be integrated? Maybe some ideas could be borrowed from https://github.com/c2corg/v6_ui/pull/217?

> OK for merging as is but we should open an issue about API-side localisation.

Thanks for the review, merging this PR.

c2corg_api/emails/email_service.py
```diff
@@ -19,6 +45,8 @@ def __init__(self, mailer, settings):
         self.mail_server = '%s:%s' % (host, port)
         self.mailer = mailer
         self.settings = settings
+        localizator = EmailLocalizator()
+        self._ = lambda lang, key: localizator.get_translation(lang, key)
```
tsauerwein: post-merge remark: You could also have used:

     self._ = localizator.get_translation

asaunier:
> Having plain files was both very fast to implement and is also very easy for the client who may simply update the email translations directly from the github website.

Sure. But that relying on github is not a convenient way to manage the translations since the people that will do the translations won't be - most likely - developers with github accounts. In addition they will need a unique and user-friendly interface (most likely Transifex) to do both the UI and API translations.

I had no time to make progress on the Transifex integration yet. See https://github.com/c2corg/v6_ui/pull/217

> could you please create the issue

https://github.com/c2corg/v6_api/issues/238

c2corg_api/emails/email_service.py
gberaudo: Thanks for the tip, thought it was like in javascript!

tsauerwein:
@gberaudo Can be closed? For gettext there is a separate issue: https://github.com/c2corg/v6_api/issues/238

PR #239, opened 2016-04-15
Add instructions about usernames normalization in v5 source database

asaunier:


gberaudo:
You should state that the normalization should only be done once.
Otherwise, LGTM, thanks.

asaunier:
OK I have updated the PR as you suggest it.

PR #241, opened 2016-04-16
Output API errors in unit tests

tsauerwein:
So far API errors were not directly visible in failing unit tests. Now, they are shown without further interaction.

Credits to @gberaudo for the idea!


gberaudo:
Thanks for generalizing the approach, please merge.

Issue #218, opened 2016-04-06
TODO accounts


asaunier:
* [x] Make sure the user cannot ask for a new password until the account is validated.
* [x] Get the guidebook name and forum name when registering.
* [x] CAPTCHA
* [x] Clean non validated users

asaunier:
For the records here are the comments from #195 

@gberaudo 
> When a user is registered but the email is not validated in the delay, the user must be deleted.
See also https://github.com/c2corg/v6_api/pull/190#discussion_r57027022.

@tsauerwein 
> Or rather:
>>Maybe it would make sense to delay the creation of the user-profile until the account is confirmed. Otherwise un-confimed users might show up in the search and listings.

@asaunier
> +1 with @tsauerwein proposal: no use creating the profile until the account is confirmed.

PR #245, opened 2016-04-20
Check account is validated before allowing account actions

gberaudo:
In addition, use enums instead of simple strings for purposes.

### DIFF
--- a/c2corg_api/models/user.py
+++ b/c2corg_api/models/user.py
@@ -19,6 +19,8 @@ import colander
 from sqlalchemy.orm import relationship, backref
 from sqlalchemy.sql.schema import ForeignKey
 
+from enum import Enum
+
 import os
 import binascii
 
@@ -42,6 +44,12 @@ class PasswordUtil():
         if isinstance(plain, str):
             plain = plain.encode('utf-8')
         return bcrypt.hashpw(plain, encrypted) == encrypted
+
+
+class Purpose(Enum):
+    registration = 'regemail'
+    new_password = 'newpass'
+    change_email = 'chgemail'
 
 
 class User(Base):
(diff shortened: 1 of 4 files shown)


gberaudo:
See first item of https://github.com/c2corg/v6_api/issues/218.

c2corg_api/models/user.py
```diff
@@ -86,14 +94,18 @@ def update_validation_nonce(self, purpose, days):
         particular actions like changing password or validating an email. It
         must have a short lifespan to avoid unused nonces to become a security
         risk."""
+        if purpose != Purpose.registration and not self.email_validated:
+            # An account must be validated before any other action is tried.
+            raise Exception('Account not validated yet')
```
tsauerwein: Is this the error message that is also returned by the API?

tsauerwein:
Looks good! Thanks for introducing the enums!

c2corg_api/models/user.py
gberaudo: No we send a 500 error in this case, which I think is fine for such a special and rare case.

tsauerwein: So, it's not an error message that the user will see? E.g. an user registers, does not confirm the account, and then tries to request a new password?

gberaudo: Yes, here is what is sent when a user registers, does not confirm the account, and then tries to request a new password:
```
HTTP/1.1 500 Internal Server Error
Content-Length: 110
Content-Type: text/plain
Date: Thu, 21 Apr 2016 07:19:39 GMT
Server: waitress

Internal Server Error

The server encountered an unexpected internal server error
```

tsauerwein: If you raise a `HTTPBadRequest` you would get a JSON error, so that the UI could display a more telling message for the user.

gberaudo: OK, now sending a 400 message:
```
HTTP/1.1 400 Bad Request
Access-Control-Expose-Headers: Content-Type
Content-Length: 119
Content-Type: application/json; charset=UTF-8
Date: Thu, 21 Apr 2016 08:21:10 GMT
Server: waitress

{"status": "error", "errors": [{"name": "Error", "description": ["Account not validated yet"], "location": "request"}]}
```

tsauerwein: Thanks!

gberaudo:
Thanks for the review.

PR #248, opened 2016-04-21
Allow changing password even with Discourse down

gberaudo:


gberaudo:
As a side effect, make local development fun again!

gberaudo:
As a reminder, the reason why we require Discourse to be up in other places is that we want to be sure the Discourse DB is kept synchronized with changes in the API. Since we use SSO and the passwords are only stored in the API, it is safe fo skip Discourse in this case.

c2corg_api/views/user.py
```diff
@@ -187,11 +187,11 @@ def post(self):
                 r = client.redirect_without_nonce(user)
                 response['redirect_internal'] = r
             except:
-                # Any error with discourse must prevent login and validation
+                # Since only the password is changed, any error with discourse
+                # must not prevent login and validation.
```
tsauerwein: If only the password is changed, could you skip the synchronization with Discourse in the first place?

gberaudo: No, we want the user to be logged in at the end of the procedure.

tsauerwein: Ok, I see. This is changing the password as part of the "requesting a new password" procedure.

tsauerwein:
Feel free to merge.

gberaudo:

PR #249, opened 2016-04-21
get current user id

ginold:
return additional info - user id


tsauerwein:
Feel free to merge when Travis passes.

PR #250, opened 2016-04-21
Rename forumname to match database naming

gberaudo:


c2corg_api/views/user.py
```diff
@@ -370,7 +370,7 @@ def get(self):
         return {
             'email': user.email,
             'toponame': user.name,
```
tsauerwein: Do you want to do the same for `toponame` (-> `name`)?

gberaudo: OK, let's simplify stuff! I will update both this PR and https://github.com/c2corg/v6_ui/pull/262.

gberaudo:

PR #251, opened 2016-04-21
Add user id to the information returned on login

ginold:


c2corg_api/views/user.py
```diff
@@ -372,6 +372,7 @@ def get(self):
             'name': user.name,
             'forum_username': user.forum_username,
             'id': user.id
+
```
gberaudo: The id line and the blank line should be removed.

tsauerwein: Why also the id line?

ginold: it only appears in the preview, there is no blank line.

ginold: or at least not in my file oO

gberaudo: @tsauerwein, user id should be get on log in, not when editing the user account.

tsauerwein: @gberaudo Ok, it was not clear for me when this view is used.

tsauerwein: @ginold The version that you want to merge contains a blank line. Make sure that you have pushed the latest version.

Issue #197, opened 2016-03-23
Outings list should be antichronologically ordered by default


asaunier:
Unless a criteria says the contrary, outings should always be listed the newest first.
As for now the default outings list page shows outings from 2015, 2004, 2002...

tsauerwein:
Like the other documents, outings are ordered by the document id (like in v5). As it is possible to date back an outing, you are seeing outings from 2004.

But it would make sense to order outings by `date_start` or `date_end`.

asaunier:
Do you mean we should always explicitely provide the sorting parameters to the API request?

tsauerwein:
> Do you mean we should always explicitely provide the sorting parameters to the API request?

asaunier:
I mean it doesn't make sense that default outings lists in v5 are not date-sorted.
Perhaps the UI should call the API specifiying

asaunier:
Oops closed the issue by mistake :P 

> But it would make sense to order outings by date_start or date_end.

:+1: 

I have just asked the client how outings should be sorted when date_start != date_end. For instance what date should be used in outings lists. Most likely ``date_end``.

Issue #29, opened 2015-10-05
ORM query object from REST url for filter


gberaudo:
The API will have to handle the transformation of REST urls like
`http://www.camptocamp.org/outings/list/summits/639674-41487-40127/orderby/act/order/asc` to
SQLAlchemy Query objects.
In v5, this is done by parsing the URL and delagating to a collection of helper functions like found at `https://github.com/c2corg/camptocamp.org/blob/master/lib/model/doctrine/generated/BaseDocument.class.php#L214`.
I have not found ready to use libraries for handling this.

asaunier:
In v5, the criteria are handled in the Document model (if I got it right). Should we do the same in v6? Dispatch in the various Waypoint/Route/... models depending on the doc type? Handle them in the views?

PR #252, opened 2016-04-21
Integrate advanced search into list views

tsauerwein:
This PR adds support for search parameters to the list views. For example for an URL like `/waypoints?wt=summit&we=4500&a=14274,14270&pl=fr`, a query for ElasticSearch will be created and the matching documents are returned (in this case all summits higher than 4500 meters in France and Italy with French locales).

Still to do:

* [x] Add 'sort by' to the ES query
* [ ] Add support for spatial queries (in a new PR)

Closes https://github.com/c2corg/v6_api/issues/29
Closes https://github.com/c2corg/v6_api/issues/197


tsauerwein:
This PR would be ready to review. I will handle the spatial search in a new PR.

gberaudo:
Will it still work with Elasticsearch 1.6.0 (Ubuntu)?

tsauerwein:
> Will it still work with Elasticsearch 1.6.0 (Ubuntu)?

Likely not. I am planning to use a version >= 2.2 because the spatial support has improved. But you can easily install a more recent version: https://www.elastic.co/downloads/elasticsearch

c2corg_api/search/mappings/waypoint_mapping.py
```diff
@@ -1,11 +1,67 @@
...
     def to_search_document(document, index):
-        return SearchDocument.to_search_document(document, index)
+        search_document = SearchDocument.to_search_document(document, index)
+
+        if document.redirects_to:
+            return search_document
+
+        SearchDocument.copy_fields(
+            search_document, document, SearchWaypoint.FIELDS)
+        search_document['has_phone'] = not (not (
```
gberaudo: Is the double not necessary?

tsauerwein: Yes. In Python `a or b` gives you `a` if `a` is a *truthy* value. By doing `not(a or b)` I get a boolean value, which I inverse again.

c2corg_api/tests/views/test_outing.py
```diff
@@ -41,6 +43,8 @@ def test_get_collection(self):
         self.assertEqual(author['user_id'], self.global_userids['contributor'])
         self._add_test_data()
 
+    # TODO check how these two requests are used (as part of advanced search?)
+    @unittest.skip
```
gberaudo: Do we still need these two methods?

gberaudo:
Great job @tsauerwein! I added some minor comments, feel free to merge when Travis is happy.
Regarding the version of elasticsearch, +1 to switch to ES 2.2.

asaunier:
> For example for an URL like /waypoints?wt=summit&we=4500&a=14274,14270&pl=fr, a query for ElasticSearch will be created and the matching documents are returned (in this case all summits higher than 4500 meters in France and Italy with French locales).

Nice!
I guess the UI will have to add the parameters to the API request depending on the selected filters.
I guess also that those parameters will be added to the advanced search page URL as in http://www.camptocamp.org/routes/list/areas/14442-14421/malt/%3E3000/act/1

I think the client expects the parameters to be as close as possible to the parameters used in v5. See https://docs.google.com/spreadsheets/d/1LTE1UIjsDTdZauDgR4rKe2UCtXF6LG-ELH53Qwy1nmI/edit#gid=0

tsauerwein:
Ok for merging?

> I think the client expects the parameters to be as close as possible to the parameters used in v5

Alex will check this, if required we can still change the names.

asaunier:
Sure, please merge.

gberaudo:
+1

PR #253, opened 2016-04-22
Background jobs

gberaudo:
Add two jobs:
- purge non activated account;
- purge expired tokens.

This PR uses APScheduler which is surprisingly straightforward to use.
An SQLAlchemy session no tied with pyramid/Zope is used.

Closes https://github.com/c2corg/v6_api/issues/70

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,11 +29,16 @@ To set up ElasticSearch
 Run the application
 -------------------
 
-The API consists of two applications, the actual web-application and a syncer script
-that synchronizes the database with ElasticSearch. The two have to be started
-separately.
+The API consists of three applications, the actual web-application, a syncer script
+that synchronizes the database with ElasticSearch and a background jobs script that
+purges the database of non activated accounts and expired tokens. The three have to
+be started separately.
 
-To start the syncer script:
+To start the background jobs script (run-background-jobs-prod for production):
+
+    make -f config/$USER run-background-jobs
+
+To start the syncer script (run-syncher-prod for production):
 
     make -f config/$USER run-syncer
 
--- a/c2corg_api/jobs/__init__.py
+++ b/c2corg_api/jobs/__init__.py
@@ -1,0 +1,40 @@
+import atexit
+from apscheduler.schedulers.background import BackgroundScheduler
+from apscheduler.events import EVENT_JOB_ERROR
+
+from c2corg_api.jobs.purge_non_activated_accounts import purge_account
+from c2corg_api.jobs.purge_expired_tokens import purge_token
+
+import logging
+log = logging.getLogger(__name__)
+
+
+def exception_listener(event):
+    log.exception('The job crashed')
+
+
+def configure_scheduler_from_config(settings):
+    scheduler = BackgroundScheduler()
+    scheduler.start()
+
+    scheduler.add_job(
+        purge_account,
+        id='purge_account',
+        name='Purge accounts which where not activated',
+        trigger='cron',
+        minute=0,
+        hour=0
+    )
+
+    scheduler.add_job(
+        purge_token,
+        id='purge_token',
+        name='Purge expired tokens',
+        trigger='cron',
+        minute=30,
+        hour=0
+    )
+
+    scheduler.add_listener(exception_listener, EVENT_JOB_ERROR)
+
+    atexit.register(lambda: scheduler.shutdown())
--- a/c2corg_api/jobs/purge_expired_tokens.py
+++ b/c2corg_api/jobs/purge_expired_tokens.py
@@ -1,0 +1,22 @@
+from c2corg_api.models.token import Token
+from datetime import datetime
+from sqlalchemy.orm import sessionmaker
+
+import logging
+log = logging.getLogger(__name__)
+
+
+def purge_token(test_session=None):
+    now = datetime.utcnow()
+    session = sessionmaker()() if not test_session else test_session
+
+    try:
+        count = session.query(Token).filter(
+                Token.expire <= now).delete(synchronize_session=False)
+
+        log.info('Deleting %d expired token', count)
+        if count > 0:
+            session.commit()
+    finally:
+        if not test_session:
+            session.close()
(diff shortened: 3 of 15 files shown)


tsauerwein:
I didn't take a closer look at your code, but one think to keep in mind when using `APScheduler` is that there might be multiple instances of the background job when running as wsgi module. Is that a problem?

c2corg_api/jobs/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+def configure_scheduler_from_config(settings):
+    scheduler = BackgroundScheduler()
+    scheduler.start()
+
+    scheduler.add_job(
+        purge_account,
+        id='purge_account',
+        name='Purge accounts which where not activated',
+        # trigger='interval',
+        # seconds=3
```
tsauerwein: is that commented-out on purpose? in which interval does this job run?

c2corg_api/jobs/purge_expired_tokens.py
```diff
@@ -0,0 +1,22 @@
...
+    now = datetime.utcnow()
+    session = sessionmaker()()
+
+    try:
+        query = session.query(Token).filter(Token.expire < now)
+        count = query.count()
+
+        log.info('Deleting %d expired token', count)
+        if count > 0:
+            query.delete(synchronize_session=False)
```
tsauerwein: `query.delete` returns the number of deleted rows, so you would not have to make a separate query:
http://docs.sqlalchemy.org/en/rel_1_0/orm/query.html#sqlalchemy.orm.query.Query.delete

c2corg_api/jobs/purge_non_activated_accounts.py
```diff
@@ -0,0 +1,35 @@
...
+             User.validation_nonce.like(Purpose.registration.value + '_%'),
+             User.validation_nonce_expire < now)).all()
+        ids = [idwrap[0] for idwrap in ids]
+
+        log.info('Deleting %d non activated users: %s', len(ids), ids)
+        if len(ids) > 0:
+            delete(DocumentVersion, DocumentVersion.document_id, ids)
+            delete(HistoryMetaData, HistoryMetaData.user_id, ids)
+            delete(User, User.id, ids)
+            delete(UserProfile, UserProfile.document_id, ids)
```
tsauerwein: I can't remember, did we say we only create the UserProfile when the account is activated?

tsauerwein:
What about a test? :)

Besides my comments/questions it is looking good!

c2corg_api/jobs/purge_non_activated_accounts.py
gberaudo: The current design of the models require that the User is created after the UserProfile to inherit its document id. I had a look to changing this but it seemed quite involved.
Do you want to have a look to this?

gberaudo:
Multi processing: if the application may be run in several instances (multi process or multiple servers) then I think we should have a separate application, running only once, on a single server.

> What about a test? :)

It is in progress, but maybe not so easy.

c2corg_api/jobs/purge_non_activated_accounts.py
tsauerwein: Right, this would not work.

The problem I see right now is that unconfirmed users profiles are sent to ElasticSearch, see https://github.com/c2corg/v6_api/issues/255.

I think you also need to delete the `DocumentLocale` for the profile.

tsauerwein:
> It is in progress, but maybe not so easy.

Maybe just test the function, that is called in the background job.

c2corg_api/jobs/purge_expired_tokens.py
```diff
@@ -0,0 +1,22 @@
...
+log = logging.getLogger(__name__)
+
+
+def purge_token(test_session=None):
+    now = datetime.utcnow()
+    session = sessionmaker()() if not test_session else test_session
+
+    try:
+        count = session.query(Token).filter(
+                Token.expire <= now).delete(synchronize_session=False)
```
tsauerwein: Have you considered creating an index for `expire` to avoid sequential scans? Maybe also for the other query.

c2corg_api/jobs/purge_non_activated_accounts.py
gberaudo: OK, done

c2corg_api/jobs/purge_expired_tokens.py
gberaudo: There is already an index on expire for Tokens.
I added some for the other query.

gberaudo:
Close: #256 
Now running as a standalone application.

README.md
```diff
@@ -29,11 +29,16 @@ To set up ElasticSearch
 Run the application
 -------------------
 
-The API consists of two applications, the actual web-application and a syncer script
-that synchronizes the database with ElasticSearch. The two have to be started
-separately.
+The API consists of three applications, the actual web-application, a syncer script
+that synchronizes the database with ElasticSearch and a background jobs script that
+purge the database of non activated accounts and expired tokens. The three have to
```
tsauerwein: purge**s**

c2corg_api/scripts/jobs/scheduler.py
```diff
@@ -0,0 +1,41 @@
...
+
+from pyramid.scripts.common import parse_vars
+from pyramid.paster import get_appsettings, setup_logging
+
+from sqlalchemy import engine_from_config
+
+from c2corg_api.models import Base, DBSession
+from c2corg_api.jobs import configure_scheduler_from_config
+
+log = logging.getLogger('c2corg_api_background_jobs')
```
tsauerwein: Have you tested if the logging works? For the syncer script I had to add a configuration:
https://github.com/c2corg/v6_api/blob/master/development.ini.in#L60-L63 (same for production.ini)

gberaudo: Yes I tested and it worked. Probably because I am logging with level > `info`.
I will add the same kind of config as yours anyway.

gberaudo:

PR #261, opened 2016-04-26
Add instructions for installing required software packages

asaunier:
Similar to https://github.com/c2corg/v6_ui/pull/324

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,22 @@
 API Application for camptocamp.org v6
 =====================================
 
+Requirements
+------------
+
+ * Python 3.4
+ * PostgreSQL 9.4 / PostGIS 2.1
+ * virtualenv (1.7 or higher)
+ * ElasticSearch 2.3.2
+ * Redis 2.8
+ * GEOS (3.4 or higher)
+
+On Debian/Ubuntu systems the required components may be installed using
+
+    sudo apt-get install virtualenv python3-dev postgresql-9.4 postgresql-9.4-postgis-2.1 postgresql-contrib-9.4 libpq-dev libgeos-dev redis-server libffi6 libffi-dev
+
+The Debian package for ElasticSearch can be downloaded from https://www.elastic.co/downloads/past-releases/elasticsearch-2-3-2
+
 Checkout
 --------
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,6 +23,7 @@ phpserialize==1.3.0
 kombu==3.0.35
 redis==2.10.5
 apscheduler==3.1.0
+requests==2.9.1
 
 # ColanderAlchemy>=0.3.2
 # needed for ColanderAlchemy: https://github.com/stefanofontanelli/ColanderAlchemy/pull/90
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -2,5 +2,4 @@ flake8==2.5.2
 pep8-naming==0.3.3
 nose==1.3.7
 WebTest==2.0.20
-requests==2.9.1
 ipdb==0.9


README.md
```diff
@@ -1,6 +1,17 @@
 API Application for camptocamp.org v6
 =====================================
 
+Requirements
+------------
+
+ * Python 3.4
+ * PostgreSQL 9.4 / PostGIS 2.1
+ * virtualenv (1.7 or higher)
```
tsauerwein: + ElasticSearch 2.3.2
+ Redis 2.8

asaunier: OK thanks.
I have added those components to the list at 727756d
Not sure what packages should be installed though.

tsauerwein: For redis: `redis-server`.
And for ElasticSearch the DEB can be downloaded from here: https://www.elastic.co/downloads/past-releases/elasticsearch-2-3-2

asaunier: Thanks, updated!

asaunier: I see that on the demo server ES 2.1.0 is still used. Should we upgrade to 2.3.2 on this server?

tsauerwein: That would make sense, yes.

asaunier: OK I will open a ticket for the sysadmins.
I don't know if it is related but when I try to run the indexing on my instance on the dev server, the load skyrockets (I have succeeded to index the main DB though...).

tsauerwein: > I don't know if it is related but when I try to run the indexing on my instance on the dev server, the load skyrockets (I have succeeded to index the main DB though...).

Thanks for noticing (my work station is too powerful...). See https://github.com/c2corg/v6_api/pull/281

asaunier:
I have just added GEOS to the requirements, see 27430fa See also https://github.com/c2corg/v6_ui/pull/446

I just wonder if this was necessary since GEOS is required by PostGIS I think and should be pulled when installing PostGIS, shouldn't it?

tsauerwein:
> I just wonder if this was necessary since GEOS is required by PostGIS I think and should be pulled when installing PostGIS, shouldn't it?

PostGIS does not need the package with the header files (libgeos-dev) which seems to be required by Shapely.

asaunier:
OK, then my recent change makes sense. Thanks for the tip!

requirements.txt
```diff
@@ -23,6 +23,7 @@ phpserialize==1.3.0
 kombu==3.0.35
 redis==2.10.5
 apscheduler==3.1.0
+requests==2.9.1
```
tsauerwein: You can then remove it from [dev-requirements.txt](https://github.com/c2corg/v6_api/blob/947c2284557a31564e121ed0082a348f882eaad1/dev-requirements.txt#L5).

And yes, it is correct. It should be listed under `requirements.txt`.

asaunier:
@tsauerwein Your link http://stackoverflow.com/a/31234304 was the good one to fix the "make install" crash with bcrypt. With the packages now listed in the README ``make -f config/alex install`` succeeds on my laptop.

@gberaudo I had to add ``requests=2.9.1`` (version that I have found in my instance on the dev server) with 3a0e4d9 else ``make -f config/alex serve`` fails because:
>   File "/home/alex/Desktop/v6_api/c2corg_api/security/discourse_client.py", line 1, in <module>
    from pydiscourse.client import DiscourseClient
  File "/home/alex/Desktop/v6_api/.build/venv/lib/python3.4/site-packages/pydiscourse/client.py", line 4, in <module>
    import requests
ImportError: No module named 'requests'

Do you think it's OK? Should ``requests`` have been install in some other way?

The PR is ready for reviewing.

requirements.txt
asaunier: OK, you answer my question below :P I will remove it from dev-requirements.txt.

asaunier: OK I have updated the PR. Looks good?

tsauerwein:
Looks good, thanks!

asaunier:
Thanks for reviewing @tsauerwein !

PR #267, opened 2016-04-29
Support bbox filters in advanced search

tsauerwein:
The point geometry of documents is now indexed in ElasticSearch and bbox searches can be made on list views, e.g. `/waypoints?bbox=699398,5785365,799498,5885465`.

### DIFF
--- a/c2corg_api/search/search_filters.py
+++ b/c2corg_api/search/search_filters.py
@@ -1,4 +1,6 @@
 import math
+
+import pyproj
 import re
 from c2corg_api.models.outing import OUTING_TYPE
 from c2corg_api.search.mapping_types import reserved_query_fields
@@ -6,7 +8,7 @@ from functools import partial
 
 from c2corg_api.search import create_search, search_documents, \
     get_text_query
-from elasticsearch_dsl.query import Range, Term, Terms, Bool
+from elasticsearch_dsl.query import Range, Term, Terms, Bool, GeoBoundingBox
 
 
 def build_query(url_params, meta_params, doc_type):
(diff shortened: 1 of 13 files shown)


asaunier:
👍

c2corg_api/search/search_filters.py
```diff
@@ -184,6 +189,33 @@ def create_date_range_filter(field, query_term):
...
+    xmax, ymax = transform(bbox3857[2], bbox3857[3])
+
+    return GeoBoundingBox(
+        geom={'left': xmin, 'bottom': ymin, 'right': xmax, 'top': ymax},
+        type='indexed'
+    )
+
+
+@lru_cache(maxsize=1)
+def get_transform():
```
gberaudo: Have you considered using a file property?

requirements.txt
```diff
@@ -14,7 +14,7 @@ Shapely==1.5.13
 pyproj==1.9.5.1
 pyramid-jwtauth==0.1.3
 bcrypt==2.0.0
-elasticsearch==2.1.0
+elasticsearch==2.2.0
```
gberaudo: Shouldn't it be `2.3.2`?

gberaudo:
I added some minor comments, otherwise looks good to me.

requirements.txt
tsauerwein: `2.3.0` is the latest version: https://pypi.python.org/pypi/elasticsearch

Thanks for the pointer.

c2corg_api/search/search_filters.py
tsauerwein: That makes it easier, yes.

tsauerwein:
Thanks for the reviews!

Issue #269, opened 2016-05-02
Incorrect locales returned for routes


asaunier:
In http://c2corgv6-demo.gis.internal:6545/routes/736485?l=fr
the "figures" and "geometries" correspond to a gully in the Mont Blanc range ("La Noire - Pointe 3386 : Beau comme un rêve"). On the other hand the "locales" come from another route named "Évenos (Gorges du Destel) - Château du diable".

The problem occurs for all routes (not always the same locales though). It seems there is a route mix when assembling the route data.

asaunier:
May be reproduced on the demo instance (up-to-date).

PR #270, opened 2016-05-04
Fix query to load route with `lang` filter

tsauerwein:
Always double-check the query SQLAlchemy is generating.. 

Closes https://github.com/c2corg/v6_api/issues/269


asaunier:
Looks good to me and actually fixing the problem in my instance, thanks!

tsauerwein:
Thanks for testing!

Issue #244, opened 2016-04-20
Make sure ES is updated when user name changes


tsauerwein:
Changing the user-name does not create an entry in the history table, so the syncer script will not know that something has changed. We could add a column `last_updated` to `User`.

asaunier:
Do we really need to update the indexing of documents created/updated by a user when the user name changes?

tsauerwein:
In the search you should be able to search for the user-name, right? So, if the user-name is changed, the corresponding profile document in ES has to be updated.

asaunier:
Indeed. I thought you meant we had to reindex all documents updated by the user, not only his profile.

Issue #205, opened 2016-03-24
Consider using a sitemap for SEO


tsauerwein:
The advanced search will replace the current listing pages. On advanced search pages, new results will be fetched with a XHR request. But for SEO reasons we would also need to render some kind of listing pages on the server-side so that search engines are able to discover all pages.

If we would use the advanced search for these pages rendered on the server-side, ElasticSearch would always be involved. To reduce the load on ES, we could consider generating sitemap files only for the search engines. These sitemaps would be very easy to cache and would be generated without sending requests to ES.

See also: http://www.sitemaps.org/protocol.html#index

asaunier:
Shouldn't this be an issue in the UI rather than the API?

tsauerwein:
It's relevant for both.

asaunier:
> It's relevant for both.

Do you plan to have a sitemap for the API as well?
Or only on the UI side?

tsauerwein:
Where will the UI get the information to generate the sitemap? From the API. :)

You won't use the existing list views to request the data for the sitemap. We will need some optimized service which returns simply the document ids (probably also the locales) for each document type (paginated).

asaunier:
OK, I thought we could use the usual API list view with ``?a=0`` to skip the associations. Don't you think it could be enough? What optimizations do you think of?
Sure we need the lastmod which is probably not available in the usual list views (but it could? what about having the info in the detail view as well?).

tsauerwein:
> OK, I thought we could use the usual API list view with ?a=0 to skip the associations. Don't you think it could be enough? What optimizations do you think of?

We only need document id, available locales and a `lastmod` date, not all the other fields that are included for normal list views. And considering that this sitemap view is going to return all documents, it should be optimized to avoid putting unnecessary load on the database.

PR #276, opened 2016-05-11
Update user profile in ES when user name is changed

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/244

### DIFF
--- a/c2corg_api/models/user.py
+++ b/c2corg_api/models/user.py
@@ -17,6 +17,7 @@ from c2corg_api.models import Base, users_schema, schema
 
 import colander
 from sqlalchemy.orm import relationship, backref
+from sqlalchemy.sql.functions import func
 from sqlalchemy.sql.schema import ForeignKey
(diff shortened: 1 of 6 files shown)


asaunier:
Looks good thanks!

Wouldn't it be relevant to add this ``last_modified`` to all documents in order to use it for sitemaps? https://github.com/c2corg/v6_api/issues/205
Unless we can simply use the ``written_at`` attribute of the document history to do so?
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/document_history.py#L28

tsauerwein:
Thanks for the review.

> Wouldn't it be relevant to add this last_modified to all documents in order to use it for sitemaps? #205
> Unless we can simply use the written_at attribute of the document history to do so?

`written_at` already contains the last-modified time. Maybe it makes sense to cache this field on the document itself, but we will see once we do the sitemaps.

Issue #255, opened 2016-04-22
Handle unconfirmed users


tsauerwein:
* Do not return user profiles of unconfirmed users (when directly requesting a document, in a list view, in a search,... ).
* Do not send unconfirmed users to ElasticSearch.
* Make sure that the profiles are send to ES the moment they are confirmed.

gberaudo:
Ideally, the user profile should only be created once the user is validated.

PR #277, opened 2016-05-11
Handle unconfirmed users

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/255

### DIFF
--- a/c2corg_api/views/document.py
+++ b/c2corg_api/views/document.py
@@ -1,4 +1,5 @@
 from c2corg_api.models.outing import Outing
+from c2corg_api.models.user import User
 from functools import partial
 
 from c2corg_api.models import DBSession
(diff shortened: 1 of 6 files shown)


c2corg_api/views/document.py
```diff
@@ -682,7 +685,17 @@ def get_all_fields(fields, activities, field_list_type):
...
+            options(contains_eager('user'))
+    return document_query
+
+
+def add_profile_filter(document_query, clazz):
+    if clazz == UserProfile:
+        # make sure only confirmed accounts are returned
+        document_query = document_query. \
+            join(User). \
+            filter(User.email_validated)
```
gberaudo: Does this line means `User.email_validated` is true?

tsauerwein: Yes. Do you find it not expressive enough?

gberaudo: Yes it looks like a test for the existence of any value (is not null). It would be more clear to explicitely check for True.

tsauerwein: Considering that `email_validated` is a boolean field, for my taste it's ok to do the check like this. You have been doing it like that [elsewhere](https://github.com/c2corg/v6_api/blob/master/c2corg_api/security/roles.py#L28). ;)

tsauerwein:
Thanks for the review.

Issue #275, opened 2016-05-11
Missing locales for some outings in the list view


asaunier:
In my API dev instance (and in the main API demo instance) I have noticed that some outings have an empty locales attributes (``locales: []``) in the list view, whereas the locales are provided in detail view. See for instance:
* http://c2corgv6-demo.gis.internal:6545/outings (ditto on the main demo URL) => locales are provided for the first 25 outings but empty for outing 736222 and the following outings (is there a limit somewhere in the API request?)
* http://c2corgv6-demo.gis.internal:6545/outings/736222 => locales are available

This seems a quite recent problem. The main demo was OK yesterday until I made a git pull.

Seems OK with other document types such as waypoints or routes.

PR #278, opened 2016-05-11
Use a deterministic order for subqueryload

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/275

For outings ordering only by the date does not return the results in the same order, because there can be multiple outings for the same day. This is a problem when using `subqueryload`, see also:
http://docs.sqlalchemy.org/en/latest/faq/ormconfiguration.html#why-is-order-by-required-with-limit-especially-with-subqueryload


gberaudo:
LGTM, please merge.

PR #279, opened 2016-05-12
 Check for main waypoint before deleting association

### LINKED ISSUE #273: Check for main waypoint before deleting association
Adapt the association web-service so that an association between waypoint and route can not be deleted if the waypoint is main waypoint of the route.

See https://github.com/c2corg/v6_ui/issues/298#issuecomment-218394059

tsauerwein:
Closes #273

### DIFF
--- a/c2corg_api/views/association.py
+++ b/c2corg_api/views/association.py
@@ -1,5 +1,6 @@
 from c2corg_api.models import DBSession
 from c2corg_api.models.document import Document
+from c2corg_api.models.route import Route
 from c2corg_common.associations import valid_associations
 from cornice.resource import resource
 from pyramid.httpexceptions import HTTPBadRequest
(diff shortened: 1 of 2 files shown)


c2corg_api/views/association.py
```diff
@@ -76,6 +78,9 @@ def collection_delete(self):
             if association is None:
                 raise HTTPBadRequest('association does not exist')
 
+        if is_main_waypoint_association(association):
+            raise HTTPBadRequest('waypoint is still main waypoint of route')
```
asaunier: Perhaps the error could be more explicit, like ``As the main waypoint of the route, this waypoint cannot be deassociated``?

asaunier:
Only a minor comment, else looks good to me, thanks.

c2corg_api/views/association.py
tsauerwein: Yes, that makes it a bit clearer. But remember that the error messages that the API returns are not intended to be shown directly to the user.

asaunier: We never know... :P

PR #280, opened 2016-05-12
Improve performance of simple search

tsauerwein:
* Create an index for column `DocumentLocale.document_id` to speed up the queries to load the documents
* Do the queries for ElasticSearch in a single request instead of doing a request for each document type


asaunier:
Looks good to me

PR #281, opened 2016-05-12
Make fill_index script more memory friendly

tsauerwein:
After the latest changes to the fill_index script, the data was no longer streamed from the database. Instead all documents for a document type were first loaded in memory and then pushed to ElasticSearch.

This PR makes sure that the documents are streamed.


asaunier:
:+1:

Issue #223, opened 2016-04-11
Geo-associations are too wide


asaunier:
The geo-associations automatically generated when creating for instance a waypoint seem a bit too wide!

For instance on the demo instance I have created a summit near Torino in Italy and region "France" is geo-associated!

Ditto for WP waypoints/736951/it in the demo instance.

ginold:
In my example, I created a new outing associated to the route "Raid in Oberland : Dal Goms: Aletschgletscher > Fieschergletscher" (629046).
On the outing view-page, all those areas are returned: 
```
Wallis
Italien
Schweiz
Frankreich
Bern
Tessiner Alpen - Goms
Berner Alpen
```

Whereas in the associated route view-page, only 

`Svizzera
Vallese
Alpi Bernesi`

PR #282, opened 2016-05-12
Use ST_Intersects instead of && operator

tsauerwein:
... which only does the intersection test with bboxes.

Closes https://github.com/c2corg/v6_api/issues/223


asaunier:
:+1:

tsauerwein:
Thanks for the reviews, @asaunier!

Issue #286, opened 2016-05-14
Searching users crash


asaunier:
at least on the demo instance.

When typing a user name in the user-associating tool in the outing editing interface, one gets 500-error results from the API, the logs saying that

> [Sat May 14 13:57:18.016505 2016] [wsgi:error] [pid 25547] [remote 10.27.71.4:20189] mod_wsgi (pid=25547): Exception occurred processing WSGI script '/var/www/vhosts/c2corgv6-api/private/v6_api/apache/app-c2corg_api.wsgi'.
[Sat May 14 13:57:18.016900 2016] [wsgi:error] [pid 25547] [remote 10.27.71.4:20189] Traceback (most recent call last):
[Sat May 14 13:57:18.017269 2016] [wsgi:error] [pid 25547] [remote 10.27.71.4:20189]   File "/var/www/vhosts/c2corgv6-api/private/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/router.py", line 223, in __call__
[Sat May 14 13:57:18.017298 2016] [wsgi:error] [pid 25547] [remote 10.27.71.4:20189]     response = self.invoke_subrequest(request, use_tweens=True)
[Sat May 14 13:57:18.017367 2016] [wsgi:error] [pid 25547] [remote 10.27.71.4:20189]   File "/var/www/vhosts/c2corgv6-api/private/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/router.py", line 198, in invoke_subrequest
[Sat May 14 13:57:18.017389 2016] [wsgi:error] [pid 25547] [remote 10.27.71.4:20189]     response = handle_request(request)
[Sat May 14 13:57:18.017487 2016] [wsgi:error] [pid 25547] [remote 10.27.71.4:20189]   File "/var/www/vhosts/c2corgv6-api/private/v6_api/c2corg_api/__init__.py", line 46, in tween
[Sat May 14 13:57:18.017512 2016] [wsgi:error] [pid 25547] [remote 10.27.71.4:20189]     return handler(request)
[Sat May 14 13:57:18.017590 2016] [wsgi:error] [pid 25547] [remote 10.27.71.4:20189]   File "/var/www/vhosts/c2corgv6-api/private/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/tweens.py", line 48, in excview_tween
[...]

PR #288, opened 2016-05-16
Do not try to run search if no valid search types

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/286


asaunier:
Are you sure this actually fix #286?
It seems the problem occured when authenticated.

tsauerwein:
Ok for merging?

asaunier:
Please do.

PR #289, opened 2016-05-17
Fix login and logout

gberaudo:


gberaudo:
See https://github.com/c2corg/v6_forum/issues/23.

asaunier:
Thanks!

PR #290, opened 2016-05-17
Add a script for checking v6 -> discourse communication

gberaudo:
This script should be run after each upgrade.

In production, testing v6 -> discourse communication should be part of the monitoring.


asaunier:
Thanks!
I guess it cannot be integrated to the unittests because it requires the forum app (not only the API app).

gberaudo:
Yes, this is for checking the demo. Thanks for the review.

PR #292, opened 2016-05-18
Fix moderator permission migration

### LINKED ISSUE #287: Moderator permission is not preserved when migrating from v5 to v6
The user accounts of all moderators in the v5 DB have ``moderator=false``in the migrated v6 DB.

gberaudo:
Close #287.


tsauerwein:
:+1:

gberaudo:
Thanks for the review.

Issue #274, opened 2016-05-11
Create/delete associations when creating/updating a document


tsauerwein:
The way creating/updating a document is done in the UI requires that associations can be created/deleted at the same time. The create/update web-service must be adapted so that associations can be passed together with the document.

ginold:
So the UI web-service should return `document : { blabla : blabla, associations : [...] }` and the API will handle the rest?

tsauerwein:
@ginold Yes, I added some details in this PR: https://github.com/c2corg/v6_api/pull/293

Issue #212, opened 2016-03-30
Do not include outing associations in editing view


tsauerwein:
Extension for https://github.com/c2corg/v6_api/pull/149: 

> We would need the associations but not the outings associations
https://github.com/c2corg/v6_ui/pull/261#issuecomment-203478063

Issue #180, opened 2016-03-11
Include associated images for documents


tsauerwein:
Once https://github.com/c2corg/v6_api/pull/154 is merged, image associations can be included in the `associations` property.

PR #293, opened 2016-05-23
Create/delete associations when creating/updating a document

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/274
Closes https://github.com/c2corg/v6_api/issues/180
Closes https://github.com/c2corg/v6_api/issues/212

When creating or updating a document, the associations can now be passed together with the document:

```json
{
    "message": "Update",
    "document": {
        "document_id": 123,
        
        "associations": {
            "users": [
                {"id": 45},
                {"id": 67}
            ],
            "routes": [
                {"document_id": 89}
            ]
        }
    }
}
```

A few remarks:

- For the edit page, a document should now be requested with `...&e=1` to get only the information needed (before there was a parameter `&a=0` to exclude associations, but that name is no longer appropriate).
- For waypoints, parent and child waypoints are returned in two different collections `waypoint_parents` and `waypoint_children`.
- The association types that can be created/updated for a certain document types are listed [here](https://github.com/c2corg/v6_api/blob/associations-edit/c2corg_api/views/validation.py#L296-L303). E.g. for a route you can update associations to other routes and waypoints.

### DIFF
--- a/c2corg_api/models/association.py
+++ b/c2corg_api/models/association.py
@@ -182,3 +182,18 @@ def add_association(
 
     DBSession.add(association)
     DBSession.add(association.get_log(user_id, is_creation=True))
+
+
+def create_associations(document, associations_for_document, user_id):
+    """ Create associations for a document that were provided when creating
+    a document.
+    """
+    main_id = document.document_id
+    for doc_key, docs in associations_for_document.items():
+        for doc in docs:
+            linked_document_id = doc['document_id']
+            is_parent = doc['is_parent']
+
+            parent_id = linked_document_id if is_parent else main_id
+            child_id = main_id if is_parent else linked_document_id
+            add_association(parent_id, child_id, user_id, check_first=False)
--- a/c2corg_api/models/association.py
+++ b/c2corg_api/models/association.py
@@ -1,7 +1,9 @@
 from c2corg_api.models.route import Route, RouteLocale
 from c2corg_api.models.user import User
-from c2corg_api.models.waypoint import Waypoint, WaypointLocale
+from c2corg_api.models.waypoint import Waypoint, WaypointLocale, WAYPOINT_TYPE
 from c2corg_api.views import set_best_locale
+from c2corg_api.views.validation import updatable_associations, \
+    association_keys, association_keys_for_types
 from colanderalchemy.schema import SQLAlchemySchemaNode
 from sqlalchemy import (
     Boolean,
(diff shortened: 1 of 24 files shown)


gberaudo:
For associating (new) images, there was a discussion about directly sending the metadata of the images to associate and have the API:
- create the Image documents;
- create the associations.
@tsauerwein, do you think it is something that could be added easily now or later?

tsauerwein:
@gberaudo, this could be added once the image processing logic is ready. Besides creating the image document and the association, the actual image file would have to be moved from the temporary to the "final" storage.

c2corg_api/models/association.py
```diff
@@ -90,33 +95,71 @@ class AssociationLog(Base):
...
+    if editing_view:
+        edit_types = updatable_associations.get(document.type, set())
+        types_to_include = types_to_include.intersection(edit_types)
+
+    associations = {}
+    if 'waypoints' in types_to_include:
+        waypoint_parents = get_linked_waypoint_parents(document)
+        waypoint_children = get_linked_waypoint_children(document)
+        associations['waypoints'] = waypoint_parents + waypoint_children
+    elif 'waypoint_parents' in types_to_include:
```
asaunier: Not that a big deal but that's a bit not intuitive to get both "waypoint_parents" and "waypoint_children" when ``types_to_include`` only indicates "waypoint_parents" :)

asaunier:
This looks like another huge and impressive PR, thanks :)

> When creating or updating a document, the associations can now be passed together with the document:

If ``associations`` is not passed in the API request, no change occurs in the document's associations?
If it is passed, then all the document's associations are "reset" to its value, right?

> For the edit page, a document should now be request with ...&e=1 to get only the information needed (before there was a parameter &a=0 to exclude associations, but that name is no longer appropriate).

What about documenting this in some README? Ditto for your other remarks?!

c2corg_api/models/association.py
tsauerwein: They can only be used together, but I changed it to make it clearer.

tsauerwein:
> If associations is not passed in the API request, no change occurs in the document's associations?
> If it is passed, then all the document's associations are "reset" to its value, right?

If `associations` is not given, no associations are changed. If for example for routes only `"associations": {"waypoints": [...]}` is given, only the the associations to waypoints are updated but not the associations to other routes.

I added some documentation.

asaunier:
Looks good to me!

gberaudo:
+1

tsauerwein:
Thanks for the reviews!

Issue #283, opened 2016-05-13
Include routes and outings of child waypoints when getting a waypoint


asaunier:
As for now the list of routes associated to a given WP (in a WP detail page) only shows the routes directly associated to the WP.

In v5, this list also contains the routes associated to child-waypoints.

See for instance http://www.camptocamp.org/summits/133422/fr/la-meije#routes

Same issue for the list of outings:
http://www.camptocamp.org/summits/133422/fr/la-meije#outings

PR #294, opened 2016-05-25
Include routes and outings of child waypoints when getting a waypoint

tsauerwein:
Based on https://github.com/c2corg/v6_api/pull/293
Closes https://github.com/c2corg/v6_api/issues/283

### DIFF
--- a/c2corg_api/views/waypoint.py
+++ b/c2corg_api/views/waypoint.py
@@ -1,10 +1,11 @@
 import functools
 
 from c2corg_api.models import DBSession
-from c2corg_api.models.association import Association
+from c2corg_api.models.association import Association, limit_route_fields
 from c2corg_api.models.document import UpdateType, Document, DocumentLocale
 from c2corg_api.models.outing import Outing, schema_association_outing
-from c2corg_api.models.route import Route, RouteLocale, ROUTE_TYPE
+from c2corg_api.models.route import Route, RouteLocale, ROUTE_TYPE, \
+    schema_association_route
 from c2corg_api.views.outing import set_author
 from c2corg_api.views.route import set_route_title_prefix
 from cornice.resource import resource, view
@@ -28,7 +29,11 @@ from c2corg_common.attributes import waypoint_types
 from functools import lru_cache
 from sqlalchemy.orm import joinedload, load_only
 from sqlalchemy.orm.util import aliased
-from sqlalchemy.sql.expression import and_
+from sqlalchemy.sql.elements import literal_column
+from sqlalchemy.sql.expression import and_, union
+
+# the number of routes that are included for waypoints
+NUM_ROUTES = 30
 
 validate_waypoint_create = make_validator_create(
     fields_waypoint, 'waypoint_type', waypoint_types)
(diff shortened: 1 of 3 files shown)


c2corg_api/views/waypoint.py
```diff
@@ -28,7 +29,11 @@
 from functools import lru_cache
 from sqlalchemy.orm import joinedload, load_only
 from sqlalchemy.orm.util import aliased
-from sqlalchemy.sql.expression import and_
+from sqlalchemy.sql.elements import literal_column
+from sqlalchemy.sql.expression import and_, union
+
+# the number of routes that are included for waypoints
+NUM_ROUTES = 30
```
asaunier: We'll have to argue about this limitation!
I have pointed the problem of v5 summits with dozens (or even hundreds of associated routes like  http://www.camptocamp.org/summits/107317/fr/massif-des-calanques) but the association seems relunctant to do anything...

At least would it be possible to raise a little this limit (for instance 50 or even 100)?

tsauerwein: The idea would be to do it like for the recent outings, show only a small number and give a link to see all (paginated).

Showing 100 routes in the card layout isn't very practical.

asaunier:
> Based on #293

OK so only commits 4f0413c and fb05698 are new.

They looks good to me even though I think the limit on the number of displayed associated routes will not please the client :P

c2corg_api/views/waypoint.py
asaunier: > The idea would be to do it like for the recent outings, show only a small number and give a link to see all (paginated).

That definitely makes sense for outings, because you can show only the very last ones.
But how can we figure out what routes to show in the WP page? That's not that logical I'am afraid. As for now you show the 30 most recently created routes, right?
I can check with the association if it would be OK (but I foresee some crying :P). 

> Showing 100 routes in the card layout isn't very practical.

I know :(

tsauerwein: I don't really see your problem. It is out of question to show 100 routes for a waypoint. So we show some (I think even 30 is too much), and if you want to see all you click on "Show all (594)" and you can browse all routes paginated.

asaunier: > It is out of question to show 100 routes for a waypoint.

?
There are several such examples in v5 :)
We'll have to explain the client why this can't be done in v6.

You don't answer my question about how to choose the route to eventually display in the WP page.
I think this is not trivial.

I agree that having too many routes in a WP most likely means the WP definition is questionable (for instance "massif des Calanques") and that it should be changed/splitted. But I'm not sure that's the association thinks so...

tsauerwein: > There are several such examples in v5 :)
> We'll have to explain the client why this can't be done in v6.

And these pages take 30 seconds to load, they are very unpractical because you are not able to find a route in a list of 100+ entries and they would take a ridiculous amount of space with the card layout. For me it is not an option to show 100+ routes on one page.

> You don't answer my question about how to choose the route to eventually display in the WP page.
> I think this is not trivial.

I don't think it is important which routes you show if you have 30+ routes. If you have more than 30 you click on the link and then you can better browse them.

asaunier: > And these pages take 30 seconds to load, they are very unpractical because you are not able to find a route in a list of 100+ entries and they would take a ridiculous amount of space with the card layout. For me it is not an option to show 100+ routes on one page.

OK thanks for the arguments.

> I don't think it is important which routes you show if you have 30+ routes. If you have more than 30 you click on the link and then you can better browse them.

OK. Then let's keep your PR as is.
I will create an issue in the UI since we need a "show more" button for routes the same way it is done for outings.

asaunier:
OK for me to merge this PR after rebasing it.

tsauerwein:
Thanks for the review!

PR #295, opened 2016-05-25
Use 'w' as code for waypoints in URL

asaunier:
and be consistent with the document type in https://github.com/c2corg/v6_common/blob/master/c2corg_common/document_types.py#L8


tsauerwein:
:+1:

asaunier:
Thanks @tsauerwein !

Issue #297, opened 2016-05-30
Search for routes/outings given an associated document


asaunier:
For instance ``routes?w=37355``

As already available for outings, that can be filtered on "w" (waypoint id) or "r" (route id).

**It should be possible to search:**

- a route given a waypoint id
- an outing given a waypoint id, an user id and a route id.

When searching with a waypoint id, children of the waypoint should also be considered (see https://github.com/c2corg/v6_api/issues/283).

tsauerwein:
> As already available for outings, that can be filtered on "w" (waypoint id) or "r" (route id).

Note that these filters can currently not be combined with other filter criteria. The queries to get the documents are directly run on the database and not on ElasticSearch.

If we add support to search for routes associated to a waypoint, I would propose to do this search in ES. We could then do the same for outings.

asaunier:
> If we add support to search for routes associated to a waypoint, I would propose to do this search in ES. We could then do the same for outings.

Yes, this would be helpful to be able to make some additional filtering on the related outings/routes (for instance to get the routes of a given activity for some waypoint). In addition it would be more intuitive for users that when showing the list of routes of a waypoint, something actually happens when changing the criterias available in the page :)

PR #298, opened 2016-06-01
Store document types with associations

tsauerwein:
To simplify queries on associations, the document type is now stored with the association (similar to how it was in v5).

Related to https://github.com/c2corg/v6_api/issues/297

PR #302, opened 2016-06-02
Search for routes/outings given an associated document

tsauerwein:
Todo

* [x] Search a route given a waypoint id (including parent waypoints)
* [x] Search an outing
  * [x] given a waypoint id (including parent waypoints)
  * [x] an user id
  * [x] a route id
* [x] Make sure the search index is updated when associations are created/updated (only if needed) 

Closes https://github.com/c2corg/v6_api/issues/297


tsauerwein:
Ready to review. The first 3 commits are the ones of https://github.com/c2corg/v6_api/pull/298.

asaunier:
Since it is quite advanced changes, it is not easy to have a full understanding of their impacts. At least I have seen nothing that would cause a problem :)

Issue #303, opened 2016-06-03
Error while saving an outing


asaunier:
(Not sure yet the problem occurs systematically)

I have edited one of my existing outings (http://localhost:6555/outings/edit/734063/fr) and added a new route (http://localhost:6555/routes/45168/fr => Mont de Grange couloir de Pertuis).

The API fails with the following SQLAlchemy error:

> File &#34;/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid_debugtoolbar/panels/performance.py&#34;, line 57, in resource_timer_handler
    result = handler(request)
  File &#34;/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/tweens.py&#34;, line 48, in excview_tween
    request_iface=request_iface.combined
  File &#34;/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/view.py&#34;, line 541, in _call_view
    response = view_callable(context, request)
  File &#34;/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/config/views.py&#34;, line 286, in _authdebug_view
    return view(context, request)
  File &#34;/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/config/views.py&#34;, line 385, in viewresult_to_response
    result = view(context, request)
  File &#34;/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/tweens.py&#34;, line 20, in excview_tween
    response = handler(request)
  File &#34;/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid_tm/__init__.py&#34;, line 101, in tm_tween
    reraise(*exc_info)
  File &#34;/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid_tm/compat.py&#34;, line 15, in reraise
    raise value
  File &#34;/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid_tm/__init__.py&#34;, line 83, in tm_tween
    response = handler(request)
  File &#34;/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/router.py&#34;, line 145, in handle_request
    view_name
  File &#34;/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/view.py&#34;, line 541, in _call_view
    response = view_callable(context, request)
[...]

PR #304, opened 2016-06-06
Add contraint for document archives

tsauerwein:
To make sure that there is only one entry for each version for a document id.

Related to https://github.com/c2corg/v6_api/issues/303


asaunier:
:+1:

Issue #301, opened 2016-06-01
Associations: routes vs all_routes & waypoints vs waypoint_children/parents


asaunier:
I am struggling with the associations in the UI https://github.com/c2corg/v6_ui/pull/328

From my perspective the newly added distinctions:
* routes / all_routes
* waypoints / waypoint_children / waypoint_parents
add a bunch of complexity and confusion!

Because in the code (directives/partials/mako) we have to make complex detections to figure out which one to use. And it's confusing to have sometimes "routes", sometimes "all_routes" (in addition the name "all_routes" is really not self explicit since it's actually a subset of associated routes).

I know having dozens (or hundreds) of associated routes is a problem and may cause slow requests but:
* couldn't we say it's the responsibility of the client not to have waypoints with hundreds of routes?
* or if we prefer to apply a limit anyway, couldn't we always have the same associations object for route, whatever the type of the current document ? For instance:
```
associations: {
  routes: {
    total: 123,
    documents: []
  }
}
```

Concerning waypoints, since it's only possible to associate a waypoint to another as a "child waypoint", couldn't we merge ``waypoints`` and ``waypoint_children`` and keep only ``waypoints``? This way when associating a waypoint to a route or a waypoint, the newly associated waypoint will be added to the ``waypoints`` list in both cases.
``waypoint_parents`` could be kept as is (and is not used by the association creation tools anyway).

tsauerwein:
> couldn't we say it's the responsibility of the client not to have waypoints with hundreds of routes?

Yes, let's simply move a route from the Verdon to the Normandy because there are only a few?!

> or if we prefer to apply a limit anyway, couldn't we always have the same associations object for route, whatever the type of the current document? For instance:
> 
> associations: {
>   routes: {
>     total: 123,
>     documents: []
>   }
> }

There is a difference between getting a document for editing and viewing. If you get a document for editing all associations are returned in the same format. For viewing, there are two additional types (`recent_outings` and `all_routes` (I am open to change that name)) that contain aggregated and transitive associations. And only these two use a different schema to account for the fact that not all can be returned at once.


> 
> Concerning waypoints, since it's only possible to associate a waypoint to another as a "child waypoint", 

Are you talking about associating a waypoint from the view or from the edit page?

> couldn't we merge waypoints and waypoint_children and keep only waypoints? 

You mean `waypoint_parents` and `waypoint_children`? Note that this is likely going to change because of https://github.com/c2corg/v6_api/issues/132.

[...]

PR #305, opened 2016-06-06
Change included associations for waypoints

tsauerwein:
* Rename waypoints_parents to waypoints
* Show routes directly associated to a waypoint first

closes https://github.com/c2corg/v6_api/issues/301

### DIFF
--- a/c2corg_api/models/association.py
+++ b/c2corg_api/models/association.py
@@ -105,16 +105,15 @@ def get_associations(document, lang, editing_view):
         types_to_include = types_to_include.intersection(edit_types)
 
     associations = {}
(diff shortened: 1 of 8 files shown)


asaunier:
Is the ``priority`` attribute used on the UI side? Or only on the API side to sort the directly associated routes first?

tsauerwein:
> Is the priority attribute used on the UI side? Or only on the API side to sort the directly associated routes first?

It is only used in the SQL query to sort. It is not even returned.

asaunier:
OK, looks good to me.

tsauerwein:
Thanks for the review.

PR #308, opened 2016-06-08
Set fixed setuptools version in requirements

arnaud-morvan:


tsauerwein:
For reference, this change is to fix a problem with ApScheduler:
https://github.com/agronholm/apscheduler/issues/145

tsauerwein:
Thanks!

PR #309, opened 2016-06-08
Synchronize discourse username

arnaud-morvan:


tsauerwein:
This is to cover the case when only `forum_username` was changed? Do you maybe want to add a simple test for that? Similar to this one: https://github.com/c2corg/v6_api/blob/master/c2corg_api/tests/views/test_user.py#L499

gberaudo:
@arnaud-morvan, your fix for forum username looks good to me.

@tsauerwein already pointed the place to add a test. The test should validate the behaviour when Discourse is up and when it is down.

In addition, change of `name` must also be handled:
see https://github.com/c2corg/v6_api/blob/master/c2corg_api/security/discourse_client.py#L57-L63 for what is synchronized to Discourse.

arnaud-morvan:
This should be good now. I've made some factorisation to avoid code duplication in tests.

gberaudo:
Looks good. Please merge when Travis is happy.

c2corg_api/tests/views/test_user.py
```diff
@@ -493,48 +519,21 @@ def test_read_and_update_account_info_discourse_up(self):
...
-        self.assertEqual(1, int(self.discourse_client.sso_sync.called_count))
-
-    def test_read_and_update_account_info_discourse_down(self):
-        self.set_discourse_down()
-        currentpassword = self.global_passwords['contributor']
-
-        # Read account info
-        url = '/users/account'
-        self.get_json_with_contributor(url, status=200)
+        #self.assertEqual(1, int(self.discourse_client.sso_sync.called_count))
```
tsauerwein: Do you want to remove this line?

tsauerwein:
@arnaud-morvan You can run `make -f config/... check` locally.

PR #323, opened 2016-06-15
Fix extraction of abstract during migration

tsauerwein:
Related to https://github.com/c2corg/v6_api/issues/165


asaunier:
Looks good to me!

Issue #311, opened 2016-06-09
Missing search fields


asaunier:
I will add here the search fields that proved missing while implementing the UI advanced search (see https://github.com/c2corg/v6_ui/pull/373):
* [x] waypoint prominence "prom" https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/waypoint.py#L37
* [x] waypoint's ``climbing_outdoor_types`` https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/waypoint.py#L52
* [x] waypoint's ``climbing_indoor_types`` https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/waypoint.py#L55

Fields that are probably not needed (to be confirmed):
* [x] ``has_phone`` https://github.com/c2corg/v6_api/blob/master/c2corg_api/search/mappings/waypoint_mapping.py#L21-L22

ginold:
prominence is available, but only for type summit.

tsauerwein:
What fields are you listing here? The fields that are indexed in ElasticSearch?

asaunier:
Yes, for instance the ones in https://github.com/c2corg/v6_api/blob/master/c2corg_api/search/mappings/waypoint_mapping.py

tsauerwein:
`prominence`, `climbing_outdoor_types` and `climbing_indoor_types` are added in #324.

tsauerwein:
I am closing this issue with https://github.com/c2corg/v6_api/pull/324 being merged. If there are other fields missing, a new issue can be opened.

Issue #313, opened 2016-06-14
Make it more convenient to search for a rating range


tsauerwein:
See https://github.com/c2corg/v6_ui/issues/368

- when searching for routes with a rating between PD and ED+, the filter would have to be set to `PD-,PD,PD+,...,ED-,ED,ED+`. It should be possible to search with `PD,ED+`. This should also be done for all other *sortable* criteria.

**Waypoint**
   - [x] quality
   - [x] access_time
   - [x] climbing_rating_max/climbing_rating_min
   - [x] paragliding_rating
   - [x] exposition_rating
   - [x] equipment_ratings

**Route**
   - [x] quality
   - [x] durations
   - [x] ski_rating
   - [x] ski_exposition
   - [x] labande_ski_rating
   - [x] labande_global_rating
   - [x] global_rating
   - [x] engagement_rating
   - [x] risk_rating
   - [x] equipment_rating
   - [x] ice_rating
   - [x] mixed_rating
   - [x] exposition_rock_rating
   - [x] rock_free_rating
   - [x] rock_required_rating
   - [x] aid_rating
   - [x] via_ferrata_rating
   - [x] hiking_rating
   - [x] hiking_mtb_exposition
   - [x] snowshoe_rating
   - [x] mtb_up_rating
   - [x] mtb_down_rating

**Outing**
   - [x] quality
   - [x] frequentation_types
   - [ ] ~~awesomeness~~
   - [x] condition_rating
   - [x] snow_quantity
   - [x] snow_quality
   - [x] glacier_rating

- `min_rating` and `max_rating` should be combined into `rating`.

**Waypoint**
   - [x] climbing_rating_max/climbing_rating_min

**Route**
   - [x] elevation_min/elevation_max?

**Outing**
   - [ ] ?

tsauerwein:
@asaunier Could you please double-check if this  are all the attributes that should be searchable with a range? And also the "combinded" attributes (like min/max rating)?

asaunier:
@tsauerwein I have double-checked, that looks good.

For outings, it seems that the "awesomeness" attribute will be removed => no need to take it into account.

tsauerwein:
Is the field `climbing_rating_median` required?

asaunier:
In my opinion no. But it's in the list of searchable fields and people at the association think it is useful. So if possible, yes.

PR #324, opened 2016-06-15
Add missing search fields

tsauerwein:
See https://github.com/c2corg/v6_api/issues/311

### DIFF
--- a/c2corg_api/search/mappings/outing_mapping.py
+++ b/c2corg_api/search/mappings/outing_mapping.py
@@ -40,8 +40,6 @@ class SearchOuting(SearchDocument):
         'swlu', range=True)
     elevation_down_snow = QInteger(
         'swld', range=True)
-    awesomeness = QEnum(
-        'ofun', model_field=Outing.awesomeness)
     condition_rating = QEnum(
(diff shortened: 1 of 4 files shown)


asaunier:
LGTM

Concerning the ``has_phone`` field https://github.com/c2corg/v6_api/blob/master/c2corg_api/search/mappings/waypoint_mapping.py#L21-L22 => where have you seen it was needed? It seems weird to me to look for WP (even huts) that have a phone...

tsauerwein:
> Concerning the has_phone field ... => where have you seen it was needed?

I am waiting on a confirmation that it can be removed.

besson-genostar:
I agree with @asaunier this field feels nonsense to me for a search.

asaunier:
> I am waiting on a confirmation that it can be removed.

I think you can remove it without further confirmation :)

tsauerwein:
From https://github.com/c2corg/v6_api/issues/313#issuecomment-227711310:

> For outings, it seems that the "awesomeness" attribute will be removed => no need to take it into account.

So can I remove that field as well?

asaunier:
> So can I remove that field as well?

Yes

PR #326, opened 2016-06-21
use pydiscourse to look for a document-related thread

ginold:
returns a thread with the topic id that will be used to display comments in the UI.

### DIFF
--- a/config/default
+++ b/config/default
@@ -60,7 +60,7 @@ export mail_debug = 0
 
 # FIXME
 export discourse_sso_secret = d836444a9e4084d5b224a60c208dce14
-export discourse_api_key = 4647c0d98e8beb793da099ff103b9793d8d4f94fff7cdd52d58391c6fa025845
+export discourse_api_key = c1706802dbf1152c91492a9dbd5dc6e1480a1de57c25089401b3f2416d8f6a1a
 
 export skip_captcha_validation = False
 export recaptcha_secret_key = 6LdWkR4TAAAAALcZLh54HFlAWFHiMQhZR5jR4p3F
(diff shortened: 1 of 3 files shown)


asaunier:
There is an "unused python var" error. You can test if in your instance using ``make lint``.

tsauerwein:
How is this service going to be used? After all it is calling an URL like http://c2corgv6-discourse.demo-camptocamp.com/t/jeudis-des-refuges-decouvrir-les-fleurs-de-haute-montagne-a-adele-planchard/146535.json ?

ginold:
This is the way it's used for now. I created an end-point '/get-topic/{document_id}-{document_lang}' which will get the JSON through pydiscourse client.

tsauerwein:
I don't quite understand how this is supposed to work. At what point are the topics created in Discourse?

config/default
```diff
@@ -60,7 +60,7 @@ export mail_debug = 0
 
 # FIXME
 export discourse_sso_secret = d836444a9e4084d5b224a60c208dce14
-export discourse_api_key = 4647c0d98e8beb793da099ff103b9793d8d4f94fff7cdd52d58391c6fa025845
+export discourse_api_key = c1706802dbf1152c91492a9dbd5dc6e1480a1de57c25089401b3f2416d8f6a1a
```
asaunier: Are you sure we should change this key? I remember it is used in the demo instance.
At least perhaps we should save the current key to ``config/main``?

ginold: I dont' remember the reason for changing this, but I see the discourse api key is still the first one... Maybe I will have to revert.

asaunier: @gberaudo has reverted to the first key last week. See https://github.com/c2corg/v6_ui/issues/452#issuecomment-228051981
I don't know why the key had changed in the meantime.

arnaud-morvan:
I've just proposed #340 that should superseed. Do not need to get topic redirection anymore as we keep topic_id in guidebook database.

arnaud-morvan:
Superseeded by #340. Closing

PR #327, opened 2016-06-23
Support enum range filters

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/313

Now it is possible to do searches like `routes?trat=2.1,5.2` (for all routes with a ski rating between 2.1 and 5.2) or `routes?ele=1400,1600` (for all routes with that elevation range).

### DIFF
--- a/c2corg_api/search/mapping.py
+++ b/c2corg_api/search/mapping.py
@@ -1,9 +1,11 @@
 import json
 
 from c2corg_api.models.document import Document
-from c2corg_api.search.mapping_types import Enum, QEnum, QEnumArray, QLong
+from c2corg_api.search.mapping_types import Enum, QEnumArray, QLong, \
+    QEnumRange
 from c2corg_api.search.utils import strip_bbcodes
 from c2corg_common.attributes import default_langs
+from c2corg_common.sortable_search_attributes import sortable_quality_types
 from elasticsearch_dsl import DocType, String, MetaField, Long, GeoPoint
 
 
@@ -33,7 +35,8 @@ class SearchDocument(DocType):
 
     id = Long()
     doc_type = Enum()
-    quality = QEnum('qa', model_field=Document.quality)
+    quality = QEnumRange(
+        'qa', model_field=Document.quality, enum_mapper=sortable_quality_types)
     available_locales = QEnumArray('l', enum=default_langs)
     geom = GeoPoint()
 
@@ -121,6 +124,10 @@ class SearchDocument(DocType):
                 search_document['description_' + locale.lang] = \
                     strip_bbcodes(locale.description)
             search_document['available_locales'] = available_locales
+
+            if document.quality:
+                search_document['quality'] = \
+                    sortable_quality_types[document.quality]
 
             if document.geometry:
                 search_document['geom'] = SearchDocument.get_geometry(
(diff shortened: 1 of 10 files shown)


asaunier:
What if the API receives ``routes?trat=2.1:5.2``? Will it returns routes who rate 2.1 or 5.2?

tsauerwein:
> What if the API receives routes?trat=2.1:5.2? Will it return routes who rate 2.1 or 5.2?

Good point. It should be "less/greater than or equal". I will change it.

asaunier:
Sorry, I meant ``routes?trat=2.1,5.2`` with a comma.

tsauerwein:
> Sorry, I meant routes?trat=2.1,5.2 with a comma.

Sorry, my example in the description was wrong. It is a comma instead of a colon.

asaunier:
The PR relies on changes in c2corg_common, especially https://github.com/c2corg/v6_common/commit/ae4cb5197617302bcba9e6e2bc62b61a02ccd51a

OK for those changes. I just wonder if the enum values could have not been converted to integer values simply (?) by using the index values in the existing attribute values lists? For instance:
https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L345-L367 instead of https://github.com/c2corg/v6_common/blob/master/c2corg_common/sortable_search_attributes.py#L149-L171

I admit it is probably less confusing and more error-proof but this means the attribute values are somehow duplicated and that changes in those values must be done at 2 places instead of one.

tsauerwein:
> I just wonder if the enum values could have not been converted to integer values simply (?) by using the index values in the existing attribute values lists?

I considered that option, but decided to list the values explicitly. That way it is less error prone to add a new value.

> .. that changes in those values must be done at 2 places instead of one.

To avoid mistakes with that, I added a test that checks that both places are updated.

asaunier:
> Sorry, my example in the description was wrong. It is a comma instead of a colon.

OK I have updated your example then.

> To avoid mistakes with that, I added a test that checks that both places are updated.

:+1: 

Concerning the min/max filtering, what are the fields that support it? ``elevation_min/max`` as well as all other "min/max" fields such as the rating fields?
Is it still possible to filter on ``elevation_max`` for instance? (just curious)

tsauerwein:
> Concerning the min/max filtering, what are the fields that support it? elevation_min/max as well as all other "min/max" fields such as the rating fields?

Only `elevation_min`/`elevation_max` for routes and `climbing_rating_min`/`climbing_rating_max` for waypoints. This were the only fields I found where it would make sense.

> Is it still possible to filter on elevation_max for instance? (just curious)

Yes!

asaunier:
> Only elevation_min/elevation_max for routes and climbing_rating_min/climbing_rating_max for waypoints. This were the only fields I found where it would make sense.

True!

The PR looks great to me. For me it's OK to merge it (and need it for the UI side :P).

tsauerwein:
> Good point. It should be "less/greater than or equal". I will change it.

I double-checked, this works as expected.

Thanks for your review!

asaunier:
I guess one must recreate the ES index after upgrading the API code?

tsauerwein:
Yes, indeed.

asaunier:
> Only elevation_min/elevation_max for routes and climbing_rating_min/climbing_rating_max for waypoints. This were the only fields I found where it would make sense.

Actually I think for waypoints we could also combine ``height_min`` and ``height_max``.

tsauerwein:
> Actually I think for waypoints we could also combine height_min and height_max.

There you go: #329

PR #322, opened 2016-06-15
Fix migration of versions

tsauerwein:
- The versions for `documents_locales_archives` were not calculated correctly.
- The versions inserted for `documents` and `documents_locales` were not inserted correctly.

Closes https://github.com/c2corg/v6_api/issues/306
Closes https://github.com/c2corg/v6_api/issues/303

PR #328, opened 2016-06-23
Make sure geometry versions are correctly migrated

tsauerwein:
This was missing in https://github.com/c2corg/v6_api/pull/322.
Closes https://github.com/c2corg/v6_api/issues/303

### DIFF
--- a/c2corg_api/scripts/migration/documents/document.py
+++ b/c2corg_api/scripts/migration/documents/document.py
@@ -98,6 +98,12 @@ class MigrateDocuments(MigrateBase):
         document_mapper.version_id_col = None
         document_mapper.version_id_generator = None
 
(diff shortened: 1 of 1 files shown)


asaunier:
OK thanks, looks good.

tsauerwein:
To avoid to run the migration again, you can update your database with:

```sql
with v as (
  select document_id, max(version) as latest_version
  from guidebook.documents_geometries_archives
  group by document_id
)
update guidebook.documents_geometries g
set version = v.latest_version
from v
where v.document_id = g.document_id;
```

asaunier:
OK thanks for the tip. I have run the command in my DB with success.
@ginold you should do the same too or recreate your DB again using mine as template.

PR #329, opened 2016-06-23
Combine height_min/max search fields

tsauerwein:
See https://github.com/c2corg/v6_api/pull/327#issuecomment-228092859


asaunier:
Thanks for the fast change! :)
:+1:

PR #334, opened 2016-06-29
Dockerize

mfournier:
This allows bundling the project in a docker image, without changing the current working habits. The idea would be to make the switch once we're sure the docker images are OK and everyone involved is confortable with this tool. And avoiding getting distracted away from the development duties by docker stuff once this PR is merged.

More in details, this PR will make travis-ci push different flavours of the project to https://hub.docker.com/r/c2corg/v6_api/tags/ where the docker tags will match the git branches and git tags (special case: git `master` branch will be mapped to the `:latest` docker tag).

This allows anyone to simply `docker pull c2corg/v6_api:latest && docker run c2corg/v6_api:latest` and have an instance up and running locally.

Every environment variable passed to the container will override those defined in the `config/*` files. This prevents from hardcoding sensitive information in the docker image. So a more realistic way to run the project in docker would be: `docker run -it --rm -p 8080:8080 -e db_host=pgsql.example.com -e db_password=xxx c2corg/v6_api:latest`


gberaudo:
LGTM

mfournier:
Thanks for taking a look !

So I'll merge this now, as I need it to setup the docker bits of the future infrastructure.

I'm pretty confident this won't have any noticeable effect on the current workflow. If I'm wrong, don't hesitate to revert/disable this docker stuff. Not interrupting the development flow is more important than the infrastructure at this point.

tsauerwein:
Thanks for setting it up, Marc!

Can I still use `make -f ... serve/check` like before?

PR #335, opened 2016-07-01
Fix post multiple images

arnaud-morvan:


arnaud-morvan:
@tsauerwein : can you take a look ?

tsauerwein:
LGTM

PR #336, opened 2016-07-01
Set up caching

tsauerwein:
See https://github.com/c2corg/v6_api/wiki/Caching-strategy#caching-for-the-api-project

* [x] Add a table `cache_versions` which has a version number for each document that reflects the state of the document and its associations.
* [x] Update cache key when a document or its associations change.
* [ ] Update cache key when map or area changes.
* [ ] Check that cache keys are updated when user name is changed (outings, document versions). 
* [x] Set up dogpile.cache for `/{type}/{document_id}` views
* [x] Set cache key as `ETAG` (using [etag_cache](http://docs.pylonsproject.org/projects/pylons-webframework/en/latest/modules/controllers_util.html#pylons.controllers.util.etag_cache))
* [x] Use cache for list views/advanced search
* [ ] Use cache for simple search
* [x] Use cache for history/version views
* [ ] Clear cache script

### DIFF
--- a/c2corg_api/models/cache_version.py
+++ b/c2corg_api/models/cache_version.py
@@ -1,5 +1,6 @@
 import logging
 
+from c2corg_api import caching
 from c2corg_api.models import Base, schema, DBSession
 from c2corg_api.models.document import Document
 from c2corg_api.models.outing import OUTING_TYPE
@@ -25,6 +26,7 @@ class CacheVersion(Base):
     document_id = Column(
         Integer, ForeignKey(schema + '.documents.document_id'),
         primary_key=True)
+
     document = relationship(
         Document, primaryjoin=document_id == Document.document_id
     )
@@ -308,9 +310,11 @@ def _format_cache_key(document_id, lang, version):
         return None
 
     if not lang:
-        return '{0}-{1}'.format(document_id, version)
+        return '{0}-{1}-{2}'.format(
+            document_id, version, caching.CACHE_VERSION)
     else:
-        return '{0}-{1}-{2}'.format(document_id, lang, version)
+        return '{0}-{1}-{2}-{3}'.format(
+            document_id, lang, version, caching.CACHE_VERSION)
 
 
 def get_cache_key(document_id, lang):
--- a/c2corg_api/models/cache_version.py
+++ b/c2corg_api/models/cache_version.py
@@ -302,6 +302,10 @@ def update_cache_version_associations(
 
 
 def get_cache_key(document_id, lang):
+    """ Returns an identifier which reflects the version of a document and
+    all its associated documents. This identifier is used as cache key
+    and as ETag value.
+    """
     version = DBSession.query(CacheVersion.version). \
         filter(CacheVersion.document_id == document_id). \
         first()
@@ -309,7 +313,7 @@ def get_cache_key(document_id, lang):
     if not version:
         # no version for this document id, the document should not exist
         log.debug('no version for document id {0}'.format(document_id))
-        return
+        return None
 
     if not lang:
         return '{0}-{1}'.format(document_id, version[0])
--- a/c2corg_api/models/cache_version.py
+++ b/c2corg_api/models/cache_version.py
@@ -1,5 +1,8 @@
 from c2corg_api.models import Base, schema, DBSession
 from c2corg_api.models.document import Document
+from c2corg_api.models.outing import OUTING_TYPE
+from c2corg_api.models.route import ROUTE_TYPE
+from c2corg_api.models.waypoint import WAYPOINT_TYPE
 from sqlalchemy.orm import relationship
 from sqlalchemy.sql.schema import Column, ForeignKey
 from sqlalchemy.sql.sqltypes import Integer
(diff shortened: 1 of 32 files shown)


tsauerwein:
Ready to review. I will create new issues for the remaining tasks before merging the PR.

c2corg_api/models/cache_version.py
```diff
@@ -0,0 +1,186 @@
...
+    When the document or one of its associated documents is updated, the
+    version of the document is incremented, so that outdated documents are not
+    served from the cache.
+    """
+    __tablename__ = 'cache_versions'
+
+    document_id = Column(
+        Integer, ForeignKey(schema + '.documents.document_id'),
+        primary_key=True)
+    document = relationship(
```
asaunier: Missing blank space maybe :)

c2corg_api/models/cache_version.py
```diff
@@ -0,0 +1,354 @@
...
+BEGIN
+  -- function to update all dependent documents if a document changes
+
+  -- update the version of the document itself
+  PERFORM guidebook.increment_cache_version(p_document_id);
+
+  -- update the version of linked documents (direct associations)
+  PERFORM guidebook.update_cache_version_of_linked_documents(p_document_id);
+
+  if p_document_type = 'w' then
```
asaunier: What about other document types to come (images, articles, profiles, books, accident reports)?
Will they required custom ``update_cache_version_*`` functions?

c2corg_api/models/cache_version.py
```diff
@@ -295,3 +299,19 @@ def update_cache_version_associations(
...
+    version = DBSession.query(CacheVersion.version). \
+        filter(CacheVersion.document_id == document_id). \
+        first()
+
+    if not version:
+        # no version for this document id, the document should not exist
+        log.debug('no version for document id {0}'.format(document_id))
+        return
+
+    if not lang:
```
asaunier: In what cases will we cache documents without considering the lang?

c2corg_api/views/__init__.py
```diff
@@ -157,3 +162,36 @@ def get_best_locale(available_locales, preferred_lang):
                 (available_locales[lang] for lang in langs_priority
                  if lang in available_locales), None)
     return best_locale
+
+
+def etag_cache(request, key):
+    """Use the HTTP Entity Tag cache for Browser side caching
+    If a "If-None-Match" header is found, and equivilant to ``key``,
```
asaunier: "equivalent" maybe?

asaunier:
This is huge! That looks really good to me :)

Since the system is quite complexe, that would probably be useful to have a README somewhere explaining how the cache system works.

I guess changes are required on the UI side to implement the etag approach. Do you plan to also have dogpile cache and etags (for the final end browsers) on the UI side?

c2corg_api/models/cache_version.py
tsauerwein: If they have special dependencies like the ones below, yes.

asaunier:
What happens if a document (nor its associated documents) does not change for a while? It stays cached forever, until a change is done?

What if the cache system (redis?) gets full? Older cache files are removed to save room for new cached documents?

Will it be possible to invalidate a cached document manually using some sort of request?
In v5, moderators have a button to clear a document cache. It is generally not needed except when using some actions that are not designed to invalidate the cache automatically. Perhaps we don't need such a feature in v6 if the cache invalidation is always done.

c2corg_api/models/cache_version.py
tsauerwein: Getting a document with `/{type}/{id}` is a valid request for the API though I think it is not used in the UI.

tsauerwein:
> Since the system is quite complex, that would probably be useful to have a README somewhere explaining how the cache system works.

I am planning to update the caching Wiki page with details.

> What happens if a document (nor its associated documents) does not change for a while? It stays cached forever, until a change is done?
> 
> What if the cache system (redis?) gets full? Older cache files are removed to save room for new cached documents?

All this is managed by Redis. Redis decides if and when a cached value is invalidated.

> Will it be possible to invalidate a cached document manually using some sort of request?

It would be easy to add an API endpoint which simply increments the `cache_key` of a document.

asaunier:
What about a dev mode? Some way to disable the caching system when modifying the code of the API. If documents are cached, it may be difficult to test changes in the code?!

tsauerwein:
> What about a dev mode? Some way to disable the caching system when modifying the code of the API. If documents are cached, it may be difficult to test changes in the code?!

Good point! I made it that a timestamp is appended to the cache key when in dev. mode. So that every time the dev. server re-loads, the cache key changes.

I also made the Git revision part of the cache key, so that the cached values are invalidated when the code changes.

Besides I am thinking about providing a "clear cache" script, which should be run when a new version of the API is deployed, which removes old cache values (basically it clears the Redis database).

asaunier:
> Besides I am thinking about providing a "clear cache" script, which should be run when a new version of the API is deployed, which removes old cache values (basically it clears the Redis database).

That would make sense indeed!

tsauerwein:
Thanks for the review!

Issue #342, opened 2016-07-20
Missing geometries in associated WP list


asaunier:
As for now the API returns the list of associated WP to, for instance, a route. Like this:
```
waypoints: [
{
protected: false,
locales: [
{
version: 2,
title: "La Noire - Pointe 3386",
lang: "fr"
}
],
elevation: 3386,
type: "w",
document_id: 736475,
version: 2,
available_langs: null,
waypoint_type: "summit"
}
],
```
Unfortunately the WP geometry is missing and cannot be used to place the associated WP on the map as expected for issue https://github.com/c2corg/v6_ui/issues/362

PR #358, opened 2016-08-05
Provide geometries of associated documents

asaunier:
Fixes https://github.com/c2corg/v6_api/issues/342

* [x] WP>WP route>WP
* [x] Routes
* [x] Images

### DIFF
--- a/c2corg_api/models/association.py
+++ b/c2corg_api/models/association.py
@@ -3,6 +3,7 @@ from c2corg_api.models.outing import OUTING_TYPE
 from c2corg_api.models.route import Route, RouteLocale, ROUTE_TYPE
 from c2corg_api.models.user import User
 from c2corg_api.models.waypoint import Waypoint, WaypointLocale, WAYPOINT_TYPE
+from c2corg_api.models.document import DocumentGeometry
 from c2corg_api.views import set_best_locale
 from c2corg_api.views.validation import updatable_associations, \
     association_keys, association_keys_for_types
@@ -163,6 +164,8 @@ def _limit_waypoint_fields(query):
 def get_linked_waypoint_parents(document):
     return _limit_waypoint_fields(
         DBSession.query(Waypoint).
+        options(joinedload(Waypoint.geometry).load_only(
+            DocumentGeometry.geom)).
         filter(Waypoint.redirects_to.is_(None)).
         join(Association,
              Association.parent_document_id == Waypoint.document_id).
@@ -173,6 +176,8 @@ def get_linked_waypoint_parents(document):
 def get_linked_waypoint_children(document):
     return _limit_waypoint_fields(
         DBSession.query(Waypoint).
+        options(joinedload(Waypoint.geometry).load_only(
+            DocumentGeometry.geom)).
         filter(Waypoint.redirects_to.is_(None)).
         join(Association,
              Association.child_document_id == Waypoint.document_id).
--- a/c2corg_api/models/association.py
+++ b/c2corg_api/models/association.py
@@ -225,17 +225,20 @@ def get_linked_users(document):
         all()
 
 
-def get_linked_images(document):
-    def limit_route_fields(query):
-        return query.\
-            options(load_only(
-                Image.document_id, Image.filename, Image.author, Image.version,
-                Image.protected)). \
-            options(joinedload(Image.locales).load_only(
-                DocumentLocale.lang, DocumentLocale.title,
-                DocumentLocale.version))
+def _limit_image_fields(query):
+    return query.\
+        options(load_only(
+            Image.document_id, Image.filename, Image.author, Image.version,
+            Image.protected)). \
+        options(joinedload(Image.locales).load_only(
+            DocumentLocale.lang, DocumentLocale.title,
+            DocumentLocale.version)). \
+        options(joinedload(Image.geometry).load_only(
+            DocumentGeometry.geom))
 
-    return limit_route_fields(
+
+def get_linked_images(document):
+    return _limit_image_fields(
         DBSession.query(Image).
         filter(Image.redirects_to.is_(None)).
         join(
--- a/c2corg_api/models/route.py
+++ b/c2corg_api/models/route.py
@@ -322,3 +322,7 @@ schema_association_route = restrict_schema(schema_route, [
     'locales.title', 'locales.title_prefix', 'elevation_min', 'elevation_max',
     'activities'
 ])
+schema_association_waypoint_route = restrict_schema(schema_route, [
+    'locales.title', 'locales.title_prefix', 'elevation_min', 'elevation_max',
+    'activities', 'geometry.geom_detail'
+])
--- a/c2corg_api/models/image.py
+++ b/c2corg_api/models/image.py
@@ -137,7 +137,7 @@ schema_update_image = get_update_schema(schema_image)
 schema_listing_image = restrict_schema(
     schema_image, fields_image.get('listing'))
 schema_association_image = restrict_schema(schema_image, [
-    'filename', 'locales.title'
+    'filename', 'locales.title', 'geometry.geom'
 ])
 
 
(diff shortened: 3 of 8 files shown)


c2corg_api/models/association.py
```diff
@@ -163,6 +163,7 @@ def _limit_waypoint_fields(query):
 def get_linked_waypoint_parents(document):
     return _limit_waypoint_fields(
         DBSession.query(Waypoint).
+        options(joinedload(getattr(Waypoint, 'geometry'))).
```
tsauerwein: If only the default geometry is needed (and not `geom_detail`), you should use `loadonly`, e.g. like here: https://github.com/c2corg/v6_api/blob/dcb10e7b0a2b233332be17c26cf84d6b55e38e36/c2corg_api/models/association.py#L158-L160

asaunier: Do you mean something like 152f98b ?

For the WP we obviously only need the ``geom``. For associated routes (not managed yet), having the default geom is not relevant, if ``geom_detail`` is not available we'd rather not display the route on the map at all. If it is available that could be cool to display the route geom in the WP map. I just wonder what will happen if we have too many route geometries: will it make the map hard to read? will it make the page too heavy?

c2corg_api/models/association.py
```diff
@@ -163,7 +164,8 @@ def _limit_waypoint_fields(query):
 def get_linked_waypoint_parents(document):
     return _limit_waypoint_fields(
         DBSession.query(Waypoint).
-        options(joinedload(getattr(Waypoint, 'geometry'))).
+        options(joinedload(getattr(Waypoint, 'geometry')).load_only(
```
tsauerwein: You use `Waypoint.geometry` instead of `getattr(...)`.

c2corg_api/models/association.py
tsauerwein: > If it is available that could be cool to display the route geom in the WP map. I just wonder what will happen if we have too many route geometries: will it make the map hard to read? will it make the page too heavy?

Something we would have to test.

asaunier:
I would like to do something similar for associated routes and update ``get_linked_routes`` https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/association.py#L193 like this:
```
     return limit_route_fields(
         DBSession.query(Route).
+        options(joinedload(Route.geometry).load_only(
+            DocumentGeometry.geom_detail)).
         filter(Route.redirects_to.is_(None)).
         join(Association, condition)). \
         all()

```
but then I get the following error:
> sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <Route at 0x7f9b245497b8> is not bound to a Session; lazy load operation of attribute 'geometry' cannot proceed

I don't really see the difference with what I have done for waypoints?

tsauerwein:
> I don't really see the difference with what I have done for waypoints?

This sounds like you used `geometry` instead of `geom_detail` in the schema.

asaunier:
You mean in
```
schema_association_route = restrict_schema(schema_route, [
    'locales.title', 'locales.title_prefix', 'elevation_min', 'elevation_max',
    'activities', 'geometry.geom_detail'
])
```
?

I have also tried with ``geom`` instead of ``geom_detail`` in both files and I got the same error.

c2corg_api/models/association.py
```diff
@@ -212,6 +212,8 @@ def get_linked_routes(document):
 
     return limit_route_fields(
         DBSession.query(Route).
+        options(joinedload(Route.geometry).load_only(
+            DocumentGeometry.geom_detail)).
```
tsauerwein: You also have to change it here:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/waypoint.py#L316

asaunier: Thanks! Done at 9eb281b

asaunier:
I have pushed a commit e65af61 that gets the geometry of associated images as well. At least only when getting the images associated to an outing (I thought showing images markers on the map was only relevant in outing detail view pages).

=> this works as expect when viewing an outing (eg. http://c2corgv6-demo.gis.internal:6545/outings/636252?l=fr) but fails for other document types (eg. http://c2corgv6-demo.gis.internal:6545/routes/397620?l=fr) because
> sqlalchemy.orm.exc.DetachedInstanceError: Parent instance <Image at 0x7f1d66f03710> is not bound to a Session; lazy load operation of attribute 'geometry' cannot proceed

=> it seems that listing ``geometry.geom`` in ``schema_association_image``
https://github.com/c2corg/v6_api/blob/e65af61aa3b18303e99111261e042bbe494162b8/c2corg_api/models/image.py#L139-L141 is not possible unless the attribute is actually provided (OK for outings, NOK for others). Does this mean image geometries should always be provided whatever the associated document type? It would make sense, of course, but I just wanted to save the retrieval of this attribute if the UI only uses it for outings.

tsauerwein:
> Does this mean image geometries should always be provided whatever the associated document type? It would make sense, of course, but I just wanted to save the retrieval of this attribute if the UI only uses it for outings.

If not you could create a separate schema for images of outings.

asaunier:
I don't know. It would make sense to provide the coords of point documents and let the frontend applications decide whether they will use them or not. It depends if it may cause a significantly highler load on the server (perhaps not that a problem with the cache system anyway?).

asaunier:
I have updated and rebased the PR. In brief:
* the ``geom`` attribute is now provided in the list of associated waypoints and images
* the ``geom_detail`` attribute is now provided (if not set, ``null`` is returned) in the list of associated routes to a waypoint (only for WP-route links).

c2corg_api/models/association.py
tsauerwein: Do you still need to load `geom_detail` here now that you have a 2nd schema?

c2corg_api/models/route.py
```diff
@@ -322,3 +322,7 @@ class ArchiveRouteLocale(_RouteLocaleMixin, ArchiveDocumentLocale):
     'locales.title', 'locales.title_prefix', 'elevation_min', 'elevation_max',
     'activities'
 ])
+schema_association_waypoint_route = restrict_schema(schema_route, [
+    'locales.title', 'locales.title_prefix', 'elevation_min', 'elevation_max',
+    'activities', 'geometry.geom_detail'
```
tsauerwein: For routes you are loading the detail geometry, that is the full track. We will have to see how well it works to load up to 30 full tracks.

asaunier: Loading the simple geom (point) for associated routes is useless: the cool thing is to display the tracks of the associated routes in the map of a WP detail view.
But you're right: not sure what will happen if there are too many routes with a ``geom_detail``. I think it's worth trying.

tsauerwein: > the cool thing is to display the tracks of the associated routes in the map of a WP detail view.

You will only see the first 30. This might be misleading.

asaunier: Most WP have less than 30 routes. And it would be really useful to visualize on the map the routes that go to / come from the current WP. That would help to figure out what route you're looking for when accessing the WP detail view.

For WP with many routes, you're probably right. But showing a subset of tracks on the map is consistent with showing a subset of cards in the associated routes section.

asaunier: Concerning the route ``geom_detail`` we should probably do something to avoid huge GPS tracks, that could be a problem event in the route detail view page. For instance:
http://c2corgv6.demo-camptocamp.com/routes/211687/fr
http://www.camptocamp.org/routes/211687/fr/mont-de-grange-traversee-bel-air-coicon
=> the GPX track in the v5 page is > 800 kB

c2corg_api/models/association.py
asaunier: You mean like this https://github.com/c2corg/v6_api/pull/358/commits/cd1ea2f60384c0c56a12d71983a005501ae95277

It seems indeed to work without loading ``geom_detail`` but I don't understand why. The second schema requires this attribute. Isn't it that second schema that is used with got getting the associatied routes of a WP? https://github.com/c2corg/v6_api/blob/associated-wp-geom/c2corg_api/models/association.py#L115-L140

tsauerwein:
I am still a bit worried about including the route tracks, but we will see. Feel free to merge.

Issue #361, opened 2016-08-11
Add clear cache script


tsauerwein:
The cache key includes the current Git revision, so that old cache entries are not used any more, if the code has changed. Additionally, the cache database could be cleared completely when deploying a new version to free space.

Same for the UI cache.

PR #365, opened 2016-08-12
Add a clear cache script

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/361

We still have to see with @mfournier how this script can be integrated in the deployment process, so that the cache is cleared when deploying a new version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,14 @@ run-background-jobs: install development.ini
 run-background-jobs-prod: install production.ini
 	.build/venv/bin/python c2corg_api/scripts/jobs/scheduler.py production.ini
 
+.PHONY: clear-cache
+clear-cache: install development.ini
+	.build/venv/bin/python c2corg_api/scripts/redis-flushdb.py development.ini
+
+.PHONY: clear-cache-prod
+clear-cache-prod: install production.ini
+	.build/venv/bin/python c2corg_api/scripts/redis-flushdb.py production.ini
+
 .PHONY: upgrade
 upgrade: .build/venv/bin/pip
 	.build/venv/bin/pip install --upgrade -r requirements.txt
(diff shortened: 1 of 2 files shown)


asaunier:
Looks good to me.

Just a question: what happens if the wrong clear cache command is run. For instance ``clear-cache-prod`` in a dev instance or ``clear-cache`` in prod?
Since prod is the "real" application, wouldn't it make more sense that ``clear-cache`` is the prod command (``clear-cache-dev`` for the dev). Or ``clear-cache-prod``/``clear-cache-dev``?

tsauerwein:
> Just a question: what happens if the wrong clear cache command is run. For instance clear-cache-prod in a dev instance or clear-cache in prod?

In the worst case, the cache entries have to be recreated on the next request.

> Since prod is the "real" application, wouldn't it make more sense that clear-cache is the prod command (clear-cache-dev for the dev). Or clear-cache-prod/clear-cache-dev?

I followed the naming schema that we used for other targets (run-syncer/run-syncer-prod or run-background-jobs/run-background-jobs-prod). I don't mind about the name. I let you change it in a separate PR if you do. :)

asaunier:
OK if it's more consistent.
And it's probably "safer": if one runs the "clear-cache" command in prod it will not actually impact the prod cache, right? The explicit "clear-cache-prod" command is required.

tsauerwein:
Actually it is more important which Makefile config you use.

On prod, you will use: `make -f config/main clear-cache-prod`.
On your test instance: `make -f config/alex clear-cache`.

But `make -f config/alex clear-cache` and `make -f config/alex clear-cache-prod` will clear the same Redis database. Because currently all dev configs share the same Redis db for dev and prod (with different cache keys so that the value do not get mixed up).

Issue #362, opened 2016-08-11
Make sure cache key is updated correctly


tsauerwein:
* [x] Update cache keys when map changes.
* [x] Update cache key when area changes.
* [x] Check that cache keys are updated when user name is changed (see https://github.com/c2corg/v6_api/pull/366)
  * [x] for outings
  * [x] for document versions

Related to https://github.com/c2corg/v6_api/pull/336

PR #366, opened 2016-08-12
Update cache keys when user name is changed

tsauerwein:
Related to https://github.com/c2corg/v6_api/issues/362


asaunier:
Looks good.
I guess we have to reimport the DB if we want to push this change to demo?

tsauerwein:
It would be enough to re-create the stored procedures:
https://github.com/c2corg/v6_api/blob/a46ff8f540f8ad603f2abfdd04c0629c17a1edc8/c2corg_api/models/cache_version.py#L40-L275

If not, this is not very critical. It could be done the next time we have to run the migration again.

Thanks for the review!

asaunier:
OK!

I have updated the demo and recreated the procedures by hand in the DB.

tsauerwein:
Cool, thanks!

Issue #363, opened 2016-08-11
IntegrityError when migrating the images data from v5 to v6


asaunier:
When running the v5>v6 migration script, the latter crashes because

> sqlalchemy.exc.IntegrityError: (psycopg2.IntegrityError) duplicate key value violates unique constraint "images_filename_key"
> DETAIL:  Key (filename)=(1279011081_1740341383.jpg) already exists.
>  [SQL: 'INSERT INTO guidebook.images (activities, image_type, document_id, filename) VALUES (CAST(%(activities)s AS guidebook.activity_type[]), %(image_type)s, %(document_id)s, %(filename)s)'] [parameters: {'image_type': 'personal', 'filename': '1279011081_1740341383.jpg', 'document_id': 227184, 'activities': ['snow_ice_mixed']}]

Seems related to that PR: https://github.com/c2corg/v6_api/pull/331

asaunier:
Temporary workaround applied in the demo API code so that the migration passes:
```
diff --git a/c2corg_api/models/image.py b/c2corg_api/models/image.py
index 647d88c..03b9bef 100644
--- a/c2corg_api/models/image.py
+++ b/c2corg_api/models/image.py
@@ -49,8 +49,7 @@ class _ImageMixin(object):
     @declared_attr
     def filename(self):
         return Column(String(30),
-                      nullable=False,
-                      unique=(self.__name__ == 'Image'))
+                      nullable=False)
```

asaunier:
Is there a chance that this issue is fixed for alpha3?
I wanted to reimport the demo DB but I notice we still have this problem. If it is about to be fixed, I will wait before running the migration script. Else I could do it now.

arnaud-morvan:
I do not have worked on this for now. I'm pretty sure you can reimport it now.

asaunier:
Should we do something about this issue?

tsauerwein:
We should check which images have this problem, and why. Maybe it could be a simple manual fix in the old database.

asaunier:
> DETAIL: Key (filename)=(1279011081_1740341383.jpg) already exists.

```
c2corg=# select id, filename from images where filename = '1279011081_1740341383.jpg';
   id   |         filename          
--------+---------------------------
 227183 | 1279011081_1740341383.jpg
 227184 | 1279011081_1740341383.jpg
(2 rows)
```
It seems that 2 image documents share the same filename:
http://www.camptocamp.org/images/227183/fr/0007-
[...]

PR #340, opened 2016-07-07
Create forum topic

arnaud-morvan:
Superseed #326

* Create table documents_topics
* Add script to fill documents_topics from discourse database using dblink
* Add topic_id in GET document response
* Add view to create guidebook related topics in discourse

PR #367, opened 2016-08-12
Update cache keys when areas or maps change

tsauerwein:
When a map or area changes, the cache keys of associated documents have to be updated.

Associations between maps and documents are now also stored in a table (same as for areas), because maps do not change that often.

Closes https://github.com/c2corg/v6_api/issues/362

The migration script has to be run again, because changes to the database were made.

### DIFF
--- a/c2corg_api/models/__init__.py
+++ b/c2corg_api/models/__init__.py
@@ -30,6 +30,7 @@ from c2corg_api.models import association  # noqa
 from c2corg_api.models import topo_map  # noqa
 from c2corg_api.models import area  # noqa
 from c2corg_api.models import area_association  # noqa
+from c2corg_api.models import topo_map_association  # noqa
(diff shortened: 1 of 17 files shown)


asaunier:
LGTM

> The migration script has to be run again, because changes to the database were made.

Should we run the migration script from scratch in the demo DB or could I simply apply the SQL commands found in https://github.com/c2corg/v6_api/pull/367/files ?

tsauerwein:
There also is a new table, which is filled during the migration.

asaunier:
So it means a full migration is needed?!
Can we wait that https://github.com/c2corg/v6_api/issues/363 is fixed before reimporting the DB?

tsauerwein:
Sure, I can wait with merging this PR.

tsauerwein:
I am merging now, because some changes are required for https://github.com/c2corg/v6_api/pull/340.

Issue #364, opened 2016-08-11
Simple ID service for documents is needed


asaunier:
We need a service that returns very basic data about a document. Eg. title (+title_prefix), id, lang, slug ?!
If the service receives no lang param but a prefered language param, it should figure what is the best locale available.

See https://github.com/c2corg/v6_ui/pull/558

asaunier:
@tsauerwein What do you think is the best approach: have a dedicated web service or use the usual "get document" service with some parameter that would skip the associations and return a subset of attribute (a bit like ``?e=1``?).

To get a subset of attributes, should we use a special schema for each document type?

tsauerwein:
I would create a new web-service. And I don't think we have to create a schema for each type, because you have the same attributes for all types (e.g. `title_prefix` for routes).

tsauerwein:
As discussed with @asaunier: The service accepts a `document_id` and a `lang`. It is supposed to return:

```
{document_id: ..., locales: [{lang: 'en', title: '...', title_prefix: '...']}
```

If the document does not have a locale in the given `lang`, the "best locale" is returned.

PR #373, opened 2016-08-19
Add a document info service

tsauerwein:
Which only returns document_id, title, lang and tile_prefix (for routes).

Closes https://github.com/c2corg/v6_api/issues/364

### DIFF
--- a/c2corg_api/views/document_info.py
+++ b/c2corg_api/views/document_info.py
@@ -1,0 +1,105 @@
+from c2corg_api.caching import cache_document_info
+from c2corg_api.models import DBSession
+from c2corg_api.models.cache_version import get_cache_key
+from c2corg_api.models.document import DocumentLocale, Document, \
+    get_available_langs
+from c2corg_api.models.route import Route, RouteLocale
+from c2corg_api.views import etag_cache, \
+    set_best_locale
+from pyramid.httpexceptions import HTTPNotFound
+from sqlalchemy.orm import contains_eager, load_only, joinedload
+from sqlalchemy.orm.util import with_polymorphic
+
+
+class DocumentInfoRest(object):
+    """ Base class for all views that return a basic/info version for a
+    document, that only contains the document_id and title/title_prefix
+    of the requested locale.
+    This view is used by the UI to generate URL slugs.
+
+    If the locale is not available in the requested lang, a locale in the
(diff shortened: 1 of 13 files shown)


c2corg_api/views/document_info.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class DocumentInfoRest(object):
+    """ Base class for all views that return a basic/info version for a
+    document, that only contains the document_id and title/title_prefix
+    of the requested locale.
+    This view is used by the UI to generate URL slugs.
+
+    If the locale is not available in the requested lang, a locale in the
+    "best" lang is returned.
```
asaunier: OK, so it's the responsibility of the UI to provide at least its interface language if the lang param is not already available in the URL to slugify, right?

tsauerwein: Yes.

c2corg_api/views/document_info.py
```diff
@@ -0,0 +1,105 @@
...
+    If the requested document is redirected, the document id and available
+    languages of the redirection document is returned.
```
asaunier: Cool!
But this info service is only called when the URL slug is missing. For a full URL it won't be called and the UI won't know that it should use another document instead. Or does it?

tsauerwein: I am not sure I understand? Can you give an example?

asaunier: Say we display the detail view page of a waypoint that has actually been merged (redirected) with another one.
If the URL has no slug, the UI will call this info service that will return the info of the second waypoint.
But if the URL has a slug, the service won't be called and the UI will display the first waypoint instead of redirecting the the second one, won't it?

asaunier: It's the use case described in https://github.com/c2corg/v6_ui/issues/190
You said that 
> When a merged document is requested, the API will return:
{"redirects_to": 12345, "available_langs": ["en", "fr"]}

So the API side is OK. I guess we have to make the UI handle those redirections.

So your PR looks good to me. Feel free to merge, thanks.

tsauerwein: Yes, this has to be handled on the UI side.

tsauerwein:
Thanks for the review.

Issue #368, opened 2016-08-16
Lock errors with Redis


arnaud-morvan:
I get this errors from v6_api:
```
  File "/home/amorvan/c2corgv6/v6_api/c2corg_api/views/waypoint.py", line 158, in get
    set_custom_associations=set_custom_associations)
  File "/home/amorvan/c2corgv6/v6_api/c2corg_api/views/document.py", line 230, in _get
    cache_key, create_response, expiration_time=-1)
  File "/home/amorvan/c2corgv6/v6_api/.build/venv/lib/python3.4/site-packages/dogpile/cache/region.py", line 657, in get_or_create
    async_creator) as value:
  File "/home/amorvan/c2corgv6/v6_api/.build/venv/lib/python3.4/site-packages/dogpile/lock.py", line 154, in __enter__
    return self._enter()
  File "/home/amorvan/c2corgv6/v6_api/.build/venv/lib/python3.4/site-packages/dogpile/lock.py", line 94, in _enter
    generated = self._enter_create(createdtime)
  File "/home/amorvan/c2corgv6/v6_api/.build/venv/lib/python3.4/site-packages/dogpile/lock.py", line 149, in _enter_create
    self.mutex.release()
  File "/home/amorvan/c2corgv6/v6_api/.build/venv/lib/python3.4/site-packages/redis/lock.py", line 135, in release
    self.do_release(expected_token)
  File "/home/amorvan/c2corgv6/v6_api/.build/venv/lib/python3.4/site-packages/redis/lock.py", line 264, in do_release
    raise LockError("Cannot release a lock that's no longer owned")
redis.exceptions.LockError: Cannot release a lock that's no longer owned
```

arnaud-morvan:
This seems to be due to timeout when using a debugger. We could increase the timeout as a workaround for development :
https://github.com/c2corg/v6_api/blob/05753ab8d8f36129f44ce12b71a724c3f1cd4634/c2corg_api/caching.py#L67

PR #374, opened 2016-08-19
Increase lock timeout for cache

tsauerwein:
dogpile.cache uses a lock to avoid the dogpile effect. Currently, there is a timeout of 5 seconds for this lock. This means if the cache value (e.g. the rendered response) is not created in 5 seconds, there might be an error. To accommodate for our slow demo server, I increased the timeout to 15 seconds.

Closes https://github.com/c2corg/v6_api/issues/368

I will do the same for the UI.

### DIFF
--- a/c2corg_api/caching.py
+++ b/c2corg_api/caching.py
@@ -66,7 +66,7 @@ def configure_caches(settings):
             arguments={
                 'connection_pool': redis_pool,
                 'distributed_lock': True,
-                'lock_timeout': 5  # 5 seconds (dogpile lock)
+                'lock_timeout': 15  # 15 seconds (dogpile lock)
             },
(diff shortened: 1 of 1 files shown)


asaunier:
If the prod server is more performant, could we set this value back to 5 sec in the future?
In the meantime, LGTM.

tsauerwein:
> If the prod server is more performant, could we set this value back to 5 sec in the future?

I tend to say that a more tolerant timeout is probably not a fault. In the normal case, the cache value should be generated fast. Only if the server is under heavy load, or if connections to the database/ES fail, the lock might time out. In that case, you probably want to make sure that not too many requests try to generate the cache value. So a longer timeout should be no harm (only if it is too long).

PR #378, opened 2016-08-31
Don't use latest version of zope.interface

tsauerwein:
Same as https://github.com/c2corg/v6_ui/pull/584


tsauerwein:
Merging without review to get our build working again.

Issue #376, opened 2016-08-31
Include upload-user for image detail view


tsauerwein:
Required for https://github.com/c2corg/v6_ui/pull/575

Include the user who uploaded the image.

PR #380, opened 2016-08-31
Include creator for single images

tsauerwein:
Include the creator, the user who uploaded the image, when requesting a single image.

Closes https://github.com/c2corg/v6_api/issues/376


asaunier:
👍

Issue #344, opened 2016-07-26
Handle broken connections in syncer scripts


tsauerwein:
Check how the syncer scripts deal with broken connections to the database.

See https://github.com/c2corg/v6_ui/issues/522#issuecomment-234980949

PR #381, opened 2016-09-01
Handle broken database connections

tsauerwein:
Validate connections in the database connection pool before using them. Otherwise errors might be raised when a broken connection after restarting the database is used.

I also tested the syncer/background-job scripts. Both continue to run if Postgres/Redis/ElasticSearch is restarted.

Closes https://github.com/c2corg/v6_api/issues/344


asaunier:
Looks good, thanks.

PR #382, opened 2016-09-01
Use fragment param for /auth page

tsauerwein:
Related to https://github.com/c2corg/v6_ui/pull/587


asaunier:
Is it related to the Discourse SSO URL you're talking about in https://github.com/c2corg/v6_ui/pull/587 ?

tsauerwein:
No, this is about the links that are contained in the mails sent by the API.

Issue #348, opened 2016-07-27
Consistent results attributes in advanced search/simple search/association requests


asaunier:
Results from various services such as the advanced search, the simple search or the list of associated documents are displayed in the UI using cards. For a given document type, a shared card partial is used.

Problem: all services do not return the same sets of attributes and some attributes are cannot be displayed in the cards. See for instance https://github.com/c2corg/v6_ui/pull/512#issue-166835614

It would be more consistent to have the same set of attributes for a given type of results, whatever the service used (with the adapted extern definition).

Probably also related (duplicate?) to https://github.com/c2corg/v6_api/issues/300

tsauerwein:
See also https://github.com/c2corg/v6_api/issues/192

fjacon:
What is the priority of this ?

asaunier:
Quite high, else the cards in a document detail view page (associated documents) are not provided all the attributes they should display.

Compare for instance the list of route cards
* in the advanced search: http://c2corgv6.demo-camptocamp.com/routes?w=104251&bbox=-935508%2C3369261%2C1735507%2C7830738
* in a WP page: http://c2corgv6.demo-camptocamp.com/waypoints/104251/fr (some attributes are missing)

tsauerwein:
I thought about this, I was a bit hesitant because I wanted to keep the results for the simple search as slim as possible.

But I think it is a good idea to use the same schema for the simple search, the advanced search and the associations. This makes it possible to use the same cache for the different API endpoints, it will simplify the code and you only have one place to configure the included attributes.

And I think it still makes sense to have a simple search and an advanced search. The simple search allows to search for multiple document types at the same time, while the advanced search is only for a single one. The simple search only searches in document titles, while the advanced search also scans the full document summary and description.

Issue #321, opened 2016-06-15
User document_id


ginold:
Similar issue to the previous I created:

users have a `document_id` normally, whereas for saving an outing the API expects to receive `id` instead of `document_ids `.
Would it be possible to make it more consistent?

tsauerwein:
What you are getting in `associations.users` are [User](https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/user.py) objects, which do not have a `document_id`. I understand that you have to treat user associations differently in the UI, but the API would also have to explicitly add a document id.

Issue #300, opened 2016-06-01
Getting a document "list attributes"?


asaunier:
I am making changes in the UI so that the AngularJS card directive and partials (used for the advanced search) are used also for:
* listing associated documents as cards (instead of Mako) => done at https://github.com/c2corg/v6_ui/pull/328
* adding a new card when a document has just been associated.

For the latter usage I am in troubles because the data returned by the association tool (based upon the simple search service) does not return all the attributes needed for the "listing" cards (in addition the "externs" types are different).

2 options:
* when creating the newly associated document card, I make a request to the API to get the document attributes. I could make a standard document request but it's a bit overkill because I only need the "listing" attributes => is it possible to make a "list" request to the API for a given document id. For instance ``waypoints?w=1234``?
* we use consistent data structure for the listing and simple-search request, with consistent "externs" types.

What do you think?

tsauerwein:
If I get you right, the actual problem is that the simple search service does not return all required attributes?

The simple search service is used in different places, for the actual search at the top of the page and to find new associations. Either we use the same schema for all these uses, or we let the simple search service return different attributes depending on the use. E.g. `/search?q=...&l=1` could return all attributes required for listings.

See also https://github.com/c2corg/v6_api/issues/192

asaunier:
> See also #192

That's indeed the same issue (duplicate? Should I move this one as a comment of #192 ?).

> If I get you right, the actual problem is that the simple search service does not return all required attributes?

As for now in https://github.com/c2corg/v6_ui/pull/328 I have been able to create cards when associating new documents with the simpleSearch/association tool and had no error. Maybe that's because the cards are currently quite simple(?). Maybe we'll have troubles later when adding more info to the cards? So it seems the problem is not that urgent.

Anyway it is probably a good thing that simple search and advanced search (=listing) responses are consistent (and share the same types in the UI?).

 > or we let the simple search service return different attributes depending on the use. E.g. /search?q=...&l=1 could return all attributes required for listings.

Maybe. But I wonder if it is really a good idea to add special cases etc.

asaunier:
> I have been able to create cards when associating new documents with the simpleSearch/association tool and had no error.

Since I rebased c2corg/v6_ui#328 to take into account the merging of https://github.com/c2corg/v6_ui/pull/372 which introduced externs for document types (eg. https://github.com/c2corg/v6_ui/blob/master/c2corg_ui/externs/appx.js#L227-L280), Closure complains for instance because
[...]

Issue #192, opened 2016-03-18
Return the same fields for simple search as for listings?


tsauerwein:
Currently the schema for the documents returned in the simple search is the same as in listings. Are all these fields required? Currently only the title is used in the UI, so I wonder if we could omit the other fields (like geometry, elevation, activities, ...).

asaunier:
Maybe geometry yes. But we will need additional info to distinguish the homonym documents.
For instance we have 2 "Mont Blanc".

Giving the area and elevation would help for WP.
For routes, we would need the activities and the ratings I think.

PR #391, opened 2016-09-08
Use consistent attributes

tsauerwein:
... for the simple search, advanced search and associated document in detail pages.

The big advantage, beside the consistency and the simpler code, is that all different services use the same cache. For example if the simple search returns a document, this document will be cached. If the document is returned again in the advanced search or as association, it can directly be loaded from the cache.

The responses from the API have changed in some places:

- Users in outings now have their id as `document_id` instead of `id` (closes https://github.com/c2corg/v6_api/issues/321)
- Route associations for waypoints no longer include `geometry_detail`
- The structure for `recent_outings` has been changed from `'recent_outings': {'outings': ...}}` to  `'recent_outings': {'documents': ...}}` (same for `all_routes`). 

TODO

* [x] I still want to optimize the query to get the documents to only include the attributes  that are needed. But I would like to this in a separate PR.

Closes https://github.com/c2corg/v6_api/issues/348
Closes https://github.com/c2corg/v6_api/issues/300
Closes https://github.com/c2corg/v6_api/issues/192
Closes https://github.com/c2corg/v6_api/issues/93


asaunier:
Is this PR still WIP? Let us know when you need a review.

> The responses from the API have changed in some places:

Those changes probably requires several adaptations in the UI. Perhaps we should wait alpha3 is released before merging the PR and do the changes in the UI?

tsauerwein:
> Is this PR still WIP? Let us know when you need a review.

No, it's ready to review. Probably you want to review commit-by-commit, because I did some heavy refactorings.

> Perhaps we should wait alpha3 is released before merging the PR and do the changes in the UI?

Ok for me.

tsauerwein:
> I still want to optimize the query to get the documents to only include the attributes that are needed. But I would like to this in a separate PR.

I have done it in this PR. Please review.

asaunier:
I will take a look soon.

asaunier:
It seems your last commit 6a388b8 breaks the test.
Else the PR looks good to me.

I have created a UI issue for listing the changes we will have to make in the UI when the PR is merged: https://github.com/c2corg/v6_ui/issues/619
Can you think of other changes that will have to be done?

tsauerwein:
Thanks for your review!

I forgot to update the common project, the tests should now pass.

> Can you think of other changes that will have to be done?

No, I think I listed all points that have changed above.

tsauerwein:
@dkocich Now that this PR is merged, you could rebase your branch on the current master.

Issue #390, opened 2016-09-08
The "info" service does not work for images


asaunier:
In the UI, image detail view URLs with no slug are 404 instead of redirecting to the URLs with a slug (which do work - 200). That's because the call to the API info service seems not available for images:

http://c2corgv6-demo.gis.internal:6545/images/736131/fr/info

> {
  status: "error",
  errors: [{
    description: "/images/736131/fr/info",
    name: "Not Found",
    location: "request"
  }]
}

PR #392, opened 2016-09-08
Add info service for images, maps and areas

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/390


asaunier:
Cool, thanks! Please merge.
Does the info service now work for all document types?

tsauerwein:
I added it also for maps and area. There is no info service for user profiles, but this would require authentication (?), so I guess it is not useful.

asaunier:
:+1:

PR #396, opened 2016-09-15
Add config for the demo@c2corg instance

asaunier:
See also https://github.com/c2corg/v6_ui/pull/624

The "secret" settings are the same than in https://github.com/c2corg/v6_api/blob/demo-config/config/default#L69-L75 and should probably be changed. Maybe also set using a more private way :P


mfournier:
Merged to alpha3 branch as ebf22fc

Issue #403, opened 2016-09-28
Activity feed


tsauerwein:
The activity feed contains changes for the following actions:

1. A user creates a new document (all document types except outing or image)
2. A user creates a new outing.
3. A user is associated to an existing outing.
4. A set of images is uploaded and associated to a document.

**TODO**

* [x] Set up database model
 * [x] Set up triggers for PK-FK-relation check for `users_ids` and `area_ids` in `DocumentChange`
 * [x] Review indices on feed table
 * [x] Add option `followed_users_updates_only` (maybe with a shorter name `followed_only`?) see https://github.com/c2corg/v6_api/issues/403#issuecomment-250703021
* [x] Import existing data
* [x] Test if performance is ok with existing data
* [x] Update web-services so that the feed tables are updated for the above actions (see https://github.com/c2corg/v6_api/pull/431)
  * [x] When a new document is created (no user or image).
  * [x] When a new user is associated to an outing.
  * [x] When a group of images is uploaded.
* [x] Implement web-services
  * [x] to get feed for homepage (public)
  * [x] to get feed for homepage (authenticated)
  * [x] to get feed for user-profile
  * [x] to follow user (see #401)
  * [x] to get info if following a user (see #401)
  * [x] to get/update filter preferences (see #402)
* [x] Add parameter to get personal homepage feed without filters (see https://github.com/c2corg/v6_api/pull/409#discussion_r82415016)
* [x] Deal with redirected documents

asaunier:
One more thing: would it be possible to take the "quality" attribute into account when filling the feed?

One of the goals of this "quality" field was to strip empty outings (poor quality) out of the homepage for instance. It means that outings (or any documents) created with quality < draft (?) should not be added to the table ``FeedDocumentChanges`` (unless its ``quality`` is upgraded?). What do you think?

asaunier:
The default quality when creating an outing is "draft".
I guess many people will forget to upgrade the quality and their outings won't be shown in the home feed...

tsauerwein:
> One of the goals of this "quality" field was to strip empty outings (poor quality) out of the homepage for instance. It means that outings (or any documents) created with quality < draft (?) should not be added to the table FeedDocumentChanges (unless its quality is upgraded?). What do you think?

This is more a UX problem, encouraging users only to create outings that have content.

asaunier:
> This is more a UX problem, encouraging users only to create outings that have content.

[...]

Issue #402, opened 2016-09-28
Add feed filter preferences web-services


tsauerwein:
Add web-services to get/update feed filter preferences (area + activities).

asaunier:
Here are possible specs/examples for the  API services needed to get/set the user's preferred filters.

* Getting the user's filters:
```
GET http://api/users/preferences?user_id=1234
{
  "activities": ["skitouring", "hiking"],
  "areas": [12345, 23456]
}
```
* Setting/updating the prefs:
```
POST/PUT? http://api/users/preferences?user_id=1234
{
  "activities": ["skitouring", "hiking"],
  "areas": [12345, 23456]
}
```
or should we pass the user id inside the JSON object?
* to build the pref UI, we also need to get the list of areas, as described in https://github.com/c2corg/v6_ui/issues/598#issuecomment-250695229
 * what about simply using the usual areas list request http://c2corgv6-demo.gis.internal:6545/areas ?
 * filtering on the ``area_type`` is needed (ditto for the area advanced search, see https://github.com/c2corg/v6_api/issues/419)
 * we would probably need to list all areas of a given ``area_type`` (with no limit as in the current list request)?

asaunier:
> to build the pref UI, we also need to get the list of areas,

Maybe we don't if the pref UI only use a search tool for areas, see https://github.com/c2corg/v6_ui/issues/598#issuecomment-252208853

Then it means we have to make sure the areas are actually indexed by ES and supported by the simple search / association tool.

tsauerwein:
> or should we pass the user id inside the JSON object?

The user-id is not needed at all. It can be obtained from the authorization header.

asaunier:
Great, this way we can also make sure that only the user can access his/her prefs.
I have updated my comment above.

Issue #401, opened 2016-09-28
Add "follow" web-services


tsauerwein:
Add web-services to:

- check if the authenticated user follows a given user.
- follow/unfollow a user.

PR #409, opened 2016-09-30
[WIP] Set up activity feed

tsauerwein:
**TODO**

* [x] Set up database model
 * [x] Set up triggers for PK-FK-relation check for `users_ids` and `area_ids` in `DocumentChange`
 * [ ] Review indices on feed table
 * [x] Add option `followed_users_updates_only` (maybe with a shorter name `followed_only`?) see https://github.com/c2corg/v6_api/issues/403#issuecomment-250703021
* [x] Import existing data
* [x] Test if performance is ok with existing data
* [ ] Update web-services so that the feed tables are updated for the above actions
  * [ ] When a new document is created (no user or image).
  * [ ] When a new user is associated to an outing.
  * [ ] When a group of images is uploaded.
* [ ] Implement web-services
  * [x] to get feed for homepage (public)
  * [x] to get feed for homepage (authenticated)
  * [x] to get feed for user-profile
  * [ ] to follow user (see #401)
  * [ ] to get info if following a user (see #401)
  * [ ] to get/update filter preferences (see #402)
* [ ] Add parameter to get personal homepage feed without filters (see https://github.com/c2corg/v6_api/pull/409#discussion_r82415016)
* [ ] Deal with redirected documents

Closes https://github.com/c2corg/v6_api/issues/403

### DIFF
--- a/c2corg_api/models/feed.py
+++ b/c2corg_api/models/feed.py
@@ -11,6 +11,7 @@ from sqlalchemy.sql.functions import func
 from sqlalchemy.sql.schema import Column, ForeignKey, PrimaryKeyConstraint, \
     CheckConstraint, Index
 from sqlalchemy.sql.sqltypes import Integer, String, DateTime, Boolean
+from sqlalchemy import event, DDL
 
 
 class FilterArea(Base):
@@ -107,7 +108,6 @@ class DocumentChange(Base):
         ArrayOfEnum(enums.activity_type), nullable=False, server_default='{}')
 
     # ids of the areas where this change happened
-    # TODO add triggers to areas/users to check pk-fk-relation
     area_ids = Column(ARRAY(Integer), nullable=False, server_default='{}')
 
     # ids of the users that were involved in this change (e.g. the user that
@@ -142,3 +142,82 @@ class DocumentChange(Base):
             postgresql_using='gin'),
         Base.__table_args__
     )
+
+# For performance reasons, areas and users are referenced in simple integer
+# arrays in 'feed_document_changes', no PK-FK relations are set up. To prevent
+# inconsistencies, triggers are used.
+trigger_ddl = DDL("""
+-- when creating a change, check that the given user and area ids are valid
+CREATE OR REPLACE FUNCTION guidebook.check_feed_ids() RETURNS TRIGGER AS
+$BODY$
+DECLARE
+  user_id int;
+  area_id int;
+BEGIN
+  -- check user ids
+  FOREACH user_id IN ARRAY new.user_ids LOOP
+    PERFORM id from users.user where id = user_id;
+    IF NOT FOUND THEN
+      RAISE EXCEPTION 'Invalid user id: %%', user_id;
+    END IF;
+  END LOOP;
+  -- check area ids
+  FOREACH area_id IN ARRAY new.area_ids LOOP
+    PERFORM document_id from guidebook.areas where document_id = area_id;
+    IF NOT FOUND THEN
+      RAISE EXCEPTION 'Invalid area id: %%', area_id;
+    END IF;
+  END LOOP;
+  RETURN null;
+END;
+$BODY$
+language plpgsql;
+
+CREATE TRIGGER guidebook_feed_document_changes_insert
+AFTER INSERT OR UPDATE ON guidebook.feed_document_changes
+FOR EACH ROW
+EXECUTE PROCEDURE guidebook.check_feed_ids();
+
+-- when deleting a user, check that there are no changes referencing the user
+CREATE OR REPLACE FUNCTION guidebook.check_feed_user_ids() RETURNS TRIGGER AS
+$BODY$
+BEGIN
+  -- check user ids
+  PERFORM change_id from guidebook.feed_document_changes
+    where user_ids @> ARRAY[OLD.id] limit 1;
+  IF FOUND THEN
+    RAISE EXCEPTION 'Row in feed_document_changes still references user id %%',
+      OLD.id;
+  END IF;
+  RETURN null;
+END;
+$BODY$
+language plpgsql;
+
+CREATE TRIGGER users_user_delete
+AFTER DELETE ON users.user
+FOR EACH ROW
+EXECUTE PROCEDURE guidebook.check_feed_user_ids();
+
+-- when deleting an area, check that there are no changes referencing the area
+CREATE OR REPLACE FUNCTION guidebook.check_feed_area_ids() RETURNS TRIGGER AS
+$BODY$
+BEGIN
+  -- check area ids
+  PERFORM change_id from guidebook.feed_document_changes
+    where area_ids @> ARRAY[OLD.document_id] limit 1;
+  IF FOUND THEN
+    RAISE EXCEPTION 'Row in feed_document_changes still references area id %%',
+      OLD.document_id;
+  END IF;
+  RETURN null;
+END;
+$BODY$
+language plpgsql;
+
+CREATE TRIGGER guidebook_areas_delete
+AFTER DELETE ON guidebook.areas
+FOR EACH ROW
+EXECUTE PROCEDURE guidebook.check_feed_area_ids();
+""")
+event.listen(DocumentChange.__table__, 'after_create', trigger_ddl)
--- a/c2corg_api/models/feed.py
+++ b/c2corg_api/models/feed.py
@@ -6,7 +6,8 @@ from c2corg_api.models.image import Image
 from c2corg_api.models.user import User
 from c2corg_api.models.utils import ArrayOfEnum
 from sqlalchemy.dialects.postgresql.base import ARRAY
-from sqlalchemy.orm import relationship
+from sqlalchemy.orm import relationship, column_property
+from sqlalchemy.sql.expression import select
 from sqlalchemy.sql.functions import func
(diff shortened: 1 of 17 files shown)


c2corg_api/views/feed.py
```diff
@@ -19,20 +22,59 @@ class FeedRest(object):
...
         self.request = request
 
-    @view(validators=[validate_preferred_lang_param])
+    @view(validators=[
+        validate_preferred_lang_param, validate_token_pagination])
     def get(self):
+        """Get the public feed.
+
+        Request:
+            `GET` `/search[?pl=...][&limit=...][&token=...]`
```
asaunier: ``/search`` => ``/feed`` I guess

c2corg_api/views/feed.py
```diff
@@ -49,23 +52,53 @@ def get(self):
...
+@resource(path='/personal-feed', cors_policy=cors_policy)
+class PersonalFeedRest(object):
+
+    def __init__(self, request):
+        self.request = request
+
+    @restricted_view(validators=[
+        validate_preferred_lang_param, validate_token_pagination])
+    def get(self):
+        """Get the personal homepage feed for the authenticated user.
```
asaunier: Do you mean that you can only see your own feed?
Isn't it possible to get the feed of some other user?

Edit: I see there's also a profile feed.

Just to make sure:
* ``FeedRest`` is the standard feed for anonymous or when a user has switchef off his filters
* ``PersonalFeedRest`` is the same feed but with the user's filters and followed changes.
Correct?

c2corg_api/models/feed.py
```diff
@@ -0,0 +1,144 @@
...
+from sqlalchemy.sql.schema import Column, ForeignKey, PrimaryKeyConstraint, \
+    CheckConstraint, Index
+from sqlalchemy.sql.sqltypes import Integer, String, DateTime, Boolean
+
+
+class FilterArea(Base):
+    """Feed filter on areas for a user.
+
+    E.g. with an entry (user_id=1, area_id=2) the user 1 will only
+    see changes in area 2 in their feed (except changes from followed users).
```
tsauerwein: See https://en.oxforddictionaries.com/usage/he-or-she-versus-they

Instead of writing "his/her" you can write "their".

c2corg_api/models/feed.py
```diff
@@ -142,3 +142,82 @@ class DocumentChange(Base):
...
+END;
+$BODY$
+language plpgsql;
+
+CREATE TRIGGER guidebook_feed_document_changes_insert
+AFTER INSERT OR UPDATE ON guidebook.feed_document_changes
+FOR EACH ROW
+EXECUTE PROCEDURE guidebook.check_feed_ids();
+
+-- when deleting a user, check that there are no changes referencing the user
```
tsauerwein: This trigger is run when a row from the user table is deleted. In case of an exception, the transaction is canceled (the user row is not deleted).

c2corg_api/models/feed.py
asaunier: Good to know, thanks :)

c2corg_api/models/user.py
```diff
@@ -103,6 +103,10 @@ class User(Base):
     feed_filter_activities = Column(
         ArrayOfEnum(enums.activity_type), nullable=False, server_default='{}')
 
+    # only show updates from followed users in the homepage feed
+    feed_followed_only = Column(
```
tsauerwein: No, if `feed_followed_only` is enabled, the area/activities filter are ignored. But you only see changes of your followed users, not the full feed.

To see the full feed, you should not set any area/activities filters.

c2corg_api/views/feed.py
tsauerwein: `FeedRest` is for unauthenticated users. `PersonalFeedRest` is for authenticated users (no matter if filter or not).

tsauerwein: In the UI you will have to use a different URL depending on if the user is authenticated or not.

c2corg_api/models/feed.py
tsauerwein: Currently there is no web-service to delete a user. I think that web-service should take of removing references in the "changes" table. We will see when we implement the service.

c2corg_api/models/user.py
asaunier: OK, I remember that we have discussed the possibility to only show followed contents in the feed.
I guess we have to add a flag into the user prefs UI and to support it in the API service (see https://github.com/c2corg/v6_ui/pull/676 and https://github.com/c2corg/v6_api/issues/402#issuecomment-252204075)

asaunier: No need to implement it now, but do you think it would be possible to temporarly switch off the filters (and subscriptions) without dropping all the activities and areas filters in the user prefs UI?

Some kind of button that we would show on top of the homepage feed, a bit like the bulb in v5:
sans titre

tsauerwein:
Thanks, for the review! I will update the common project and merge this PR. I will do the remaining points in a separate PR, so that you already can use the web-services in the UI.

c2corg_api/models/user.py
tsauerwein: Yes, that would be possible. For example by setting a parameters when requesting the feed (e.g. `/personal-feed?filter=0`).

c2corg_api/views/feed.py
tsauerwein: I wanted to have different URLs `/feed` and `/personal-feed` mainly for caching reasons. Because the public feed can be cached by Varnish (for a short period, eg 30 seconds).

asaunier: Excellent!

asaunier:
Yesterday I have updated the API demo, without upgrading the DB.

It seems not a big deal, most of the page in the UI works as expected. Except the outing detail view pages (outings lists are OK), because the API crashes for such requests with the following error in the logs:

> sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) column user.feed_filter_activities does not exist
[Sat Oct 08 17:42:57.065892 2016] [wsgi:error] [pid 20364] [remote 127.0.0.1:18120] LINE 1: ...t_modified, users."user".lang AS users_user_lang, users."use...
[Sat Oct 08 17:42:57.065906 2016] [wsgi:error] [pid 20364] [remote 127.0.0.1:18120]                                                              ^
[Sat Oct 08 17:42:57.065924 2016] [wsgi:error] [pid 20364] [remote 127.0.0.1:18120]  [SQL: 'SELECT guidebook.documents.version AS guidebook_documents_version, guidebook.documents.protected AS guidebook_documents_protected, guidebook.documents.type AS guidebook_documents_type, guidebook.user_profiles.document_id AS guidebook_user_profiles_document_id, guidebook.documents.document_id AS guidebook_documents_document_id, guidebook.user_profiles.activities AS guidebook_user_profiles_activities, guidebook.user_profiles.categories AS guidebook_user_profiles_categories, users."user".password AS users_user_password, users."user".id AS users_user_id, users."user".username AS users_user_username, users."user".name AS users_user_name, users."user".forum_username AS users_user_forum_username, users."user".email AS users_user_email, users."user".email_validated AS users_user_email_validated, users."user".email_to_validate AS users_user_email_to_validate, users."user".moderator AS users_user_moderator, users."user".validation_nonce AS users_user_validation_nonce, users."user".validation_nonce_expire AS users_user_validation_nonce_expire, users."user".last_modified AS users_user_last_modified, users."user".lang AS users_user_lang, users."user".feed_filter_activities AS users_user_feed_filter_activities, users."user".feed_followed_only AS users_user_feed_followed_only, documents_locales_1.id AS documents_locales_1_id, documents_locales_1.version AS documents_locales_1_version, documents_locales_1.title AS documents_locales_1_title, documents_locales_1.type AS documents_locales_1_type, documents_locales_1.lang AS documents_locales_1_lang, documents_1.version AS documents_1_version, documents_1.protected AS documents_1_protected, documents_1.type AS documents_1_type, areas_1.document_id AS areas_1_document_id, documents_1.document_id AS documents_1_document_id, areas_1.area_type AS areas_1_area_type \\nFROM guidebook.documents JOIN guidebook.user_profiles ON guidebook.documents.document_id = guidebook.user_profiles.document_id JOIN users."user" ON guidebook.user_profiles.document_id = users."user".id LEFT OUTER JOIN (guidebook.area_associations AS area_associations_1 JOIN (guidebook.documents AS documents_1 JOIN guidebook.areas AS areas_1 ON documents_1.document_id = areas_1.document_id) ON areas_1.document_id = area_associations_1.area_id) ON guidebook.documents.document_id = area_associations_1.document_id LEFT OUTER JOIN guidebook.documents_locales AS documents_locales_1 ON documents_1.document_id = documents_locales_1.document_id \\nWHERE guidebook.documents.redirects_to IS NULL AND users."user".email_validated AND guidebook.user_profiles.document_id IN (%(document_id_1)s, %(document_id_2)s) ORDER BY guidebook.user_profiles.document_id DESC'] [parameters: {'document_id_1': 8742, 'document_id_2': 8845}]

I guess that's because of the user data requested as outing attributes and because user attributes such as ``user.feed_filter_activities`` are computed using an additional query (which fails because the table is missing). But do we really need to do such a query here since we don't need the ``user.feed_filter_activities`` attribute in the outing?

tsauerwein:
> But do we really need to do such a query here since we don't need the user.feed_filter_activities attribute in the outing?

Good find! I opened an issue: https://github.com/c2corg/v6_api/issues/420

PR #415, opened 2016-10-06
Adds books to API

dkocich:
Requires to merge changes in v6_common https://github.com/c2corg/v6_common/pull/33

### DIFF
--- a/c2corg_api/models/book.py
+++ b/c2corg_api/models/book.py
@@ -1,0 +1,98 @@
+from c2corg_api.models import schema, Base, enums
+from c2corg_api.models.document import (
+    ArchiveDocument, Document, schema_document_locale, schema_attributes)
+from c2corg_api.models.schema_utils import get_update_schema,\
+    restrict_schema, get_create_schema
+from c2corg_api.models.utils import copy_attributes, ArrayOfEnum
+from sqlalchemy.dialects.postgresql.array import ARRAY
+from c2corg_common.fields_book import fields_book
+from colanderalchemy import SQLAlchemySchemaNode
+from sqlalchemy import (
+    Column,
+    Integer,
+    String,
+    SmallInteger,
+    ForeignKey
+    )
+from c2corg_common import document_types
+
+BOOK_TYPE = document_types.BOOK_TYPE
+
+
+class _BookMixin(object):
+    author = Column(String(100))
+    editor = Column(String(100))
+    activities = Column(ArrayOfEnum(enums.activity_type))
+    url = Column(String(255))
+    isbn = Column(String(17))
+    book_types = Column(ArrayOfEnum(enums.book_type))
+    nb_pages = Column(SmallInteger)
+    publication_date = Column(String(100))
+    langs = Column(ARRAY(String(2)))
+
+attributes = ['author', 'editor', 'activities', 'url', 'isbn',
+              'book_types', 'nb_pages', 'publication_date', 'langs']
+
+
+class Book(_BookMixin, Document):
+    __tablename__ = 'books'
+
+    document_id = Column(
+        Integer,
+        ForeignKey(schema + '.documents.document_id'), primary_key=True)
+
+    __mapper_args__ = {
+        'polymorphic_identity': BOOK_TYPE,
+        'inherit_condition': Document.document_id == document_id
+    }
+
+    def to_archive(self):
+        book = ArchiveBook()
+        super(Book, self)._to_archive(book)
+        copy_attributes(self, book, attributes)
+
+        return book
+
+    def update(self, other):
+        super(Book, self).update(other)
+        copy_attributes(other, self, attributes)
+
+
+class ArchiveBook(_BookMixin, ArchiveDocument):
+    __tablename__ = 'books_archives'
+
+    id = Column(
+        Integer,
+        ForeignKey(schema + '.documents_archives.id'), primary_key=True)
+
+    __mapper_args__ = {
+        'polymorphic_identity': BOOK_TYPE,
+        'inherit_condition': ArchiveDocument.id == id
+    }
+
+    __table_args__ = Base.__table_args__
+
+schema_book_locale = schema_document_locale
+schema_book_attributes = list(schema_attributes)
+schema_book_attributes.remove('geometry')
+
+schema_book = SQLAlchemySchemaNode(
+    Book,
+    # whitelisted attributes
+    includes=schema_book_attributes + attributes,
+    overrides={
+        'document_id': {
+            'missing': None
+        },
+        'version': {
+            'missing': None
+        },
+        'locales': {
+            'children': [schema_book_locale]
+        },
+    })
+
+schema_create_book = get_create_schema(schema_book)
+schema_update_book = get_update_schema(schema_book)
+schema_listing_book = restrict_schema(
+    schema_book, fields_book.get('listing'))
--- a/c2corg_api/scripts/migration/documents/books.py
+++ b/c2corg_api/scripts/migration/documents/books.py
@@ -17,13 +17,6 @@ class MigrateBooks(MigrateDocuments):
     def get_model_archive_document(self, locales):
         return ArchiveDocumentLocale if locales else ArchiveBook
 
-    def get_document_geometry(self, document_in, version):
-        return dict(
-            document_id=document_in.id,
-            id=document_in.id,
-            version=version
-        )
-
     def get_count_query(self):
         return (
             ' select count(*) '
--- a/c2corg_api/scripts/migration/documents/books.py
+++ b/c2corg_api/scripts/migration/documents/books.py
@@ -1,0 +1,114 @@
+from c2corg_api.models.book import Book, ArchiveBook, BOOK_TYPE
+from c2corg_api.models.document import DocumentLocale, ArchiveDocumentLocale, \
+    DOCUMENT_TYPE
+from c2corg_api.scripts.migration.documents.document import MigrateDocuments, \
+  DEFAULT_QUALITY
+from c2corg_api.scripts.migration.documents.routes import MigrateRoutes
+
+
+class MigrateBooks(MigrateDocuments):
+
+    def get_name(self):
+        return 'books'
+
+    def get_model_document(self, locales):
+        return DocumentLocale if locales else Book
+
+    def get_model_archive_document(self, locales):
+        return ArchiveDocumentLocale if locales else ArchiveBook
+
+    def get_document_geometry(self, document_in, version):
+        return dict(
+            document_id=document_in.id,
+            id=document_in.id,
+            version=version
+        )
+
+    def get_count_query(self):
+        return (
+            ' select count(*) '
+            ' from app_books_archives ba join books b on ba.id = b.id '
+            ' where '
+            ' b.redirects_to is null; '
+        )
+
+    def get_query(self):
+        return (
+            ' select '
+            '   ba.id, ba.document_archive_id, ba.is_latest_version, '
+            '   ba.is_protected, ba.redirects_to, '
+            '   ba.activities, ba.book_types, '
+            '   ba.author, ba.editor, ba.url, ba.isbn, ba.langs, '
+            '   ba.nb_pages, ba.publication_date '
+            ' from app_books_archives ba join books b on ba.id = b.id '
+            ' where b.redirects_to is null '
+            ' order by ba.id, ba.document_archive_id;'
+        )
+
+    def get_count_query_locales(self):
+        return (
+            ' select count(*) '
+            ' from app_books_i18n_archives ba '
+            '   join books b on ba.id = b.id '
+            ' where b.redirects_to is null;'
+        )
+
+    def get_query_locales(self):
+        return (
+            ' select '
+            '    ba.id, ba.document_i18n_archive_id, ba.is_latest_version, '
+            '    ba.culture, ba.name, ba.description '
+            ' from app_books_i18n_archives ba '
+            '   join books b on ba.id = b.id '
+            ' where b.redirects_to is null '
+            ' order by ba.id, ba.culture, ba.document_i18n_archive_id;'
+        )
+
+    def get_document(self, document_in, version):
+        if document_in.langs and 'ot' in document_in.langs:
+            document_in.langs.remove('ot')
+            document_in.langs.append('it')
+        return dict(
+            document_id=document_in.id,
+            type=BOOK_TYPE,
+            version=version,
+
+            activities=self.convert_types(
+                document_in.activities, MigrateRoutes.activities),
+            book_types=self.convert_types(
+                document_in.book_types, MigrateBooks.book_types),
+            author=document_in.author,
+            editor=document_in.editor,
+            url=document_in.url,
+            isbn=document_in.isbn,
+            nb_pages=document_in.nb_pages,
+            publication_date=document_in.publication_date,
+            langs=document_in.langs,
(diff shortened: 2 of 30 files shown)


c2corg_api/scripts/migration/documents/books.py
```diff
@@ -0,0 +1,110 @@
...
+            ' where '
+            ' b.redirects_to is null; '
+        )
+
+    def get_query(self):
+        return (
+            ' select '
+            '   ba.id, ba.document_archive_id, ba.is_latest_version, '
+            '   ba.is_protected, ba.redirects_to, '
+            '   ST_Force2D(ST_SetSRID(ba.geom, 3857)) geom, '
```
tsauerwein: Do books have a geometry?

c2corg_api/models/book.py
```diff
@@ -0,0 +1,101 @@
...
+    url = Column(String(255))
+    isbn = Column(String(17))
+    book_types = Column(ArrayOfEnum(enums.book_type))
+    nb_pages = Column(SmallInteger)
+
+    # it is varchar so I cant use - Column(DateTime)
+    publication_date = Column(String(100))
+
+    # (psycopg2.ProgrammingError) column books.lang does not exist
+    # lang = Column(String())
```
tsauerwein: `langs` should be of type `Column(ARRAY(String(2)))`. See also: http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html#sqlalchemy.dialects.postgresql.ARRAY

c2corg_api/models/book.py
```diff
@@ -0,0 +1,101 @@
...
+class _BookMixin(object):
+    author = Column(String(100))
+    editor = Column(String(100))
+    activities = Column(ArrayOfEnum(enums.activity_type))
+    url = Column(String(255))
+    isbn = Column(String(17))
+    book_types = Column(ArrayOfEnum(enums.book_type))
+    nb_pages = Column(SmallInteger)
+
+    # it is varchar so I cant use - Column(DateTime)
```
tsauerwein: This field can be something like ""1982-1998" so it is ok.

c2corg_api/scripts/migration/documents/books.py
```diff
@@ -0,0 +1,110 @@
...
+
+            activities=self.convert_types(
+                document_in.activities, MigrateRoutes.activities),
+            book_types=self.convert_types(
+                document_in.book_types, MigrateBooks.book_types),
+            author=document_in.author,
+            editor=document_in.editor,
+            url=document_in.url,
+            isbn=document_in.isbn,
+            # langs=document_in.langs,
```
tsauerwein: When migrating `langs` we have to decide if we only accept languages defined in [default_langs](https://github.com/c2corg/v6_common/blob/master/c2corg_common/attributes.py#L1) or any language. E.g. "ot".

To get all used languages, you can run:
```
select lang from (select unnest(langs) lang from app_books_archives)t group by lang;
```

c2corg_api/views/book.py
```diff
@@ -0,0 +1,73 @@
...
+class BookRest(DocumentRest):
+
+    @view(validators=[validate_pagination, validate_preferred_lang_param])
+    def collection_get(self):
+        return self._collection_get(BOOK_TYPE, book_documents_config)
+        # BOOK_TYPE - shouldn't be in the request because it causes error
+
+    @view(validators=[validate_id, validate_lang_param])
+    def get(self):
+        return self._get(Book, schema_book)
```
tsauerwein: You can set `include_areas=False` because books should not have a geometry.

c2corg_api/search/mappings/book_mapping.py
```diff
@@ -0,0 +1,45 @@
...
+    book_types = QEnumArray(
+        'btyp', model_field=Book.book_types)
+
+    isbn = QEnum(
+        'isbn', model_field=Book.isbn)
+    author = QEnum(
+        'auth', model_field=Book.author)
+    editor = QEnum(
+        'edit', model_field=Book.editor)
+    publication_date = QEnum(
```
tsauerwein: Should you be able to search by the publication date?

c2corg_api/search/mappings/book_mapping.py
```diff
@@ -0,0 +1,45 @@
...
+    activities = QEnumArray(
+        'act', model_field=Book.activities)
+    book_types = QEnumArray(
+        'btyp', model_field=Book.book_types)
+
+    isbn = QEnum(
+        'isbn', model_field=Book.isbn)
+    author = QEnum(
+        'auth', model_field=Book.author)
+    editor = QEnum(
```
tsauerwein: I am not sure about the type for `author` and `editor`. Should these two fields be searchable? In the full-text-search?

c2corg_api/views/document_associations.py
```diff
@@ -12,22 +13,23 @@
...
     route_documents_config, user_profile_documents_config, \
     image_documents_config, article_documents_config, area_documents_config, \
-    outing_documents_config
+    outing_documents_config, book_documents_config
 from c2corg_api.views.validation import updatable_associations
 from sqlalchemy.sql.expression import or_, and_
 
 associations_to_include = {
     WAYPOINT_TYPE: {
         'waypoints', 'waypoint_children', 'routes', 'images', 'articles'},
```
tsauerwein: Waypoints can also be associated to books.

c2corg_api/views/document_associations.py
```diff
@@ -12,22 +13,23 @@
...
     IMAGE_TYPE:
         {'waypoints', 'routes', 'images', 'users', 'outings', 'articles',
-         'areas'},
+         'areas', 'books'},
     ARTICLE_TYPE:
-        {'waypoints', 'routes', 'images', 'users', 'outings', 'articles'},
+        {'waypoints', 'routes', 'images', 'users', 'outings', 'articles',
+         'books'},
     AREA_TYPE: {'images'},
     USERPROFILE_TYPE: {'images'}
```
tsauerwein: If you want that associations are included when requesting the details for a book, you will have to add an entry `BOOK_TYPE: {...}`.

c2corg_api/views/document_associations.py
```diff
@@ -242,5 +247,27 @@ def get_linked_articles(document, lang):
         article_ids, lang, article_documents_config).get('documents')
 
 
+def get_linked_books(document, lang):
+    if document.type in [ARTICLE_TYPE,
```
tsauerwein: Do you need this check `if document.type in [...]`?

c2corg_api/models/book.py
```diff
@@ -0,0 +1,96 @@
...
+
+class _BookMixin(object):
+    author = Column(String(100))
+    editor = Column(String(100))
+    activities = Column(ArrayOfEnum(enums.activity_type))
+    url = Column(String(255))
+    isbn = Column(String(17))
+    book_types = Column(ArrayOfEnum(enums.book_type))
+    nb_pages = Column(SmallInteger)
+    publication_date = Column(String(100))
```
asaunier: > Do those langs represent that the book is printed (available to buy) in that languages

Yes, "langs" is the language(s) the book is written in.

c2corg_api/scripts/migration/documents/books.py
```diff
@@ -0,0 +1,108 @@
...
+    def get_name(self):
+        return 'books'
+
+    def get_model_document(self, locales):
+        return DocumentLocale if locales else Book
+
+    def get_model_archive_document(self, locales):
+        return ArchiveDocumentLocale if locales else ArchiveBook
+
+    def get_document_geometry(self, document_in, version):
```
tsauerwein: Is this function used?

c2corg_api/models/book.py
tsauerwein: Please remember to add `langs` (also in the migration script).

c2corg_api/tests/views/test_book.py
```diff
@@ -0,0 +1,571 @@
...
+                        {'document_id': self.article2.document_id}
+                    ],
+                    'images': [
+                        {'document_id': self.image2.document_id}
+                    ],
+                    'routes': [
+                        {'document_id': self.route2.document_id}
+                    ]
+                }
+                # ,
```
tsauerwein: Remove this block if it is not needed. Or did you explicitly want to test that no geometry can be added?

c2corg_api/tests/views/test_book.py
```diff
@@ -0,0 +1,571 @@
...
+
+        association_img_log = self.session.query(AssociationLog). \
+            filter(AssociationLog.parent_document_id ==
+                   book1.document_id). \
+            filter(AssociationLog.child_document_id ==
+                   self.image2.document_id). \
+            first()
+        self.assertIsNotNone(association_img_log)
+
+        # check that a link to the associated article is created
```
tsauerwein: article?

c2corg_api/scripts/migration/documents/books.py
dkocich: Yes and no... If I remove it then it crashes during migration -

  File "/home/dkocich/projects/c2corg/v6_api/c2corg_api/scripts/migration/migrate.py", line 93, in <module>
    main()
  File "/home/dkocich/projects/c2corg/v6_api/c2corg_api/scripts/migration/migrate.py", line 83, in main
    MigrateBooks(connection_source, session, batch_size).migrate()
  File "/home/dkocich/projects/c2corg/v6_api/c2corg_api/scripts/migration/documents/document.py", line 76, in migrate
    self._migrate(locales=False)
  File "/home/dkocich/projects/c2corg/v6_api/c2corg_api/scripts/migration/documents/document.py", line 146, in _migrate
    document_in, version)
  File "/home/dkocich/projects/c2corg/v6_api/c2corg_api/scripts/migration/documents/document.py", line 61, in get_document_geometry_archive
    doc['id'] = document_in.document_archive_id
TypeError: 'NoneType' object does not support item assignment

c2corg_api/tests/views/test_book.py
dkocich: Yes, I want to test geometry like I did also for the articles, which should not have geometry also and we tested it.

c2corg_api/scripts/migration/documents/books.py
tsauerwein: It looks like this is also a problem with the articles, see https://github.com/c2corg/v6_api/issues/456.

We should make sure that no empty geometries are created for books and articles during the migration.

dkocich:
Note: I should look here before merge ... https://github.com/c2corg/v6_api/blob/de6710114c289072a07bfe02064ee144400e5dda/c2corg_api/tests/views/test_book.py#L377-L390

c2corg_api/tests/views/test_book.py
```diff
@@ -0,0 +1,564 @@
...
+
+        # version with lang 'fr'
+        version_fr = self.get_latest_version('fr', versions)
+        archive_locale = version_fr.document_locales_archive
+        self.assertEqual(archive_locale.title, 'Escalades au Thaurac')
+
+        # check if geometry is not stored in database afterwards
+        self.assertIsNone(book1.geometry)
+
+        # IF I TRY TO USE 'image2' here, then only the LOG part don't work!
```
tsauerwein: Did you still plan to add a check for the association with the image?

c2corg_api/views/document_associations.py
```diff
@@ -219,11 +225,12 @@ def get_linked_articles(document, lang):
         Association.parent_document_id == Article.document_id
     )
 
-    if document.type == IMAGE_TYPE:
+    if document.type in [IMAGE_TYPE]:
```
tsauerwein: Why this change?

dkocich: mistake-changed back again

c2corg_api/tests/views/test_book.py
dkocich: added and deleted comment, it should be ok now.

dkocich:
I added that check for empty geom in articles / books and tested it. I did not have any geometry in new DB and in archives. Then I imported also huts and I had them only for huts, so it should be fine.

c2corg_api/views/validation.py
```diff
@@ -333,7 +331,10 @@ def _add_associations(
 
 
 def _is_parent_of_association(main_document_type, other_document_type):
-    if (main_document_type, other_document_type) in valid_associations:
+    # book is always main_d_t=parent?
+    if main_document_type == 'b' and other_document_type in valid_associations:
```
tsauerwein: Hmm, I am not sure why you would even need a special case for books. If the associations with books are listed in `valid_associations`, it should be detected properly.

But thinking about it: We should test that associations between books and books are handled correctly.

dkocich: Ok, my mistake.... I think that now we have no association between Book and Book...?https://github.com/c2corg/camptocamp.org/blob/master/data/sql/tables/db_associations.sql#L20-L71

c2corg_api/views/validation.py
```diff
@@ -317,9 +317,7 @@ def _add_associations(
                 'body', 'associations.' + document_key,
                 'invalid association type')
         else:
-            if main_document_type == 'b':  # book is always main=parent?
-                is_parent = False
-            elif document_key == 'waypoints':
+            if document_key == 'waypoints' and main_document_type != 'b':
```
tsauerwein: Please use `BOOK_TYPE` instead of `'b'`.

c2corg_api/views/validation.py
tsauerwein: Yes, you are right. I confused books with articles.

c2corg_api/views/validation.py
```diff
@@ -333,6 +331,7 @@ def _add_associations(
 
 
 def _is_parent_of_association(main_document_type, other_document_type):
+    # book is always main_d_t=parent?
```
dkocich: It does not. I fixed the last two comments.

dkocich:
Close https://github.com/c2corg/v6_api/issues/456

This issue should be ok and closed now for articles/books, but we still need to adjust code if it makes difference in empty geometry of some reports.

dkocich:
so the migration passed without error and geometry table in DB looks ok now

c2corg_api/scripts/migration/documents/books.py
```diff
@@ -0,0 +1,104 @@
...
+            editor=document_in.editor,
+            url=document_in.url,
+            isbn=document_in.isbn,
+            nb_pages=document_in.nb_pages,
+            publication_date=document_in.publication_date,
+            langs=document_in.langs
+        )
+
+    def get_document_locale(self, document_in, version):
+        description, summary = self.extract_summary(document_in.description)
```
tsauerwein: With https://github.com/c2corg/v6_api/pull/478 being merged, please do it like for the other documents, like here: https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/documents/articles.py#L87-L89

c2corg_api/views/book.py
```diff
@@ -42,14 +43,17 @@ def get(self):
 
     @restricted_json_view(
             schema=schema_create_book,
-            validators=[validate_book_create,
+            validators=[colander_body_validator,
+                        validate_id,
```
tsauerwein: I think this causes the failing unit tests. For `POST` you don't need `validate_id`.

tsauerwein:
Thanks a lot for your continuous work on this, good job! :+1: 

Please merge after Travis has passed.

Issue #419, opened 2016-10-06
Todo for area searching


asaunier:
- [x] support the `atyp` (area_type) filter in the search mapping https://github.com/c2corg/v6_api/blob/master/c2corg_api/search/mappings/area_mapping.py
- [x] ~~add areas to the simple search results~~ (already OK)
- [ ] return (simplified?) geometry in list requests? (see https://github.com/c2corg/v6_common/commit/3ff44c360bcbe87d366ecff1f4d11e47017de3ea)

tsauerwein:
This issue is now a duplicate of https://github.com/c2corg/v6_api/issues/489, closing.

PR #421, opened 2016-10-10
Add web-services for feed preferences and user following

tsauerwein:
Adds the following web-services:

- GET `/users/feed-preferences`: Get the preferences for the feed.
- POST `/users/feed-preferences`: Set the preferences for the feed.
- POST `/users/follow`: Follow a user given in the request body.
- POST `/users/unfollow`: Unfollow a user given in the request body.
- GET `/users/following/{user_id}`: Return if the authenticated user is following the given user.
- GET `/users/following`: Return a list with all users that the authenticated user is following.

Closes https://github.com/c2corg/v6_api/issues/401
Closes https://github.com/c2corg/v6_api/issues/402
Related to https://github.com/c2corg/v6_api/issues/403

### DIFF
--- a/c2corg_api/models/area_association.py
+++ b/c2corg_api/models/area_association.py
@@ -147,7 +147,8 @@ def get_areas(document, lang):
             AreaAssociation,
             Area.document_id == AreaAssociation.area_id). \
         options(load_only(
-            Area.document_id, Area.area_type, Area.version, Area.protected)). \
+            Area.document_id, Area.area_type, Area.version, Area.protected,
+            Area.type)). \
         options(joinedload(Area.locales).load_only(
             DocumentLocale.lang, DocumentLocale.title,
             DocumentLocale.version)). \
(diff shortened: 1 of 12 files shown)


asaunier:
Looks good to me.
Just wondering if ``/users/feed-preferences`` is really a relevant name? Perhaps it will be used for something else in the future (for instance to set default filters in the advanced search pages).
What about ``/users/filters-preferences`` or simply ``/users/preferences``?

asaunier:
Could be done in a separated PR, but for "areas" in the user prefs UI we need that the areas are actually indexed by ES and returned by the simple search / association tool for type "a". See https://github.com/c2corg/v6_api/issues/419 and https://github.com/c2corg/v6_ui/pull/676

tsauerwein:
> What about /users/filters-preferences or simply /users/preferences?

Yes, I renamed to `/users/preferences`.

Thanks for the review!

PR #422, opened 2016-10-10
Add ImageProxyRest service

arnaud-morvan:
Relate https://github.com/c2corg/v6_ui/issues/712

Add a GET /images/proxy/{id} view which return 302 if image is found or 404 if not found.


arnaud-morvan:
@tsauerwein : could you take a look, no cache for now, not sure this is necessary.

PR #424, opened 2016-10-12
Add size param to image proxy view

arnaud-morvan:
Relate https://github.com/c2corg/v6_ui/issues/712

### DIFF
--- a/c2corg_api/tests/views/test_image.py
+++ b/c2corg_api/tests/views/test_image.py
@@ -705,7 +705,20 @@ class TestImageProxyRest(BaseTestRest):
         self.app.get('/images/proxy/{}'.format(999),
                      status=404)
 
-    def test_success(self):
+    def test_bad_size(self):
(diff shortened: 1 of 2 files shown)


asaunier:
Are query string params like ``?size=BI`` in ``/images/proxy/5555?size=BI`` correctly taken into account by Varnish? Wouldn't it better to use `` /images/proxy/5555/BI``?

tsauerwein:
> Are query string params like ?size=BI in /images/proxy/5555?size=BI correctly taken into account by Varnish (ie. cached)? Wouldn't it better to use /images/proxy/5555/BI?

That does not make a difference for Varnish.

tsauerwein:
@arnaud-morvan Please fix the linter errors, and then feel free to merge.

asaunier:
Is there still something preventing this PR from merging?

arnaud-morvan:
If it's ok for the UI point of view, we can merge

Issue #425, opened 2016-10-13
User profiles access rights


asaunier:
It should be possible for user to choose whether their profile page is public or only accessible to authenticated users (or even completely private?? not available in v5).

See https://github.com/c2corg/v6_ui/issues/335 and https://github.com/c2corg/v6_ui/pull/728 for the UI part

If the visitor is not granted access to the current profile, perhaps we could return very minimal information, such as the guidebook name (and forum name?) since this attribute is anyway available in outings for instance.

tsauerwein:
TODO:

* [ ] Add field `is_profile_public` to `User`
* [ ]  Migrate field `is_profile_public`
* [ ] Include field in `/users/account` web-service
* [ ] Adapt web-service `/profiles/{user_id}` to check for `is_profile_public`

PR #427, opened 2016-10-13
Handle public/private profiles

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/425

### DIFF
--- a/c2corg_api/views/user_profile.py
+++ b/c2corg_api/views/user_profile.py
@@ -1,15 +1,17 @@
-
+from c2corg_api.models import DBSession
+from c2corg_api.models.user import User
 from c2corg_api.models.user_profile import schema_update_user_profile, \
     UserProfile, schema_user_profile, schema_internal_user_profile, \
     USERPROFILE_TYPE
 from c2corg_api.views.document_schemas import user_profile_documents_config
-from cornice.resource import resource
+from cornice.resource import resource, view
 
 from c2corg_api.views.document import DocumentRest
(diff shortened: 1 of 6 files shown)


c2corg_api/views/user_profile.py
```diff
@@ -22,9 +24,32 @@ def collection_get(self):
...
+            raise HTTPNotFound('user not found')
+        elif requested_user.is_profile_public or \
+                self.request.has_permission('authenticated'):
+            # only show the full profile if authenticated or if the user marked
+            # the profile as public
+            return self._get(UserProfile, schema_user_profile)
+        else:
+            # otherwise only return the user name
+            return {
+                'not_authenticated': True,
```
asaunier: What about returning ``not_authorized`` instead?

tsauerwein: To be nit-picky: Every user that is authenticated can see every profile, no special authorization is required. So, `not_authenticated` is more correct. ;)

asaunier: I think ``not_authorized`` is clearer: it's the main reason why you don't see the full profile. ``not authenticated`` is the partial reason why you're not authorized (the other part being that the profile is available to authenticated users only). By the way, in the future users might choose to make their profile not available to anyone => then every request would be ``not authorized`` (even for authenticated people).

tsauerwein:
Ok, I renamed to `not_authorized`. Thanks for the review!

asaunier:
Thanks for the change!

Issue #353, opened 2016-07-29
Requests are blocked with Gunicorn


tsauerwein:
I noticed that requests are sometimes blocked for 2-20 secs when using Gunicorn, before getting into the actual application logic. I did not observe this behavior with pserve.

@mfournier, any idea what could go wrong?

tsauerwein:
I increased the log-level by running Gunicorn with `.build/venv/bin/gunicorn --paste development.ini --reload --log-level debug`. For these blocking requests I see `Closing connection` before the request starts getting processed. For some reason the connections seem not to be closed properly.

```
[2016-07-29 10:47:50 +0200] [24310] [DEBUG] Closing connection. 
[2016-07-29 10:47:50 +0200] [24310] [DEBUG] Closing connection. 
[2016-07-29 10:47:50 +0200] [24310] [DEBUG] GET /waypoints
...

```

asaunier:
> I noticed that requests are sometimes blocked for 2-20 secs when using Gunicorn

I didn't know it was related to Gunicorn, but I have also observed very long response time for API requests!

asaunier:
Can we do something for this issue? My API instance often takes a while to answer and it's a pain in the neck for developing.

Do we have the same problem in prod mode?

tsauerwein:
> Can we do something for this issue? My API instance often takes a while to answer and it's a pain in the neck for developing.

You can switch back to paster:

    .build/venv/bin/pserve development.ini

> Do we have the same problem in prod mode?

I don't know.

mfournier:
Can you provide steps to reproduce this problem ? Or does it appear at random ?

Also, does it occur when the UI app is talking to the API ? Or when a standard browser does ?

tsauerwein:
I only noticed it when directly using the site in a browser. It often happens when I am on a listing page (e.g. http://localhost:6554/routes), browse to a detail page, and then go back to the listing page.

For listing pages the browser directly talks to the API. For the detail pages, it's the ui server side that makes a request to the API.

asaunier:
I noticed the problem when working on the UI advanced search page (UI on my PC, API on the dev server): I used to reload the UI page several times (because making changes in the UI code) and often had very long response time from the API (>45s). Reloading the UI (ie. triggering a new request to the API) generally helped having a faster response from the API.

asaunier:
Do we still have this problem?
[...]

PR #430, opened 2016-10-18
Use pserve instead Gunicorn for development server

tsauerwein:
Gunicorn is hanging too often, I now also had the problem with the UI dev server.

Closes https://github.com/c2corg/v6_api/issues/353


tsauerwein:
The build is failing because of https://github.com/eliben/pycparser/issues/151

asaunier:
OK for me. Do we have to do something to make the build succeed?

tsauerwein:
A fixed version of pycparser was released, the build passes again. I restarted all Travis builds.

PR #431, opened 2016-10-18
Update feed on changes

tsauerwein:
The feed is updated...

* When a new document is created (no user or image).
* When a new user is associated to an outing.
* When a group of images is uploaded.

Related to https://github.com/c2corg/v6_api/issues/403

### DIFF
--- a/c2corg_api/models/feed.py
+++ b/c2corg_api/models/feed.py
@@ -13,11 +13,12 @@ from c2corg_api.models.route import ROUTE_TYPE
 from c2corg_api.models.user import User
 from c2corg_api.models.user_profile import USERPROFILE_TYPE
 from c2corg_api.models.utils import ArrayOfEnum
+from c2corg_api.views.validation import association_keys
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.dialects.postgresql.base import ARRAY
 from sqlalchemy.orm import relationship, column_property
 from sqlalchemy.sql.elements import literal_column
-from sqlalchemy.sql.expression import select
+from sqlalchemy.sql.expression import select, text
 from sqlalchemy.sql.functions import func
 from sqlalchemy.sql.schema import Column, ForeignKey, PrimaryKeyConstraint, \
     CheckConstraint, Index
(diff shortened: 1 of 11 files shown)


c2corg_api/models/feed.py
```diff
@@ -367,6 +401,53 @@ def update_feed_association_update(
     update_participants_of_outing(child_document_id, user_id)
 
 
+def update_feed_images_upload(images, images_in, user_id):
```
asaunier: What's the difference between ``images`` and ``images_in``?

c2corg_api/views/feed.py
```diff
@@ -276,11 +276,18 @@ def load_feed(changes, lang):
...
 
-    # TODO deal with redirected documents
-
     documents_to_load = get_documents_to_load(changes)
     documents = load_documents(documents_to_load, lang)
 
+    # only return changes for which the document/user could be loaded
+    changes = [
+        c for c in changes
+        if documents.get(c.user_id) and documents.get(c.document_id)
```
asaunier: Not sure to understand how this is related to redirected documents?

asaunier:
Looks very good to me, thanks!

Just 2 comments/questions:
* is this ``filter=0`` parameter really needed to disable filters in the personal feed? Why not simply using the standard feed instead?
* is the ``quality`` attribute taken into account when adding documents to the feed? (it seems not :P)

tsauerwein:
> is this filter=0 parameter really needed to disable filters in the personal feed? Why not simply using the standard feed instead?

Indeed. If that works for you on the client-side, I will remove the parameter.

> is the quality attribute taken into account when adding documents to the feed? (it seems not :P)

It wasn't very clear what you wanted. We have to discuss this again.

c2corg_api/models/feed.py
tsauerwein: `images` are image entities, `images_in` is what was received in the request data. This naming schema `document/document_in` is also used elsewhere.

c2corg_api/views/feed.py
tsauerwein: Redirected documents are not included in the response to `load_documents` (one line above).

asaunier:
> Indeed. If that works for you on the client-side, I will remove the parameter.

I don't know. I have added a comment in the UI PR: https://github.com/c2corg/v6_ui/pull/731#issuecomment-254745091
@ginold what do you think?
Is it more convenient to add a parameter to the request or to switch the URL?

> It wasn't very clear what you wanted. We have to discuss this again.

OK I will create a dedicated issue to discuss this part (no big deal if not ready for the beta). It can probably be handled in a separate PR, can't it?

PR #433, opened 2016-10-19
Optimize indices for feed queries

tsauerwein:
The queries on the feed table are always made with an `order by` on `time`(desc) and `change_id`. I've set up a specific index that caters for this case.

See also: https://www.postgresql.org/docs/current/static/indexes-ordering.html

Closes https://github.com/c2corg/v6_api/issues/403

### DIFF
--- a/c2corg_api/models/feed.py
+++ b/c2corg_api/models/feed.py
@@ -113,7 +113,7 @@ class DocumentChange(Base):
 
     change_id = Column(Integer, primary_key=True)
 
-    time = Column(DateTime, default=func.now(), nullable=False, index=True)
+    time = Column(DateTime, default=func.now(), nullable=False)
 
     # the actor: who did the change?
     user_id = Column(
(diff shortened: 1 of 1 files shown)


c2corg_api/models/feed.py
```diff
@@ -161,16 +161,12 @@ class DocumentChange(Base):
     more_images = Column(Boolean, server_default='FALSE', nullable=False)
 
     __table_args__ = (
-        # TODO index for enum array? as text?
-        # Index(
-        #     'ix_guidebook_feed_document_changes_activities', activities,
-        #     postgresql_using='gin'),
+        # the queries on the feed table always order by time (desc) and
+        # change_id, therefor create an index for these two columns.
```
asaunier: Just curious if ``therefor`` without the trailing ``e`` is on purpose or not?
I have found this https://www.translegal.com/legal-english-lessons/therefore-vs-therefor

asaunier:
Looks good to me.
Do you have observed a significant gain in performance?

c2corg_api/models/feed.py
tsauerwein: Thanks! I didn't know the difference.

tsauerwein:
> Do you have observed a significant gain in performance?

Before the query took ~60ms, now with that index ~10ms.

Issue #429, opened 2016-10-18
Info service for users/profiles is missing


asaunier:
In the UI we will have user profile pages with URLs like http://www.camptocamp.org/users/377/fr

Omitting the lang parameter should be supported as well: http://www.camptocamp.org/users/377

For other document types we use the "info" service to figure out the available langs for a given document.
For instance http://c2corgv6-demo.gis.internal:6545/areas/282688/fr/info

But the service seems not available for "profiles":
http://c2corgv6-demo.gis.internal:6545/profiles/377/fr/info

PR #434, opened 2016-10-19
Add info service for user profiles

tsauerwein:
For example the request http://localhost:6544/profiles/377/fr/info returns:

```json
{
   "document_id":377,
   "locales":[
      {
         "lang":"fr",
         "title_prefix":null,
         "title":"alex"
      }
   ]
}
```

Closes https://github.com/c2corg/v6_api/issues/429

### DIFF
--- a/c2corg_api/tests/views/test_user_profile.py
+++ b/c2corg_api/tests/views/test_user_profile.py
@@ -118,6 +118,11 @@ class TestUserProfileRest(BaseDocumentTestRest):
     def test_get_404(self):
         self.get_404(user='contributor')
 
+    def test_get_info(self):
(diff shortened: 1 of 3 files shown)


asaunier:
:+1: 

Concerning the ``title``, I don't know if ``username`` is the good attribute (that's the login, right?). Perhaps ``name`` (guidebook name?) would make more sense.
Or maybe it's not really needed anyway, since I'm not sure we really want a slug for the profile documents (not really designed for being indexed by search engines).

tsauerwein:
> Concerning the title, I don't know if username is the good attribute (that's the login, right?). Perhaps name (guidebook name?) would make more sense.

Ok, I changed it to `name`.

Thanks for the review.

Issue #437, opened 2016-10-20
Feed KeyError 'c'


ginold:
on `/personal-feed?pl=es&token=50199%2C2016-03-21T17%3A11%3A54.627402` (after scrolling for a while) i get this error 

`KeyError: 'c'`

is it suppose to happen? something unfinished with the articles?

tsauerwein:
Do you have some details about the error message, e.g a bit more of the stacktrace?

ginold:
```
Traceback (most recent call last):
  File "/home/agina/v6_api/.build/venv/lib/python3.4/site-packages/pyramid_debugtoolbar/panels/performance.py", line 57, in resource_timer_handler
    result = handler(request)
  File "/home/agina/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/tweens.py", line 51, in excview_tween
    request_iface=request_iface.combined
  File "/home/agina/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/view.py", line 547, in _call_view
    response = view_callable(context, request)
  File "/home/agina/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/viewderivers.py", line 354, in _authdebug_view
    return view(context, request)
  File "/home/agina/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/viewderivers.py", line 413, in viewresult_to_response
    result = view(context, request)
  File "/home/agina/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/tweens.py", line 22, in excview_tween
    response = handler(request)
  File "/home/agina/v6_api/.build/venv/lib/python3.4/site-packages/pyramid_tm/__init__.py", line 101, in tm_tween
    reraise(*exc_info)
  File "/home/agina/v6_api/.build/venv/lib/python3.4/site-packages/pyramid_tm/compat.py", line 15, in reraise
    raise value
  File "/home/agina/v6_api/.build/venv/lib/python3.4/site-packages/pyramid_tm/__init__.py", line 83, in tm_tween
    response = handler(request)
  File "/home/agina/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/router.py", line 158, in handle_request
    view_name
  File "/home/agina/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/view.py", line 547, in _call_view
    response = view_callable(context, request)
  File "/home/agina/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/config/views.py", line 182, in __call__
[...]

PR #438, opened 2016-10-20
Add missing articles to the document_configs

dkocich:
It caused `KeyError: 'c'` on the feed page before.

More info in the issue thread that should be closed after this merge

Close https://github.com/c2corg/v6_api/issues/437


tsauerwein:
Great, thanks!

(actually it was me who forgot to add this entry... ;)

Issue #389, opened 2016-09-08
Associations are not returned for all document types


asaunier:
For instance in v6 http://c2corgv6.demo-camptocamp.com/images/733245/fr/contraste-sudnord has no associated doc whereas in v5 http://www.camptocamp.org/images/733245/fr/contraste-sud-nord has an associated outing.

@tsauerwein said

> I checked for this image, there is an association in the database:
 733237;"o";733245;"i"

But http://c2corgv6-api.demo-camptocamp.com/images/733245 returns no association at all.

Transfered from https://github.com/c2corg/v6_api/issues/412#issuecomment-254470582

> It seems also that the profile service does not provide the user's associations.
We probably don't need all associated documents (for instance outings) but it would be useful for associated images. For instance: http://www.camptocamp.org/users/377/fr#images

tsauerwein:
It looks like that outings are not included for images. Do you have an example that also other document types are missing?

asaunier:
None as far as I know, but double-checking could be useful.

tsauerwein:
Some documents might have a lot of associated images? So, we probably do not want to return all? We could do it like for outings and only include the latest x.

asaunier:
That looks consistent indeed.

PR #439, opened 2016-10-20
Add missing associations

tsauerwein:
- Associated outings for images and articles
- Associated images for user profiles

Closes https://github.com/c2corg/v6_api/issues/389


tsauerwein:
Thanks for the review!

Issue #399, opened 2016-09-20
Create images without title


arnaud-morvan:
@ginold says that API return error 500 when posting images without title.
He wants API to accept images without title.

I observe that in image upload dialog, title is filled with source file name, so it is necessary to empty it by hand.

IMHO this really seems tricky to accept images without title.
=> guidebook.documents_locales.title is mandatory
Not sure document wihout any locale will be correctly handled by the application.

asaunier:
That would be a pity that one cannot upload an image without a title.
In v5 people often use a dummy title for images (eg. "****" or "1234" or the outing title). It's better to have no title at all.
The uploaded filename is indeed not a suitable default title.

arnaud-morvan:
Ok for the no title concept, but no locale ?

asaunier:
No clue :)
Perhaps we could have a locale with an empty title ``''``?
@tsauerwein what do you think?

arnaud-morvan:
empty string is not NULL, should be accepted

arnaud-morvan:
I will try to reproduce and get better idea what is posted and what case the error 500

PR #440, opened 2016-10-20
Allow to create images without title

tsauerwein:
The `title` field can be omitted (e.g. `"locales":[{"lang":"fr"}]`) or set to `null` (e.g. `"locales":[{"lang":"fr", "title": null}]`).

Closes https://github.com/c2corg/v6_api/issues/399


tsauerwein:
Thanks for the review!

PR #442, opened 2016-10-20
Fix cornice branch

tsauerwein:
For Cornice we are currently still using an unmerged branch in my Cornice repository. I have a PR open with this branch. But today I rebased this branch with the latest version of Cornice, which unfortunately makes our tests fail.

So, I created a separate branch containing the fix on an old Cornice version.


asaunier:
What about using a branch in this repo instead in yours?

Any chance that we may use the latest version of Cornice one day?

tsauerwein:
> What about using a branch in this repo instead in yours?

We can until wait until the "go live". If we can't use an official release until then, we can create a fork in the orgs repo.

I am merging because the builds are failing currently.

Issue #436, opened 2016-10-19
Do not transfer the login for public requests


asaunier:
We have 3 kinds of user names:
* username (login?)
* name (aka "ful name")
* forum_username
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/user.py#L77-L79

As far as I know, the username is only used to authenticate, right?
Then it's probably a security issue to pass this attribute in non protected API responses.
Anyway we probably don't need it in the UI, ``name`` and ``forum_username`` should be enough (by the way it's confusing to have one more user name).

We should probably make sure it is never transmitted to the UI.

Related to https://github.com/c2corg/v6_ui/issues/555

Issue #420, opened 2016-10-10
Only query required fields of user for outing detail page


tsauerwein:
See https://github.com/c2corg/v6_api/pull/409#issuecomment-252431883

Use `loadonly` here: https://github.com/c2corg/v6_api/blob/6a388b8/c2corg_api/views/document_listings.py#L155

PR #447, opened 2016-10-21
Do not expose login user name

tsauerwein:
It looks like `username` is still used on the UI side, I will take a look. We can wait with merging this PR after the beta release.

Closes https://github.com/c2corg/v6_api/issues/436
Closes https://github.com/c2corg/v6_api/issues/420

### DIFF
--- a/c2corg_api/models/user.py
+++ b/c2corg_api/models/user.py
@@ -1,7 +1,6 @@
 import bcrypt
 from c2corg_api.models.utils import ArrayOfEnum
 from c2corg_common.attributes import default_langs
-from c2corg_api.models.schema_utils import restrict_schema
 from c2corg_api.models.user_profile import UserProfile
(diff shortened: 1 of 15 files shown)


asaunier:
OK thanks!

I have created a UI issue: https://github.com/c2corg/v6_ui/issues/765

We can indeed figure this out after the beta.

By the way, in the user profile it would make sense to shown the ``forum_username`` as well. Perhaps something that could be returned by the profile data service?

tsauerwein:
> By the way, in the user profile it would make sense to shown the forum_username as well. Perhaps something that could be returned by the profile data service?

+1

asaunier:
What do you recommend then?

tsauerwein:
The latest commit includes the forum username in the profile output.

asaunier:
Good!

PR #448, opened 2016-10-23
Remove duration and duration_difficulties fields

asaunier:
Requires https://github.com/c2corg/v6_common/pull/35


tsauerwein:
Ok for me.

Issue #451, opened 2016-10-25
Route attribute "external_resources" is not imported from v5


asaunier:
* v5: https://github.com/c2corg/camptocamp.org/blob/master/data/sql/tables/db_routes.sql#L59
* v6: https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/route.py#L232

but missing in the migration script: https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/documents/routes.py

PR #452, opened 2016-10-25
Add attribute external_resources to the route migration script

asaunier:
Fixes https://github.com/c2corg/v6_api/issues/451

Running the migration script with this change has not been tested yet. Same change than for the ``route_history`` attribute.


asaunier:
Thanks for reviewing :)

PR #461, opened 2016-10-28
Fix some inconstencies in config

arnaud-morvan:
This could have some side effect on you local configs.
But it should be really simple to fix (see my config).

### DIFF
--- a/config/default
+++ b/config/default
@@ -72,4 +72,4 @@ export discourse_api_key = 4647c0d98e8beb793da099ff103b9793d8d4f94fff7cdd52d5839
 export image_backend_secret_key = test
 
 export skip_captcha_validation = False
-export recaptcha_secret_key = 6LdWkR4TAAAAALcZLh54HFlAWFHiMQhZR5jR4p3F
+export recaptcha_secret_key = 6LfWUwoUAAAAABDo4_HRfru4HfLxOKmcqBRBObGj
--- a/config/default
+++ b/config/default
@@ -1,11 +1,9 @@
-export instanceid = main
-export base_url = /
+export instanceid ?= main
+export base_url ?= /
+export debug_port ?= 6543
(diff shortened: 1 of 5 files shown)


config/main
```diff
@@ -1,8 +1,9 @@
+instanceid = api
+base_url = /
+ui_url = http://c2corgv6-discourse.demo-camptocamp.com/
```
asaunier: really discourse?

arnaud-morvan: good catch, bad copy/paste, will fix it

config/default
```diff
@@ -1,11 +1,9 @@
-export instanceid = main
-export base_url = /
+export instanceid ?= main
+export base_url ?= /
+export ui_url ?= "TO BE DEFINED"
```
asaunier: What about using "http://c2corgv6.demo-camptocamp.com" here?
It would be consistent with 
```
export image_backend_url = http://c2corgv6-images.demo-camptocamp.com
export image_url = http://c2corgv6-images.demo-camptocamp.com/active/

export discourse_url = http://c2corgv6-discourse.demo-camptocamp.com
```
a few lines below.

config/default
```diff
@@ -1,11 +1,9 @@
...
+export debug_port ?= 6543
```
asaunier: Just curious: what's the difference between
```
export debug_port ?= 6543
```
and
```
export debug_port = 6543
```
?

arnaud-morvan: The difference is that ?= do not override previsouly set value

config/main
```diff
@@ -1,8 +1,9 @@
+instanceid = api
+base_url = /
+ui_url = http://c2corgv6.demo-camptocamp.com/
```
asaunier: Do we need this if this is already the value in config/default?

asaunier:
Looks good to me!

PR #467, opened 2016-11-03
Mailing list service

### LINKED ISSUE #426: Mailing list service
- [x] service returning the list of mailing lists + the registration status for the passed user id
- [x] service to register/unregister to a mailing list
- [x] update the sympa table when a user updates their email address => trigger
- [x] migration from v5? (dump/restore of the v5 "sympa" table?)
 * [x] remove emails that are no available in the user account table (syncro errors in v5?)
- [ ] ~~install the tools existing in v5 to get the snow reports from SLF or MF and send them via the Sympa system~~ => done by the association

asaunier:
Fixes #426

### DIFF
--- a/c2corg_api/models/mailinglist.py
+++ b/c2corg_api/models/mailinglist.py
@@ -1,0 +1,35 @@
+from sqlalchemy import Column, DateTime, ForeignKey, Integer, String
+from sqlalchemy.orm import relationship
+from sqlalchemy.sql.functions import func
+
+from c2corg_api.models import Base, sympa_schema, users_schema
+from c2corg_api.models.user import User
+
+
+class Mailinglist(Base):
+    """
+    Class containing mailing list subscriptions.
+    Based upon Sympa mailing list server https://www.sympa.org/
+    """
+    __tablename__ = 'subscriber_table'
+    __table_args__ = {"schema": sympa_schema}
+
+    listname = Column('list_subscriber', String(50), primary_key=True)
+    email = Column('user_subscriber', String(200), primary_key=True)
+    user_id = Column(
+        Integer, ForeignKey(users_schema + '.user.id'), nullable=False)
+    user = relationship(User, primaryjoin=user_id == User.id)
+
+    # additional fields, used only by Sympa
+    date_subscriber = Column(
+        DateTime, default=func.now(), server_default=func.now(),
+        nullable=False)
+    update_subscriber = Column(DateTime)
+    visibility_subscriber = Column(String(20))
+    reception_subscriber = Column(String(20))
+    bounce_subscriber = Column(String(35))
+    bounce_score_subscriber = Column(Integer)
+    comment_subscriber = Column(String(150))
+    subscribed_subscriber = Column(Integer)
+    included_subscriber = Column(Integer)
+    include_sources_subscriber = Column(String(50))
--- a/c2corg_api/models/user.py
+++ b/c2corg_api/models/user.py
@@ -13,12 +13,13 @@ from sqlalchemy import (
 
 from colanderalchemy import SQLAlchemySchemaNode
 
-from c2corg_api.models import Base, users_schema, schema, enums
+from c2corg_api.models import Base, users_schema, schema, sympa_schema, enums
 
 import colander
 from sqlalchemy.orm import relationship, backref
 from sqlalchemy.sql.functions import func
 from sqlalchemy.sql.schema import ForeignKey
+from sqlalchemy import event, DDL
 
 from enum import Enum
 
@@ -172,3 +173,24 @@ schema_create_user = SQLAlchemySchemaNode(
             'validator': colander.OneOf(default_langs)
         }
     })
+
+# Make sure that user email changes are propagated to mailing lists as well
+trigger_ddl = DDL("""
+CREATE OR REPLACE FUNCTION users.update_mailinglists_email() RETURNS TRIGGER AS
+$BODY$
+BEGIN
+  UPDATE """ + sympa_schema + """.subscriber_table
+  SET user_subscriber = NEW.email
+  WHERE user_subscriber = OLD.email;
+  RETURN null;
+END;
+$BODY$
+language plpgsql;
+
+CREATE TRIGGER users_email_update
+AFTER UPDATE ON users.user
+FOR EACH ROW
+WHEN (OLD.email IS DISTINCT FROM NEW.email)
+EXECUTE PROCEDURE users.update_mailinglists_email();
+""")
+event.listen(User.__table__, 'after_create', trigger_ddl)
--- a/c2corg_api/tests/views/test_user_mailinglists.py
+++ b/c2corg_api/tests/views/test_user_mailinglists.py
@@ -1,0 +1,81 @@
+from c2corg_api.models.user import User
+from c2corg_api.models.mailinglist import Mailinglist
+from c2corg_api.tests.views import BaseTestRest
+from c2corg_common.attributes import mailinglists
+
+
+class TestUserMailinglistsRest(BaseTestRest):
+
+    def setUp(self):  # noqa
+        super(TestUserMailinglistsRest, self).setUp()
+        self._prefix = '/users/mailinglists'
+
+        self.contributor = self.session.query(User).get(
+            self.global_userids['contributor'])
+        ml1 = Mailinglist(
+            listname='meteofrance-74',
+            email=self.contributor.email,
+            user_id=self.contributor.id,
+            user=self.contributor
+        )
+        ml2 = Mailinglist(
+            listname='avalanche.en',
+            email=self.contributor.email,
+            user_id=self.contributor.id,
+            user=self.contributor
+        )
+        self.session.add_all([ml1, ml2])
+        self.session.flush()
+
+    def test_get_mailinglists_unauthenticated(self):
+        self.app.get(self._prefix, status=403)
+
+    def test_get_mailinglists(self):
+        headers = self.add_authorization_header(username='contributor')
+        response = self.app.get(self._prefix, status=200, headers=headers)
+        body = response.json
+
+        self.assertEqual(len(body), len(mailinglists))
+        for ml in mailinglists:
+            self.assertIn(ml, body)
+            if ml in ['meteofrance-74', 'avalanche.en']:
+                self.assertTrue(body[ml])
+            else:
+                self.assertFalse(body[ml])
+
+    def test_post_mailinglists_unauthenticated(self):
+        self.app_post_json(self._prefix, {}, status=403)
+
+    def test_post_mailinglists_invalid(self):
+        request_body = {
+            'wrong_mailinglist_name': True,
+            'avalanche': 'incorrect_value'
+        }
+
+        headers = self.add_authorization_header(username='contributor')
+        response = self.app_post_json(
+            self._prefix, request_body, status=400, headers=headers)
+
+        body = response.json
+        self.assertEqual(body.get('status'), 'error')
+        errors = body.get('errors')
+        self.assertIsNotNone(self.get_error(errors, 'wrong_mailinglist_name'))
+        self.assertIsNotNone(self.get_error(errors, 'avalanche'))
+
+    def test_post_mailinglists(self):
+        request_body = {
+            'meteofrance-66': True,
+            'meteofrance-74': False
+        }
+
+        headers = self.add_authorization_header(username='contributor')
+        self.app_post_json(
+            self._prefix, request_body, status=200, headers=headers)
+
+        mls = self.session.query(Mailinglist.listname).filter(
+            Mailinglist.email == self.contributor.email).all()
+        subscribed_mailinglists = [l[0] for l in mls]
+        self.assertEqual(len(subscribed_mailinglists), 2)
+        self.assertIn('meteofrance-66', subscribed_mailinglists)
+        self.assertIn('avalanche.en', subscribed_mailinglists)
+        self.assertNotIn('meteofrance-74', subscribed_mailinglists)
--- a/c2corg_api/views/user_mailinglists.py
+++ b/c2corg_api/views/user_mailinglists.py
@@ -1,0 +1,91 @@
+from c2corg_api import DBSession
+from c2corg_api.models.user import User
+from c2corg_api.models.mailinglist import Mailinglist
+from c2corg_api.views import cors_policy, restricted_json_view, \
+    restricted_view
+from c2corg_common.attributes import mailinglists as valid_mailinglists
+from cornice.resource import resource
+
+
+def validate_mailinglist_statuses(request, **kwargs):
+    mailinglists = {}
+    data = request.json_body
+    for ml in data:
+        if ml not in valid_mailinglists:
+            request.errors.add(
+                'body', ml, 'Mailing list `{0}` does not exist'.format(ml))
+        elif type(data[ml]) != bool:
+            request.errors.add(
+                'body', ml,
+                'Status `{0}` of mailing list `{1}` should be boolean'.format(
+                    data[ml], ml))
+        else:
+            mailinglists[ml] = data[ml]
+    request.validated['mailinglists'] = mailinglists
+
+
+@resource(path='/users/mailinglists', cors_policy=cors_policy)
+class UserMailinglistsRest(object):
+
+    def __init__(self, request):
+        self.request = request
+
+    @restricted_view()
+    def get(self):
+        """Get the mailinglists subscriptions of the authenticated user.
+
+        Request:
+            `GET` `/users/mailinglists`
+
+        Example response:
+
+            {'avalanche': False, 'lawinen': True, 'valanghe': False, ...}
+
+        """
+        user_id = self.request.authenticated_userid
+        res = DBSession.query(Mailinglist.listname). \
+            filter(Mailinglist.user_id == user_id).all()
+        subscribed_mailinglists = [l[0] for l in res]
+
+        return {ml: ml in subscribed_mailinglists for ml in valid_mailinglists}
+
+    @restricted_json_view(validators=[validate_mailinglist_statuses])
+    def post(self):
+        """Update mailinglist subscription statuses.
+
+        Request:
+            `POST` `/users/mailinglists`
+            {'avalanche': False, 'lawinen': True, 'valanghe': False}
+
+        """
+        user_id = self.request.authenticated_userid
+        user = DBSession.query(User).get(user_id)
+
+        subscribed_lists = DBSession.query(Mailinglist). \
+            filter(Mailinglist.user_id == user_id).all()
+        subscribed_lists = {l.listname: l for l in subscribed_lists}
+        subscribed_listnames = set(subscribed_lists.keys())
+
+        lists_to_add = []
+        removed_lists = False
+        data = self.request.validated['mailinglists']
+        for listname in data:
+            subscribe_status = data.get(listname, False)
+            if subscribe_status and listname not in subscribed_listnames:
+                # Add list
+                lists_to_add.append(Mailinglist(
+                    listname=listname,
+                    email=user.email,
+                    user_id=user_id,
+                    user=user
+                ))
+            elif not subscribe_status and listname in subscribed_listnames:
+                # Remove list
+                removed_lists = True
+                DBSession.delete(subscribed_lists[listname])
+
+        if lists_to_add:
+            DBSession.add_all(lists_to_add)
+        if lists_to_add or removed_lists:
+            DBSession.flush()
+        return {}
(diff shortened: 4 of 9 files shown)


c2corg_api/models/mailinglist.py
```diff
@@ -0,0 +1,22 @@
...
+    Class containing mailing list subscriptions.
+    Based upon Sympa mailing list server https://www.sympa.org/
+    """
+    __tablename__ = 'subscriber_table'
+    __table_args__ = {"schema": sympa_schema}
+
+    listname = Column('list_subscriber', String(50), nullable=False)
+    email = Column(
+        'user_subscriber', String(200),
+        ForeignKey(users_schema + '.user.email'), nullable=False)
```
tsauerwein: Making the email a foreign key might be problematic when the email is updated in the users table. Maybe it would make sense to have to columns `user_id` (foreign key on the users table) and `email`.

The `email` column could be updated with a trigger when the column `email` in the users table changes.

c2corg_api/views/user_mailinglists.py
```diff
@@ -0,0 +1,48 @@
...
+        """
+        user_id = self.request.authenticated_userid
+        listnames = DBSession.query(Mailinglist.listname). \
+            options(joinedload(Mailinglist.user).load_only(User.email)). \
+            filter(User.id == user_id). \
+            all()
+
+        return {ml: ml in mailinglists for ml in listnames}
+
+    @restricted_json_view(schema=MailinglistSchema())
```
tsauerwein: With the update of Cornice you will have to specify the schema validator explicitly:

```
from cornice.validators import colander_body_validator
...
    @restricted_json_view(schema=MailinglistSchema()), validators=[colander_body_validator])

```

c2corg_api/models/mailinglist.py
asaunier: Right, we must propagate any email change in the user account table to the "mailinglist" table. Using a trigger seems a good idea. I have added it to the todo list at https://github.com/c2corg/v6_api/issues/426

By the way ``list_subscriber`` and ``user_subscriber`` are the fields expected by Sympa but I have set aliases (``listname`` and ``email``) to make them more explicit.

Let me recap:
* I should use a ``user_id`` attribute to this ``subscriber_table``, make it a foreign key on the users table
* I keep the ``email`` attribute but it is no longer the foreign key
* I had a trigger to update this ``email`` attribute when the column ``email`` in the users table changes.

Is my ``user`` relationship correct? Should I set it explicitely when creating a ``Mailinglist`` object? As in https://github.com/c2corg/v6_api/blob/dc24a7e486a72e3b99d099125e47f96212cc026f/c2corg_api/tests/views/test_user_mailinglists.py#L18

c2corg_api/views/user_mailinglists.py
```diff
@@ -0,0 +1,48 @@
...
+    def post(self):
+        """Update a mailinglist subscription status.
+
+        Request:
+            `POST` `/users/mailinglists`
+            {'avalanche': False, 'lawinen': True, 'valanghe': False}
```
asaunier: @tsauerwein Does this make sense to you?
I plan to pass changed lists into the POST requests with a boolean value to tell if the list must be enabled or disabled for the current user.
For instance in the UI, if you check the "lawinen" list, an API request with ``{'lawinen': True}`` is triggered.

Or should I encapsulate those data in order to define a validator?

c2corg_api/models/mailinglist.py
tsauerwein: > Is my user relationship correct? Should I set it explicitely when creating a Mailinglist object?

Yes, that's ok.

c2corg_api/views/user_mailinglists.py
tsauerwein: > I plan to pass changed lists into the POST requests with a boolean value to tell if the list must be enabled or disabled for the current user.

What ever works best for you in the UI. You are planning to do a request every time you toggle the subscription for a list? And not send one request containing the status for all lists?

Maybe you don't even need an explicit schema. Simply a validator that checks that the list names are one of `c2corg_common.attributes.mailinglists` could be enough.

asaunier: > You are planning to do a request every time you toggle the subscription for a list?

Yes, because:
* this way no need to press a "save" button in the UI, which is what modern UI do now (including our preferences page)
* most people only subscribe to a few lists. In Switzerland, most likely only one (or 2 if you live near Haute Savoie). If you leave in Chambery, maybe 3 (mf-74, mf-73, mf-38).
* people don't need to change all there subscriptions at a time, maybe 1 or 2 from times to times, for instance if you plan a trip in a remote place and want to follow the avalanche situation a week before leaving (personally I have never changed my 2-list subscription I think :P)

> Maybe you don't even need an explicit schema. Simply a validator that checks that the list names are one of c2corg_common.attributes.mailinglists could be enough.

OK. I admit I am not really sure what schema are used for. I guess something simple is enough.
Then I will skip the schema and use only a validator if you think a schema is not mandatory here.

asaunier:
I have updated/rebased my PR. It's almost complete (the migration from v5 is still missing) but I am not very sure the validator and the POST request handler are correct. @tsauerwein what do you think?

As for now the tests fails because of
```
ERROR: test_get_mailinglists (c2corg_api.tests.views.test_user_mailinglists.TestUserMailinglistsRest)
  File "/home/alex/v6_api/c2corg_api/tests/views/test_user_mailinglists.py", line 35, in test_get_mailinglists
    response = self.app.get(self._prefix, status=200, headers=headers)
...
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/simplejson/encoder.py", line 373, in iterencode
    return _iterencode(o, 0)
nose.proxy.TypeError: keys must be a string


FAIL: test_post_mailinglists (c2corg_api.tests.views.test_user_mailinglists.TestUserMailinglistsRest)
  File "/home/alex/v6_api/c2corg_api/tests/views/test_user_mailinglists.py", line 84, in test_post_mailinglists
    self.assertIn('meteofrance-66', mls)
nose.proxy.AssertionError: 'meteofrance-66' not found in [('meteofrance-74',), ('avalanche.en',)]


FAIL: test_post_mailinglists_invalid (c2corg_api.tests.views.test_user_mailinglists.TestUserMailinglistsRest)
  File "/home/alex/v6_api/c2corg_api/tests/views/test_user_mailinglists.py", line 58, in test_post_mailinglists_invalid
    self._prefix, request_body, status=400, headers=headers)
  File "/home/alex/v6_api/c2corg_api/tests/__init__.py", line 208, in app_post_json
    return self.app_send_json('post', *args, **kwargs)
  File "/home/alex/v6_api/c2corg_api/tests/__init__.py", line 227, in app_send_json
    errors))
nose.proxy.AssertionError: Bad response: 200 OK (not 400)
```

=> it seems the POST requests are not taken into account correctly.

c2corg_api/models/user.py
```diff
@@ -172,3 +173,23 @@ def validate_password(self, plain_password):
...
+
+# Make sure that user email changes are propagated to mailing lists as well
+trigger_ddl = DDL("""
+CREATE OR REPLACE FUNCTION users.update_mailinglists_email() RETURNS TRIGGER AS
+$BODY$
+BEGIN
+  UPDATE """ + sympa_schema + """.subscriber_table
+  SET user_subscriber = NEW.email
+  WHERE user_subscriber = OLD.email;
+END;
```
tsauerwein: In the logs is an error related to the trigger:

```
sqlalchemy.exc.InternalError: (psycopg2.InternalError) control reached end of trigger procedure without RETURN

CONTEXT:  PL/pgSQL function users.update_mailinglists_email()
```

c2corg_api/views/user_mailinglists.py
```diff
@@ -0,0 +1,82 @@
...
+        Example response:
+
+            {'avalanche': False, 'lawinen': True, 'valanghe': False, ...}
+
+        """
+        user_id = self.request.authenticated_userid
+        listnames = DBSession.query(Mailinglist.listname). \
+            filter(Mailinglist.user_id == user_id).all()
+
+        return {ml: ml in mailinglists for ml in listnames}
```
tsauerwein: The error message is: `nose.proxy.TypeError: keys must be a string`

Here you are constructing a dictionary with a tuple as key (`listnames` is a list of tuples).

c2corg_api/views/user_mailinglists.py
```diff
@@ -0,0 +1,82 @@
...
+from c2corg_api.models.mailinglist import Mailinglist
+from c2corg_api.views import cors_policy, restricted_json_view, \
+    restricted_view
+from c2corg_common.attributes import mailinglists
+from cornice.resource import resource
+
+
+def validate_mailinglist_statuses(request, **kwargs):
+    tested = []
+    for ml in request.validated:
```
tsauerwein: Is the data in `validated`? If you are not using a schema or another validator that has written data to `validated`, you must get your data from `request.GET` or `request.POST`.

asaunier: > If you are not using a schema or another validator that has written data to validated

I use no schema and no other validator. I have read that the standard ``colander_body_validator`` does nothing if no schema is provided so I have not added it to the list of validators of the post handler. Perhaps I should, in order to have ``request.validated`` in my validator? Or simply use ``request.POST`` (but that seems not very robust if we add another validator later?)?

c2corg_api/views/user_mailinglists.py
asaunier: OK thanks for pointing the problem. I was unsure where this ``nose.proxy.TypeError: keys must be a string`` error came from.

c2corg_api/models/user.py
asaunier: OK, so I should simply add a ``RETURN`` statement to the trigger.

c2corg_api/views/user_mailinglists.py
tsauerwein: Simply work with `request.POST` if you do not want to use a schema.

asaunier:
Thanks @tsauerwein for your help!
I will make changes to fix the problems.

I would also like to add the migration part (or in a dedicated PR?). That should not be too complicated since the source and target tables/fields are similar. I see 2 difficulties though:
* entries with email address no longer matching a user account should be ignored
* the target table requires to set the ``user_id`` (and also the ``user`` relationship?).

Does this mean that for each migrated entry I have to make a ``DBSession.query(User).filter(User.email == <email>).first()`` query?

tsauerwein:
> I see 2 difficulties though:

When making the query on the v5 database to get the data, you can do a join on the old users table to only get valid entries and and also the user id.

asaunier:
@tsauerwein I need your help for the mailing lists migration script.

I have created a first version available at 02864bb using
* https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/map_associations.py
* https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/documents/associations.py
as examples.

2 questions:

1/ Is the vaccum analyze useful for mailing lists (it's about 5000 rows)?

2/ When inserting the subscriptions in the target table, I need to set the ``user_id`` (and ``user``?) as well. See https://github.com/c2corg/v6_api/blob/02864bb4671ec50901c34f196159416703422539/c2corg_api/models/mailinglist.py#L18-L20
What's the best approach to do so?

Should I get the user id for every single row (making a select query to ``users.user`` every time) in https://github.com/c2corg/v6_api/blob/02864bb4671ec50901c34f196159416703422539/c2corg_api/scripts/migration/mailinglists.py#L47-L52 ?

Or can I run an addition update request to set all the ``user_id`` at once with a request like https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/documents/associations.py#L188-L217 ?
By the way I guess my initial requests will fail if not setting ``user_id`` since this field is not nullable. What about the ``user`` relationship? Do I need to do something special when making the SQL request or is it implicitely built when setting ``user_id``?

Thanks!

tsauerwein:
> 2/ When inserting the subscriptions in the target table, I need to set the user_id (and user?) as well. 
> What's the best approach to do so?

I am not sure I understand your question. Currently your query is:
```
select list_subscriber, user_subscriber from subscriber_table ml
join app_users_private_data u on ml.user_subscriber = u.email;
```
Why can't you do:
```
select list_subscriber, user_subscriber, u.id as user_id from subscriber_table ml
join app_users_private_data u on ml.user_subscriber = u.email;
```

tsauerwein:
> 1/ Is the vaccum analyze useful for mailing lists (it's about 5000 rows)?

Actually `analyze` would be enough (it should also be only `analyze` for all other tables). `vaccum` makes only sense when rows were deleted. But updating the statistics for the query planner with `analyze` is reasonable for all tables.

asaunier:
> Why can't you do:

You're right! My question was stupid :P 

> Actually analyze would

OK thanks!

asaunier:
This PR is almost ready!

The addition to the migration script works (I recently did a full migration including the mailing lists).
Updating one's subscriptions in the UI works as expected (in combination with https://github.com/c2corg/v6_ui/pull/904).

The only problem remaining is that the mailinglist validator https://github.com/c2corg/v6_api/blob/0d74b1bd5c418ed7c61e9c7b417ec0b971fd8652/c2corg_api/views/user_mailinglists.py#L10-L26 seems to never be called. As a result the "invalid request" test fails: https://github.com/c2corg/v6_api/blob/0d74b1bd5c418ed7c61e9c7b417ec0b971fd8652/c2corg_api/tests/views/test_user_mailinglists.py#L49-L58

Is there something else I should do then add the validator at https://github.com/c2corg/v6_api/blob/0d74b1bd5c418ed7c61e9c7b417ec0b971fd8652/c2corg_api/views/user_mailinglists.py#L54 ?

In addition to read the request data, I had to use ``self.request.json_body`` https://github.com/c2corg/v6_api/blob/0d74b1bd5c418ed7c61e9c7b417ec0b971fd8652/c2corg_api/views/user_mailinglists.py#L73 Correct?

@tsauerwein perhaps one last tip? :)

c2corg_api/tests/views/test_user_mailinglists.py
```diff
@@ -49,8 +49,7 @@ def test_post_mailinglists_unauthenticated(self):
     def test_post_mailinglists_invalid(self):
         request_body = {
             'wrong_mailinglist_name': True,
-            'avalanche': 'incorrect_value',
-            'avalanche': False
```
asaunier: There were 2 ``avalanche`` entries in order to test duplicate entries. Do ypu think it's not relevant?

c2corg_api/views/user_mailinglists.py
```diff
@@ -3,27 +3,31 @@
...
+from c2corg_common.attributes import mailinglists as valid_mailinglists
 from cornice.resource import resource
 
 
 def validate_mailinglist_statuses(request, **kwargs):
-    tested = []
-    data = request.POST
+    mailinglists = {}
+    tested = set()
+    data = request.json_body
```
asaunier: Do! Of course, I had forgotten to use ``json_body`` here as well.

c2corg_api/views/user_mailinglists.py
```diff
@@ -66,21 +70,23 @@ def post(self):
...
-        subscribed_listnames = list(subscribed_lists.keys())
+        subscribed_listnames = set(subscribed_lists.keys())
 
         lists_to_add = []
         removed_lists = False
-        data = self.request.json_body
-        for listname in data:
-            if data[listname] and listname not in subscribed_listnames:
+        data = self.request.validated['mailinglists']
+        for listname in valid_mailinglists:
```
asaunier: I don't think this is correct. Here we should loop among the mailing lists provided by the request (ie. the actual changes to do), not among all the available mailing lists.

Reminder: the request provides only the ML whose status has changed. Eg. ``"avalanche": True, "lawinen": False``.

c2corg_api/tests/views/test_user_mailinglists.py
tsauerwein: I think duplicate entries are not really a problem, because you are sending a JSON object, which can not contain multiple values with the same key. :)

asaunier: True. Then perhaps the duplicate-detection part could be removed form the validator?

c2corg_api/views/user_mailinglists.py
tsauerwein: I changed also something below to make it work, but looping over `data` makes it easier.

c2corg_api/tests/views/test_user_mailinglists.py
tsauerwein: Yep, I removed it.

asaunier:
The PR is ready to be reviewed!

c2corg_api/models/mailinglist.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    Class containing mailing list subscriptions.
+    Based upon Sympa mailing list server https://www.sympa.org/
+    """
+    __tablename__ = 'subscriber_table'
+    __table_args__ = {"schema": sympa_schema}
+
+    listname = Column('list_subscriber', String(50), primary_key=True)
+    email = Column('user_subscriber', String(200), primary_key=True)
+    user_id = Column(
```
tsauerwein: I would have used `user_id` instead of `email` for the primary key. Because the email depends on the user id.

c2corg_api/models/user.py
```diff
@@ -172,3 +173,24 @@ def validate_password(self, plain_password):
...
     })
+
+# Make sure that user email changes are propagated to mailing lists as well
+trigger_ddl = DDL("""
+CREATE OR REPLACE FUNCTION users.update_mailinglists_email() RETURNS TRIGGER AS
+$BODY$
+BEGIN
+  UPDATE """ + sympa_schema + """.subscriber_table
+  SET user_subscriber = NEW.email
+  WHERE user_subscriber = OLD.email;
```
tsauerwein: Here I would also use the user id, because this is the field used to setup the relation between table `users` and `subscriber_table`. Because the email should be unique, your solution would also work.

c2corg_api/models/mailinglist.py
asaunier: That makes sense. I think I have used ``listname`` + ``email`` as a composite primary key because that was the one used in v5:
```
                        Table "public.subscriber_table"
           Column           |           Type           |       Modifiers        
----------------------------+--------------------------+------------------------
 list_subscriber            | character varying(50)    | not null
 user_subscriber            | character varying(100)   | not null
 date_subscriber            | timestamp with time zone | not null default now()
 update_subscriber          | timestamp with time zone | 
 visibility_subscriber      | character varying(20)    | 
 reception_subscriber       | character varying(20)    | 
 bounce_subscriber          | character varying(35)    | 
 bounce_score_subscriber    | integer                  | 
 comment_subscriber         | character varying(150)   | 
 subscribed_subscriber      | smallint                 | 
 included_subscriber        | smallint                 | 
 include_sources_subscriber | character varying(50)    | 
Indexes:
    "ind_subscriber" PRIMARY KEY, btree (list_subscriber, user_subscriber)
    "subscriber_idx" btree (user_subscriber, list_subscriber)
```
By the way we also have to take into account the fact that this table is used by another system (Sympa) that ignores the ``user_id`` and for which the "email" primary key is more relevant.
What do you think?

There's also a "reverse" index (user_subscriber, list_subscriber). I don't know if it is really needed. For instance because we don't use Sympa for writing in the table and because the table is not that big (~5000 rows as for now) so perhaps indices are not that critical?

c2corg_api/models/user.py
asaunier: OK I can change this. I guess it must be consistent with the primary key definition: use ``email`` or ``user_id`` depending what one is used in the primary key?

By the way I have just tested that the email is actually updated in the ML table when changing it in ``users.user``.

asaunier:
@tsauerwein Do you think it is safe to make the primary key change (use ``user_id`` instead of ``email``) considering the additional info about sympa I have provided?
If yes, I will make the change but I feel a bit concern that the sympa side is a bit penalized?!

c2corg_api/models/mailinglist.py
tsauerwein: > There's also a "reverse" index (user_subscriber, list_subscriber). Not created in v6. I don't know if it is really needed.

It depends on how Sympa queries this table. The original sympa database uses an index on `user_subscriber`:
https://github.com/softwerkskammer/server-sympa/blob/dad6a483420ed87583999848f7db055c4059713b/src/etc/script/create_db.Pg#L200-L226

It would make sense to do the same.

tsauerwein:
> Do you think it is safe to make the primary key change (use user_id instead of email) considering the additional info about sympa I have provided?

I think it's minor, because user id and email are in sync. I'll let you decide. :)

asaunier:
If it's quite equivalent, I prefer to keep it as is, it's simpler :)

PR #470, opened 2016-11-04
Views for sitemaps

tsauerwein:
This PR adds web-services that are consumed by the UI server-side to generate sitemaps for SEO:

* `/sitemaps` returns a list of URLs to request sitemap information for a document type (paginated). The web-service is used by the UI to generate a [sitemap index](http://www.sitemaps.org/protocol.html#index).
For example:
```
{
	"sitemaps": [{
				"url": "/sitemaps/w/0",
				"i": 0,
				"doc_type": "w"
			}, {
				"url": "/sitemaps/i/0",
				"i": 0,
				"doc_type": "i"
			}, {
				"url": "/sitemaps/i/1",
				"i": 1,
				"doc_type": "i"
			},
```

* `/sitemaps/{doc_type}/{i}` returns the information needed to generate a sitemap for a document type.
For example: `/sitemaps/w/0`
```
{
	"pages": [{
		"lastmod": "2016-10-28T11:07:39.540612",
		"document_id": 37191,
		"title": "Patraflon",
		"lang": "fr"
	}, {
		"lastmod": "2016-10-28T11:07:39.540612",
		"document_id": 37192,
		"title": "Pointe de Tosse",
		"lang": "fr"
	}, 
```

The sitemaps are cached for 1 day.

Closes https://github.com/c2corg/v6_api/issues/205

### DIFF
--- a/c2corg_api/views/sitemap.py
+++ b/c2corg_api/views/sitemap.py
@@ -1,0 +1,148 @@
+import logging
+
+from c2corg_api import DBSession
+from c2corg_api.models.cache_version import CacheVersion
+from c2corg_api.models.document import Document, DocumentLocale
+from c2corg_api.models.route import ROUTE_TYPE, RouteLocale
+from c2corg_api.models.user_profile import USERPROFILE_TYPE
+from c2corg_api.views import cors_policy
+from c2corg_api.views.validation import create_int_validator, \
+    validate_document_type
+from cornice.resource import resource, view
+from pyramid.httpexceptions import HTTPNotFound
+from sqlalchemy.sql.functions import func
+from math import ceil
+
+log = logging.getLogger(__name__)
+
+# Search engines accept not more than 50000 urls per sitemap,
+# see http://www.sitemaps.org/protocol.html
+PAGES_PER_SITEMAP = 50000
+
+
+validate_page = create_int_validator('i')
+
+
+@resource(
+    collection_path='/sitemaps', path='/sitemaps/{doc_type}/{i}',
+    cors_policy=cors_policy)
+class SitemapRest(object):
+
+    def __init__(self, request):
+        self.request = request
+
+    @view()
+    def collection_get(self):
+        """ Returns the information needed to generate a sitemap index file.
+        See: http://www.sitemaps.org/protocol.html
+
+        The response consists of a list of URLs to request the information
+        needed to generate the sitemap linked from the sitemap index.
+
+        E.g.
+
+            {
+                "sitemaps": [
+                    "/sitemaps/w/0",
+                    "/sitemaps/a/0",
+                    "/sitemaps/i/0",
+                    "/sitemaps/i/1",
+                    "/sitemaps/i/2",
+                    "/sitemaps/i/3",
+                    "/sitemaps/i/4",
+                    "/sitemaps/i/5",
+                    ...
+                ]
+            }
+        """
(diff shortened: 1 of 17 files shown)


c2corg_api/views/sitemap.py
```diff
@@ -0,0 +1,148 @@
...
+                    ...
+                ]
+            }
+        """
+        document_locales_per_type = DBSession. \
+            query(Document.type, func.count().label('count')). \
+            join(
+                DocumentLocale,
+                Document.document_id == DocumentLocale.document_id). \
+            filter(Document.type != USERPROFILE_TYPE). \
```
asaunier: Does it mean that the user profile are excluded from the sitemaps? If yes that looks good to me.

c2corg_api/views/sitemap.py
```diff
@@ -55,27 +59,11 @@ def collection_get(self):
...
-            ]
-            sitemaps.extend(sitemaps_for_type)
+        cache_key = _get_cache_key()
+        etag_cache(self.request, cache_key)
 
-        return {
-            'sitemaps': sitemaps
-        }
+        return cache_sitemap.get_or_create(
+            cache_key, _get_sitemap_index, expiration_time=-1)
```
asaunier: Does ``expiration_time=-1`` is the 1-day cache setting?

tsauerwein:
> Just curious: are the responses (with 50000 entries) big and long to generate?

For example `/sitemaps/o/0.xml` is ~9MB and took 3 seconds to download on my system (uncached).

c2corg_api/views/sitemap.py
tsauerwein: Exactly.

asaunier:
And how long to generate? Or is 3 sec the total time to get the answer on your system?
How many sub-sitemaps are needed? 10? 20?

What about the actual sitemaps generated by the UI? How long does it take to generate them?
Are they actually lighter then 10 MB?
http://www.sitemaps.org/faq.html#faq_sitemap_size

tsauerwein:
The request `/sitemaps/o/0.xml` goes towards the UI server. 3 seconds is the time to download the XML sitemap (this includes the time to get the info from the API and to generate the XML). This sitemap contains 50 000 entries.


> Are they actually lighter then 10 MB?

Maybe we should reduce the number of entries per sitemap to be on the safe side.

asaunier:
> Maybe we should reduce the number of entries per sitemap to be on the safe side.

They can be gzipped: http://www.sitemaps.org/faq.html#faq_compression

tsauerwein:
> They can be gzipped: http://www.sitemaps.org/faq.html#faq_compression

gzip should be enabled on nginx, for all requests.

tsauerwein:
> They can be gzipped: http://www.sitemaps.org/faq.html#faq_compression

And unzipped they must be less than 10MB.

asaunier:
> gzip should be enabled on nginx, for all requests

Is it already enabled or should we create an issue for this? @mfournier @arnaud-morvan ?

c2corg_api/views/sitemap.py
asaunier: Not sure to have understood where and how cache keys are configured.
What if we finally want to set the cache expiration time to another value? (just curious again)

asaunier:
Please merge if no other change is planned at short term.

tsauerwein:
Thanks for the review!

Issue #468, opened 2016-11-04
Problem with the use of sequence ID during the creation of a new article


dkocich:
There is a problem with the creation of the articles and ID for a new document.

Related to this UI issue https://github.com/c2corg/v6_ui/issues/839

Here I continue with previous discussion
...

@tsauerwein my sequence looks OK in DB

```
c2corg_dkocich=# \d guidebook.documents
                                            Table "guidebook.documents"
    Column    |          Type          |                                 Modifiers                                 
--------------+------------------------+---------------------------------------------------------------------------
 version      | integer                | not null default 1
 protected    | boolean                | 
 quality      | guidebook.quality_type | not null default 'draft'::guidebook.quality_type
 type         | character varying(1)   | 
 document_id  | integer                | not null default nextval('guidebook.documents_document_id_seq'::regclass)
 redirects_to | integer                | 
Indexes:
    "documents_pkey" PRIMARY KEY, btree (document_id)
    "ix_guidebook_documents_type" btree (type)
Foreign-key constraints:
    "documents_redirects_to_fkey" FOREIGN KEY (redirects_to) REFERENCES guidebook.documents(document_id)
Referenced by:
    TABLE "guidebook.area_associations" CONSTRAINT "area_associations_document_id_fkey" FOREIGN KEY (document_id) REFERENCES guidebook.documents(document_id)
    TABLE "guidebook.areas" CONSTRAINT "areas_document_id_fkey" FOREIGN KEY (document_id) REFERENCES guidebook.documents(document_id)
    TABLE "guidebook.articles" CONSTRAINT "articles_document_id_fkey" FOREIGN KEY (document_id) REFERENCES guidebook.documents(document_id)
    TABLE "guidebook.association_log" CONSTRAINT "association_log_child_document_id_fkey" FOREIGN KEY (child_document_id) REFERENCES guidebook.documents(document_id)
    TABLE "guidebook.association_log" CONSTRAINT "association_log_parent_document_id_fkey" FOREIGN KEY (parent_document_id) REFERENCES guidebook.documents(document_id)
    TABLE "guidebook.associations" CONSTRAINT "associations_child_document_id_fkey" FOREIGN KEY (child_document_id) REFERENCES guidebook.documents(document_id)
    TABLE "guidebook.associations" CONSTRAINT "associations_parent_document_id_fkey" FOREIGN KEY (parent_document_id) REFERENCES guidebook.documents(document_id)
[...]

PR #471, opened 2016-11-04
Fix issue 468 - problem when creating a new article (and waypoint)

dkocich:
Can it be this simple (one line fix)? Test for WP/Article works now. On UI I do not experience any problem now.

When we are creating new document without any previous associations, we have no `documents_to_update` and we can't remove any ignored documents when it has nothing archived yet that can be ignored.

Close https://github.com/c2corg/v6_api/issues/468

### DIFF
--- a/c2corg_api/models/cache_version.py
+++ b/c2corg_api/models/cache_version.py
@@ -357,6 +357,9 @@ def update_cache_version_for_map(topo_map):
 def update_cache_version_associations(
         added_associations, removed_associations, ignore_document_id=None):
     changed_associations = added_associations + removed_associations
+    if not changed_associations:
+        return
+
     documents_to_update = set()
     waypoints_to_update = set()
     routes_to_update = set()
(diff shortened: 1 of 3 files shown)


tsauerwein:
Thanks for setting up the test-case! I slightly changed your fix to stop earlier when no associations are available.

Please merge when Travis has passed.

c2corg_api/models/cache_version.py
```diff
@@ -372,7 +372,7 @@ def update_cache_version_associations(
                 association['child_type'] == OUTING_TYPE:
             routes_to_update.add(association['parent_id'])
 
-    if ignore_document_id is not None:
+    if ignore_document_id is not None and len(documents_to_update):
```
tsauerwein: It's considered more *pythonic* to write:
```
if my_list:
    ...
```
instead of:
```
if len(my_list):
    ...
```
or
```
if len(my_list) > 0:
    ...
```

See also: https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#True/False_evaluations

Issue #417, opened 2016-10-06
Text formating conversions to do in the migration script


asaunier:
- [x] `[q]` => `[quote]`
- [x] `[c]` => `[code]`
- [ ] ~~`[toc(.*)]` (regexp) => `[toc]`~~

Related to https://github.com/c2corg/v6_ui/issues/131#issuecomment-237195445 and https://github.com/c2corg/v6_ui/pull/695

asaunier:
* [ ] Also convert the wikilinks of old v5 URLs that are no longer supported in v6. For instance:
 * ``[[users/12345|Toto]]`` => ``[[profiles/12345|Toto]]``
 * ditto with ``summits``, ``huts``, ``parkings``, etc. => ``waypoints``.

asaunier:
* [ ] converting common bbcode tags to markdown?? See https://github.com/c2corg/v6_ui/issues/131#issuecomment-256396802

stef74:
@asaunier for me you conversion is probably an good idea but not realistic in term of planning ... So i Propose to close (no other action needed ?).

asaunier:
I prefer to leave it open until the migration is actually done.

tsauerwein:
I created a separate issue for the remaining problem (bbcode -> Markdown): c2corg/v6_api#516

PR #478, opened 2016-11-06
Migration convert tags

asaunier:
Related to https://github.com/c2corg/v6_api/issues/417

### DIFF
--- a/c2corg_api/scripts/migration/documents/document.py
+++ b/c2corg_api/scripts/migration/documents/document.py
@@ -210,7 +210,10 @@ class MigrateDocuments(MigrateBase):
 
         text = self.convert_q_tags(text)
         text = self.convert_c_tags(text)
-        text = self.convert_wikilinks(text)
+        text = self.convert_wikilinks(
+            text, MigrateDocuments.wikilink_waypoints_regex, 'waypoints')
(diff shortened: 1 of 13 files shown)


c2corg_api/scripts/migration/documents/document.py
```diff
@@ -203,3 +203,33 @@ def extract_summary(self, text):
...
+
+    wikilink_regex = re.compile(
+        '\[\[(users|summits|huts|parkings|sites|products)([^|]*)\|([^\]]+)\]\]')  # noqa
+
+    def convert_wikilinks(self, text):
+        match = MigrateDocuments.wikilink_regex.search(text)
+        if match:
+            route = 'profiles' if match.group(1) == 'users' else 'waypoints'
+            text = MigrateDocuments.wikilink_regex.sub(
+                r'[[%s\2|\3]]' % route, text)
```
tsauerwein: The problem was that you are doing the replacing with the regex that matches both `users` and `summit|huts|...`. So after the first call to `sub` all the wiki tags have already been replaced, either with `profiles` or `waypoints`.

I fixed it by having two different regexes for the two cases.

asaunier:
@tsauerwein Thanks for fixing the wikilinks replacements.
OK for merging?

Issue #481, opened 2016-11-07
Email change confirmation is not i18n-proof


asaunier:
See https://github.com/c2corg/v6_api/blob/de9e3a730f299b5fcc98183edaee04de7027e1a5/c2corg_api/emails/email_service.py#L74-L79

and not translate, see https://github.com/c2corg/v6_api/issues/406#issuecomment-258814201 for the strings translations.

We could use ``_send_email_with_link`` but perhaps ``user.lang`` is not available?

PR #485, opened 2016-11-07
Use i18n for message sent to confirm an email address change

asaunier:
Close https://github.com/c2corg/v6_api/issues/481


asaunier:
Thanks for reviewing.
Please note that I have not been able to test the change for real because Discourse is needed to change one's email address...

PR #493, opened 2016-11-09
Add migrating of V5 waterfall type - change migrating of V5 glacier

fbunoz:
In V5, I added a new type for the "summits" : "cascade".
Then, on V5 we are be able to change the V5 cliff summits to V5 cascade summits.
In the V6, the corresponding WP will have the good type, because the V5 cliff type doesn't exist in V6 (it will be migrated to climbing_outdoor, which is incompatible to manage ice fall routes).
In the migration, we need to migrate V5 cascade summit type to V6 cascade WP type. It is the purpose of this pull request.

More over, I changed the migrating of V5 glacier summit type, to locality instead of misc, because a glacier is similar to a valley, but colder :-)


arnaud-morvan:
I need it to set up the migration server, so I merge it.

PR #494, opened 2016-11-10
Remove unique contraint on image filename

arnaud-morvan:
Relate #363

I need this in master to automate the migration process.
This could be reverted when filenames will be unique in V5 database.

### DIFF
--- a/c2corg_api/models/image.py
+++ b/c2corg_api/models/image.py
@@ -10,7 +10,6 @@ from sqlalchemy import (
     SmallInteger,
     String
     )
-from sqlalchemy.ext.declarative import declared_attr
 
 from colander import MappingSchema, SchemaNode, Sequence
 from colanderalchemy import SQLAlchemySchemaNode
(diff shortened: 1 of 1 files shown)


asaunier:
Are you sure?
I think we need this constraint (for instance if we set an image removal tool => if filename is not unique, it could result in deleting a file potentially used by another image).

We have asked the association to remove the duplicates in the v5 DB before the final migration (in the meantime such a change must be used but not committed).

asaunier:
> I've removed the unique contraint on the filename to automate the migration process. This could possibly be reverted later.

OK then. But I'm afraid it will never be reverted. At least don't close the issue.

arnaud-morvan:
Sorry but I don't see another simple possibility to automate the migration process using c2corg/v6_image:master docker image.
Maybe continue the import on such error ?

asaunier:
It's OK to drop the constraints temporarly. I hope we will able to revert the change just before going live. So please merge the PR.

The  association devs told me that will most likely not fix the problem that causes image duplicates in the v5 DB but add tools to remove those duplicates. Then when v5 is made read-only they will run the tool and get rid of the duplicates before the guidebook migration tool is run.

Issue #492, opened 2016-11-09
Update SMTP config


asaunier:
Current SMTP server does not work for the demo server. Registering to a commercial service seems required. Once the service is set up, the config should be updated in Discourse and in the API:
https://github.com/c2corg/v6_api/blob/0d11bad2a003d20324c90f449c4c45fe3cbf8494/config/default#L64-L65

mfournier:
I sent the details (API key, etc) to @arnaud-morvan by email yesterday evening. So we should be close to done with this issue.

arnaud-morvan:
It seems OK for the settings, emails where sent from my workstation. I'm working problems with encoding to read mail templates files in utf-8.

fjacon:
tested on demov6 for the creation of a new account.
Dont see any encoding problem
Something to solve yet ?

arnaud-morvan:
Encoding problem has been fixed by https://github.com/c2corg/v6_api/pull/497/commits/355d456fb96f78f356d9eefb6d2bc0a183fdf9d6
Note that we should make a short test using this SMTP server with discourse.

asaunier:
@arnaud-morvan Shall we close this issue? It seems the emails are sent as expected in the demo.

stef74:
@arnaud-morvan do you close some tets confirm no issue ?

stef74:
@arnaud-morvan work needed on this topic ?

stef74:
fee free to reopen

PR #495, opened 2016-11-10
Add mail settings in config

arnaud-morvan:
This is needed for the demov6 server as we need SMTP authentication.

Relate #492


arnaud-morvan:
@tsauerwein : here is my last PR for today to enable emails on demov6

c2corg_api/emails/email_service.py
```diff
@@ -1,3 +1,5 @@
+import transaction  # NOQA
```
tsauerwein: Have you tested this?

arnaud-morvan: Yes with demov6 SMTP server, and I've received the email, I've change my password !!!

PR #497, opened 2016-11-10
Set env encoding in dockerfile

arnaud-morvan:
This is needed to read email templates with utf-8 characters.


arnaud-morvan:
@tsauerwein : can you take a look

PR #498, opened 2016-11-11
Loadtests tools

asaunier:
This PR adds scripts and data files needed to perform load testing with Gatling.

Commit bcb2731 is supposed to create some user accounts that will be used by Gatling to simulate authenticated user actions. But it seems buggy, when I run it:
* no error is output, creation messages are printed as expected
* instead of creating 10 new accounts, it removes all existing user accounts (0 row in table ``users.user`` after running the script!)
Not sure it explains the problem, but I think ``user.id`` of the "created" accounts start at 1!?

Gatling scenarios files also be commited into this ``loadtests`` directory.


c2corg_api/scripts/loadtests/create_test_users.py
```diff
@@ -0,0 +1,105 @@
...
+
+    engine = engine_from_config(settings, 'sqlalchemy.')
+
+    logging.basicConfig()
+    logging.getLogger('sqlalchemy.engine').setLevel(logging.WARN)
+
+    # create a fresh schema on the target database
+    Session = sessionmaker(extension=ZopeTransactionExtension())  # noqa
+    session = Session(bind=engine)
+    Base.metadata.drop_all(engine, checkfirst=True)
```
tsauerwein: You want to add the test users to a migrated database (containing all documents, users, ... of v5)? In that case you shouldn't use `drop_all`.

c2corg_api/scripts/loadtests/create_test_users.py
```diff
@@ -0,0 +1,105 @@
...
+
+    connection = engine.connect()
+
+    for i in range(1, 11):
+        username = 'testuserc2c' + str(i)
+        password = username
+        email = username + '@foo.bar'
+        lang = 'fr'
+
+        user = schema_create_user.objectify({
```
tsauerwein: Why are you using the schema? You could directly create a `User` object:
```
user = User(username=username, ...)
```

c2corg_api/scripts/loadtests/create_test_users.py
```diff
@@ -0,0 +1,105 @@
...
+
+        print('Created user ' + username)
+
+    # profile can be indexed by ES
+    # FIXME Is request available here?
+    # notify_es_syncer(request.registry.queue_config)
+
+    # whatever value? not really used by UpdateSequences?
+    batch_size = 1000
+    UpdateSequences(connection, session, batch_size).migrate()
```
tsauerwein: This should not be required.

c2corg_api/scripts/loadtests/create_test_users.py
asaunier: Arf :) Thanks for pointing this out. It explains why the users table (and all others?) was empty :P 
It seems I have copied-pasted a bit too much code from the migration script.

c2corg_api/scripts/loadtests/create_test_users.py
asaunier: I use the logic from https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/user.py#L130-L131
OK, I will simply use the ``User`` model.

asaunier:
I have added a fixup commit to fix errors @tsauerwein has reported: c94e60e

There is then an error when running the script:
```
sqlalchemy.exc.IntegrityError: (psycopg2.IntegrityError) duplicate key value violates unique constraint "es_sync_status_pkey"
DETAIL:  Key (id)=(1) already exists.
 [SQL: 'INSERT INTO guidebook.es_sync_status (last_update, id) VALUES (%(last_update)s, %(id)s)'] [parameters: {'id': 1, 'last_update': None}]
```

c2corg_api/scripts/loadtests/create_test_users.py
```diff
@@ -38,29 +35,27 @@ def main(argv=sys.argv):
     # create a fresh schema on the target database
     Session = sessionmaker(extension=ZopeTransactionExtension())  # noqa
     session = Session(bind=engine)
-    Base.metadata.drop_all(engine, checkfirst=True)
     setup_db(engine, session)
```
tsauerwein: `setup_db` tries to re-create the database schema. I guess you don't want to do that?

asaunier:
I have removed ``setup_db`` which is indeed not wanted. The script now passes and reports it has created 10 new accounts... that I cannot see in my database (no user was actually added to table ``users.user``). Is there something missing to actually write the new data?

tsauerwein:
> that I cannot see in my database (no user was actually added to table users.user). Is there something missing to actually write the new data?

Can you try do run in a transaction like here? https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/set_default_geometries.py#L15-L22

asaunier:
> Can you try do run in a transaction like here? 

Thanks for the tip!
I have done so with 7068405 and added a minimal ``DocumentGeometry`` (else the profile crashes) => the script now works and the user accounts are created as expected. It's possible to authenticate, show the profile and create an outing.
Looks like we finally did it :) Thanks again for the precious help @tsauerwein

tsauerwein:
That is looking good!

It would probably also make sense to load test the advanced search by directly querying the API. Because this is a service that will be used a lot.

asaunier:
Thanks for the review @tsauerwein 
I am working on more scenarios, including the advanced search (but will make an additional PR).

Issue #262, opened 2016-04-27
3D/4D geometries seem not supported


asaunier:
When sending a 4D route geometry (with elevation + timestamp for each point) like

```
{"type":"MultiLineString","coordinates":[[[849344.3376596108,5800875.452693369,1691,1460269443],[849385.5258712043,5800824.053639301,1679,1460270468]
```

the API refuses the document arguing that "Invalid geometry"
https://github.com/c2corg/v6_api/blob/947c2284557a31564e121ed0082a348f882eaad1/c2corg_api/ext/colander_ext.py#L76-L81

OK with the same geometry passed in 2D with only lat/lon:

```
{"type":"MultiLineString","coordinates":[[[849344.3376596108,5800875.452693369],[849385.5258712043,5800824.053639301]
```

No idea if this works with the elevation but no timestamp.

Supporting elevation and timestamp would be useful for displaying a profile of the route/outing as well as a "time profile". See in v5: http://www.camptocamp.org/outings/745686/fr/blanc-de-moming-en-boucle-mountet-glacier-de-moming-par-l-arete-sw#elevation_profile_container

By the way timestamps in geometries (for outings) would be useful in the future for instance to automatically geolocalize the pictures associated to the outing.

tsauerwein:
Yes, not (yet) supported, see https://github.com/c2corg/v6_api/issues/181.

asaunier:
> Yes, not (yet) supported, see #181.

OK, good to know: the 3D is lost during the migration as well.

fjacon:
@asaunier @tsauerwein 
Should we close this ?
Or create a new one about migration because 3D is lost ?

asaunier:
Why closing the issue? We still have the problem:
* elevation/time dimensions of the outing geometries are lost during the migration (I think route geometries are 2D in v5 - a bit a pity!?)
* 3D/4D tracks (eg. provided by submitting a GPX file) are not supported by the API (because of the geometry cols definition in the DB?)

By the way, the map editing tool provided in v6 is 2D. No elevations are detected when drawing a track on the map. An elevation service would be needed. (not in the specs)

tsauerwein:
@fjacon What is the priority for this?

asaunier:
Don't we need this before 6.0? If those data are not migrated for 6.0, they will be lost, won't they?

tsauerwein:
> If those data are not migrated for 6.0, they will be lost, won't they?

Indeed.

PR #504, opened 2016-11-17
Support 3D and 4D geometries

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/262

I had to do quite a few changes because Shapely does not support 4D geometries when parsing/writing WKB values.

I am currently running a full migration to test if all documents can be migrated.

### DIFF
--- a/c2corg_api/ext/colander_ext.py
+++ b/c2corg_api/ext/colander_ext.py
@@ -4,12 +4,10 @@
 from colander import (null, Invalid, SchemaType)
 
 from geoalchemy2 import WKBElement
-from geoalchemy2.shape import to_shape, from_shape
-from shapely.geometry import mapping, shape
-from shapely.ops import transform
-from functools import partial
-import pyproj
+from geomet import wkb
+from geoalchemy2.compat import buffer, bytes
 import json
+import geojson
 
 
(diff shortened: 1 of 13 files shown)


asaunier:
That looks like a great job, thanks!

If I understand your comment in https://github.com/geoalchemy/geoalchemy2/pull/141#issuecomment-261048195 we use the same geometry field to store 2D/3D/4D geometries, whether point, line or polygon, right?

So it should be transparent and support indifferently geometries drawn in the UI editing maps (2D) or uploaded GPX tracks?

tsauerwein:
> If I understand your comment in geoalchemy/geoalchemy2#141 (comment) we use the same geometry field to store 2D/3D/4D geometries, whether point, line or polygon, right?
> 
> So it should be transparent and support indifferently geometries drawn in the UI editing maps (2D) or uploaded GPX tracks?

Yes, correct. Currently we do not enforce that geometries for routes and outings are only lines, or only polygons for areas (see https://github.com/c2corg/v6_api/issues/506).

tsauerwein:
Ok, I verified that 4D geometries are migrated correctly and that they are returned with 4 dimensions.

But the UI only sends 2D geometries to the API when trying to upload a GPX.

PR #508, opened 2016-11-18
Gatling scenarios

asaunier:


asaunier:
This PR is almost ready and reviewing may be started.

4 scenarios are now available:
* forum anonymous (reading)
* guidebook anonymous (reading)
* guidebook advanced search
* guidebook auth (authenticating, uploading an image, adding an outing)
+ a global scenario that pick randomly among the scenarios above.

I am still struggling with the forum login which does not work as expected (a redirection that fails when requesting the forum's ``/session/sso`` URL)...

tsauerwein:
Looking good! I am curious to get some feedback from these tests.

asaunier:
@tsauerwein thanks for the review!

I still have a problem with the "auth in forum and post a message" scenario (a commit to come hopefully).

Actually Gatling reports a 401 error instead of a 302 status when accessing http://forum.demov6.camptocamp.org/session/sso?return_path=%2F => it should redirect to the UI auth page with a URL such as http://www.demov6.camptocamp.org/auth-sso?sso=bm9uY2U9OTczMjE5NDAxYzY3MmIwMTFiNjNlZmM3YTNhZTNiYWUmcmV0dXJu%0AX3Nzb191cmw9aHR0cCUzQSUyRiUyRmZvcnVtLmRlbW92Ni5jYW1wdG9jYW1w%0ALm9yZyUyRnNlc3Npb24lMkZzc29fbG9naW4%3D%0A&sig=5d67dd428e256c7f2921497d06cec61be9907b790aeb086f9a67ac127166777a

Here is a minimal gatling scenario that illustrate the problem:
```
package c2corg
import scala.concurrent.duration._
import io.gatling.core.Predef._
import io.gatling.http.Predef._

class TestLogin extends Simulation {
  val httpProtocol = http.baseURL("http://forum.demov6.camptocamp.org")
  val scn = scenario("TestLogin")
    .exec(http("Forum SSO")
      .get("/session/sso?return_path=%2F")
      .headers(Map("Accept" -> "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"))
      .check(
        status.is(302)
      ))
  setUp(scn.inject(atOnceUsers(1))).protocols(httpProtocol)
}
```

It fails with error
> status.find.is(302), but actually found 401

No problem with a simple cURL request:
```
$ curl -I http://forum.demov6.camptocamp.org/session/sso?return_path=%2F
HTTP/1.1 302 Found
Server: nginx
Date: Thu, 24 Nov 2016 14:58:58 GMT
Content-Type: text/html; charset=utf-8
X-Frame-Options: SAMEORIGIN
X-XSS-Protection: 1; mode=block
X-Content-Type-Options: nosniff
X-Discourse-Route: session/sso
Location: http://www.demov6.camptocamp.org/auth-sso?sso=bm9uY2U9ZTllNDYzMzhlMjQ2YjIwZjY4YWRhYzA2YjhkMWJlMzkmcmV0dXJu%0AX3Nzb191cmw9aHR0cCUzQSUyRiUyRmZvcnVtLmRlbW92Ni5jYW1wdG9jYW1w%0ALm9yZyUyRnNlc3Npb24lMkZzc29fbG9naW4%3D%0A&sig=314543fa02d4a2a3688405eb05f701ad9650d2d4584e6db980cde0fadc2a41f5
Set-Cookie: _forum_session=bllrNjhKdG4vZzM1SnZZRU9yL29iK0NOQVpWQTJKNk84UnppYWdMSUtjWkhwWnFCS0xmclErNTF6M1RJbmRPV0VoR3dXVVduVXV3NS80Y2JicEFQaXc9PS0tTjB2ZGN2NGw0S01Dc1hSNjBFaUJHQT09--2756be2164ba9e18e2f9404f114fc2cf13323983; path=/; HttpOnly
X-Request-Id: 71c77a92-1a28-4e78-9b59-7cc14f968439
X-Runtime: 0.007142
X-UA-Compatible: IE=edge
```

tsauerwein:
I am afraid I won't have the time to help with that problem now. Do you want to exclude this scenario so that we already can test with the others?

asaunier:
@bquartier gave me a tip about the gatling error (use ``disableFollowRedirect`` to prevent gatling from automatically be redirected).

I have updated and rebased the PR and will merge when Travis is OK, unless you have other comments.
An overview of the scenarios can be found in the global scenario with the ponderation/weight of each of the 5 scenarios available: https://github.com/c2corg/v6_api/pull/508/commits/dd91489364c8a7d5828ae2a021ab3f0e0334f0a5

Next step is to test in the actual load testing environnement but it's not available yet (in the mean time I have only made light testing on the demov6 instance).

PR #513, opened 2016-11-21
Create dummy routes for climbing sites

### LINKED ISSUE #351: Migration of the outings associated to climbing sites
In v5, it is possible to create an outing from a "climbing site" page => no route is associated to the outing. See for instance: http://www.camptocamp.org/outings/538339/fr/gaillands (associated to the climbing site http://www.camptocamp.org/sites/102326/fr/gaillands)

In v6:
- the climbing sites pages are transformed into waypoint pages, for ex. http://c2corgv6.demo-camptocamp.com/waypoints/102326/fr
- the matching v5 outings are actually migrated to v6: http://c2corgv6.demo-camptocamp.com/outings/538339/fr but have a waypoint-outing association, which is no longer possible in the v6 interface (the association tool in outing pages only allows to associate route).

@tsauerwein do you think it would be possible when migrating such outings (not associated to a route but to a climbing site) to create a basic route (if not already existing) which would be named the same way than the climbing_site waypoint and to associate it between the WP and the outing?

(Of course we could keep the current way those outings are migrated if the above solution is too complicated - I didn't realize until now it was possible- but that would be a bit confusing and not consistent to have those old outings still associated to the climbing site waypoints.)

tsauerwein:
See https://github.com/c2corg/v6_api/issues/351#issuecomment-258868461

Related to #351 

Now, direct associations between outings and waypoints are no longer possible.

### DIFF
--- a/c2corg_api/scripts/migration/climbing_site_routes.py
+++ b/c2corg_api/scripts/migration/climbing_site_routes.py
@@ -1,0 +1,290 @@
+import transaction
+from c2corg_api.models.association import Association, AssociationLog
+from c2corg_api.models.document import DocumentGeometry, \
+    ArchiveDocumentGeometry
+from c2corg_api.models.document_history import HistoryMetaData, DocumentVersion
+from c2corg_api.models.outing import OUTING_TYPE
+from c2corg_api.models.route import Route, RouteLocale, ArchiveRoute, \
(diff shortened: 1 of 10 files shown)


asaunier:
Since it's an important issue, I would prefer the PR does not close the issue (should be marked as "ready for demo" instead). So that the association figures out they have to test it in the demo.

Issue #414, opened 2016-10-06
Access rights for changing associations of outings


tsauerwein:
Only the creator, associated users and moderators should be able to change associations of outings.

See https://github.com/c2corg/v6_ui/issues/694#issuecomment-251709488

stef74:
@tsauerwein when do you plan to fix this issue ?

PR #517, opened 2016-11-22
Check permission for associations with outings

tsauerwein:
Only moderators and participants can create or remove associations to an outing.

Closes https://github.com/c2corg/v6_api/issues/414


tsauerwein:
Thanks for the review!

PR #528, opened 2016-11-23
Make sure docs can be created/updated without quality field

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/394


asaunier:
Looks good to me.

Just curious though to know in what cases we would have documents created with no quality attribute provided?

tsauerwein:
> Just curious though to know in what cases we would have documents created with no quality attribute provided?

If the UI always provides a value for quality, the better. But it's not a required field.

Issue #529, opened 2016-11-23
Simplify route and outing geometries


tsauerwein:
From https://github.com/c2corg/v6_api/issues/489#issuecomment-262524104

- During the migration simplify the geometries for outings and routes.
- If a new geometry is set on an outing or route, store a simplified geometry.

Geometries are already simplified on the client-side before being sent to the API, so the 2nd part is not that critical for the Go Live.

asaunier:
Example of big route geometry:
http://www.camptocamp.org/routes/211687/fr/mont-de-grange-traversee-bel-air-coicon
The GPX file available on v5 is > 800 kB!

PR #532, opened 2016-11-24
Simplify route and outing geometries

tsauerwein:
Geometries for routes and outings are simplified with a tolerance of 5m during the migration and on insert/update.

Closes https://github.com/c2corg/v6_api/issues/529

### DIFF
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -1,6 +1,8 @@
 import enum
 
 import abc
+from functools import partial
+
 import geoalchemy2
 from c2corg_api.ext import colander_ext
 from c2corg_api.models import Base, schema, DBSession, enums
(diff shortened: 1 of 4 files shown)


c2corg_api/models/document.py
```diff
@@ -437,6 +439,66 @@ class ArchiveDocumentGeometry(Base, _DocumentGeometryMixin):
...
+guidebook_documents_geometries_archives_geometries_insert_or_update
+BEFORE INSERT OR UPDATE OF geom_detail
+ON guidebook.documents_geometries_archives
+FOR EACH ROW
+EXECUTE PROCEDURE guidebook.simplify_geom_detail();
+""")
+event.listen(Base.metadata, 'after_create', trigger_ddl)
+
+
+def should_drop_trigger(table_name, ddl, target, connection, **kw):
```
tsauerwein: This is only relevant for our tests, not for the production database.

tsauerwein:
Thanks for the review!

PR #535, opened 2016-11-25
During cliff mifgration, add multi-pitch attribute

fbunoz:
On the V6, a multi-pitch cliff or a single-pitch cliff are described by a climbing_outdoor waypoint, which has a field "routes_type" to indicate if the cliff has single pitch, multi pitch, clock, etc.
During the migration : 
- The V5 'sites' are migrated into climbing_outdoor waypoint, with 'routes_types' taking the V5 value.
- The V5 'summit' cliffs are migrated into climbing_outdoor waypoint. The 'routes_types' must be set to "multi', as a V5 summit-cliff can only has multi-pitch routes.

This PR changes the migration script to add the 'multi' attribute during the summti-cliff migration.

The script has not been tested.

### DIFF
--- a/c2corg_api/scripts/migration/documents/waypoints/summit.py
+++ b/c2corg_api/scripts/migration/documents/waypoints/summit.py
@@ -49,6 +49,7 @@ class MigrateSummits(MigrateWaypoints):
             version=version,
             waypoint_type=self.convert_type(
                 document_in.summit_type, MigrateSummits.summit_types),
+            climbing_outdoor_types=outdoor_types,
             protected=document_in.is_protected,
             redirects_to=document_in.redirects_to,
             elevation=document_in.elevation,
@@ -69,6 +70,14 @@ class MigrateSummits(MigrateWaypoints):
             description=description,
             summary=summary
         )
+
+    def get_climbing_outdoor_types(self, document_in):
+        outdoor_types = []
+        if document_in.summit_type == 4:
+            outdoor_types.append('multi')
+        
+        outdoor_types = outdoor_types if outdoor_types else None
+        return outdoor_types
 
     summit_types = {
         '1':   'summit',            # was 'culmen'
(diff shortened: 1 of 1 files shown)


tsauerwein:
Thanks, that is looking good! I will test when I run the migration the next time.

Small tip: Locally you can run the linter with `make -f config/... lint`. And to avoid these "fixup commits", you can commit with `git commit --amend` or use `git commit --fixup=...`. See also:
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History
https://robots.thoughtbot.com/autosquashing-git-commits

c2corg_api/scripts/migration/documents/waypoints/summit.py
```diff
@@ -70,6 +72,13 @@ def get_document_locale(self, document_in, version):
             summary=summary
         )
 
+    def get_climbing_outdoor_types(self, document_in):
+        outdoor_types = []
+        if document_in.summit_type == 4:
+            outdoor_types.append('multi')
+        outdoor_types = outdoor_types if outdoor_types else None
+        return outdoor_types
```
asaunier: This function is a bit weird. Why making ``outdoor_types`` a list if you make it ``None`` when empty. I would suggest:
```
return ['multi'] if document_in.summit_type == 4 else None
```
By the way do you really need a function to do this?

fbunoz: I simply copied the structure of this function : https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/migration/documents/waypoints/sites.py#L112-L141
I didn't try understanding why each line has been coded like this.
You can change the code as you want.

asaunier: > I didn't try understanding why each line has been coded like this.

You should. Understanding what you do is important.

The code you are pointing is much more complex than the one you have written. Then you could have optimized your code as I have suggested it in order to have a simpler and more readable code. Please also squash your commits when creating a PR unless having distinct commits helps understand the changes.

I won't change the code. Not my job to make the clean up.

PR #540, opened 2016-11-28
Search and sort reports by date to show them in cards correctly

dkocich:
Simple search sorting by date works now, but I am still not sure how to sort reports also in the advanced search...


tsauerwein:
> but I am still not sure how to sort reports also in the advanced search...

See https://github.com/c2corg/v6_api/blob/master/c2corg_api/search/search_filters.py#L45-L49

dkocich:
Thanks. Works fine now.

PR #543, opened 2016-11-28
Allow changes only to collaborative/my own articles

dkocich:
Ready to review
Close https://github.com/c2corg/v6_ui/issues/1029

### DIFF
--- a/c2corg_api/views/article.py
+++ b/c2corg_api/views/article.py
@@ -66,7 +66,8 @@ class ArticleRest(DocumentRest):
             article_id = self.request.validated['id']
             article = DBSession.query(Article).get(article_id)
             if article is None:
-                raise HTTPNotFound('No article found for id %d' % article_id)
+                raise HTTPNotFound(
+                    'No article found for id {}'.format(article_id))
             if article.article_type == 'collab':
                 article_type = \
                     self.request.validated['document']['article_type']
--- a/c2corg_api/views/article.py
+++ b/c2corg_api/views/article.py
@@ -1,4 +1,6 @@
 import functools
+
+from c2corg_api.models import DBSession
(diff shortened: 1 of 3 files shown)


tsauerwein:
Could you please add tests for this?

c2corg_api/views/article.py
```diff
@@ -57,6 +62,23 @@ def collection_post(self):
...
+                    self.request.validated['document']['article_type']
+                if article_type != article.article_type:
+                    raise HTTPBadRequest(
+                        'Personal article cannot be changed'
+                    )
+            # personal articles should only be modifiable by
+            # their creator and moderators
+            elif not has_been_created_by(article_id,
+                                         self.request.authenticated_userid):
+                raise HTTPForbidden('No permission to change this image')
```
tsauerwein: image?

c2corg_api/views/article.py
```diff
@@ -57,6 +62,24 @@ def collection_post(self):
                 validate_article_update,
                 validate_associations_update])
     def put(self):
+        if not self.request.has_permission('moderator'):
+            article_id = self.request.validated['id']
+            article = DBSession.query(Article).get(article_id)
+            if article is None:
+                raise HTTPNotFound('No article found for id %d' % article_id)
```
tsauerwein: Elsewhere we are using `'... {}'.format(article_id)`.

dkocich: Ok, sometimes yes, sometimes no - like [here for image](https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/image.py#L144). I will change both now but there are still around +-50 occurrences of "...' % ..." elsewhere. I am not sure if they all can be just simply changed to `.format` instead

PR #545, opened 2016-11-29
log in json in production

mfournier:
The goal is to have all log messages wrapped in json objects, so they can be cleanly indexed in the downstream elasticsearch. Most importantly: this way the stacktraces are no longer split on multiple lines.

Example:
https://c2corgv6.kibana.infra.internal/kibana/#dashboard/temp/AVivtz-lM9QtOkkMtIRW

(BTW, you might want to take a look at this stack trace, as it seems to be a genuine error)

### DIFF
--- a/production.ini.in
+++ b/production.ini.in
@@ -28,7 +28,7 @@ port = 6543
 ###
 
 [loggers]
-keys = root, c2corg_api, sqlalchemy, c2corg_api_syncer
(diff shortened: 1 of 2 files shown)


tsauerwein:
About the error ('relation "guidebook.reports" does not exist'): The version of the API seems not to be the same version that was used to make the migration.

mfournier:
Yes, I confirm there was a mismatch between the code version and the dataset. My bad!

PR #546, opened 2016-11-29
health.py: emit HTTP 500 errors on critical conditions

mfournier:


tsauerwein:
In case of an error, would it make sense to return the original structure and only set the return code to 500? This way you would know if the other parts of the system are still ok.

mfournier:
On Tue, Nov 29, 2016 at 04:56:14AM -0800, Tobias Sauerwein wrote:
> In case of an error, would it make sense to return the original structure
> and only set the return code to 500? This way you would know if the other
> parts of the system are still ok.

So in production, these probes will be hidden behind haproxy anyways. ie:
they'll be out of reach exactly when we need them the most unfortunately.
The good news is that we have plenty of other ways to know what's wrong
down the pipes.

That said, this PR "allow edits from maintainers", so feel free to rework
this to your taste ;-)

I'd really need to have at least this first version in the v6.0-beta.2
tomorrow though.

Thanks !

PR #547, opened 2016-11-29
Fix username normalization script

arnaud-morvan:


c2corg_api/scripts/migration/create_unique_normalized_usernames.sql
```diff
@@ -108,16 +108,19 @@ CREATE UNIQUE INDEX ix_punbb_users_lower_username_unique
   (lower(username_unique::text) COLLATE pg_catalog."default");
 
 UPDATE punbb_users
-SET username_unique = substring(username_normalized from 1 for  24) || number
+SET
+  username_unique = CASE
+    WHEN number > 1 THEN substring(username_normalized from 1 for  24) || number
+    ELSE username_normalized
```
asaunier: Does this mean that when duplicates are found, the first one is not changed, only the following ones?

arnaud-morvan:
Now I use the id as suffix like proposed in https://github.com/c2corg/v6_forum/issues/22#issuecomment-222550729.
All seems to work fine, will run the next topoguide migration with this.

arnaud-morvan:
Another problem with the migration, we now have duplicated on username (topoguide username) in V5 database.

asaunier:
You mean the username used for authenticating (login)?
Why do we have such duplicates? Does the normalisation script change this attribute?

arnaud-morvan:
No, but we now have 24 duplicated logins in V5 database:
```
SELECT login_name, count(*)
FROM app_users_private_data
GROUP BY login_name
HAVING count(*) > 1
```
give me 25 lines

asaunier:
OK so it means there are integrity problems in the v5 database with attribute ``login_name``. I guess it's not  new. Can we live with this problem or does it break the migration?

arnaud-morvan:
We have, and we must have an unique constraint on login name in V6 database, this can NOT be overridden.
There should be one in V5 database too.
This is a real blocking issue. Will open one in v6_api for today ... will see that tomorrow.

asaunier:
OK, I will ask the association to find a solution.

I'm just a bit surprised: why do we have this problem now? it's not the first time the users are migrated though and all the duplicates have been created between 1998 and 2001 according to your query in https://github.com/c2corg/v6_api/issues/552#issuecomment-263688795

tsauerwein:
> I'm just a bit surprised: why do we have this problem now?

@arnaud-morvan Can you check with a fresh import of the v5 database? If I am not mistaken you've run your script multiple times, and the error could be a result of that.

arnaud-morvan:
No, because the lines are really different (different emails, different names, nothing in common except login name)

asaunier:
I have run the SQL query on the v5 prod server: there is indeed 24 duplicates, created before 2002.

Issue #476, opened 2016-11-06
Set up Alembic for database migrations


tsauerwein:
After the go-live we will need Alembic to update the production database.

* [x] Set up Alembic
* [x] Integrate with the Docker infrastructure

asaunier:
I don't think this issue should be tagged "blocking".

Alembic is a tool used to propagate structure changes in an existing database:
http://alembic.zzzcomputing.com/en/latest/

As for now, when a new table is added to the database or a new column is added to a table, we simply drop and recreate the database of the demo. When the database will be in prod, it will no longer be possible to do so.

I don't think we need such a tool before the golive, or even not before changes in the productive data structure are required.

@tsauerwein could you confirm?

tsauerwein:
If we urgently have the need to do structural changes to the production database, it would be very useful to have Alembic already set up. For sure, you can also simply run the commands on the production database but then how do you remember which state your database is in, which commands you have run,... These kind of problems Alembic was invented for.

stef74:
@tsauerwein according to your remarks ... we can consider is important but not blocking ?

tsauerwein:
Yes, I started working on it.

tsauerwein:
Alembic is set up. What is left to do is to figure out how to run a database migration in the Docker environment. But this is not blocking (I removed the label) and can be done after the go live.

tsauerwein:
Upgrading the production database with Alembic is documented here: https://github.com/c2corg/v6_api/wiki/Upgrade-the-prod-server

PR #554, opened 2016-11-30
Set up alembic for database migrations

tsauerwein:
Related to https://github.com/c2corg/v6_api/issues/476


alembic_migration/README.md
```diff
@@ -0,0 +1,33 @@
...
+[documentation](http://alembic.zzzcomputing.com/en/latest/cookbook.html#replaceable-objects)
+is used.
+
+## Run a migration
+
+A migration should be run each time the application code is updated or if you have just created a migration script.
+
+```
+.build/venv/bin/alembic upgrade head
+```
```
asaunier: This command should be run on the target server, right?

asaunier:
All the triggers and function definitions are moved to an initial alembic script. Right?
When creating a new instance with a new DB, does this mean alembic is called automatically? I guess it's here https://github.com/c2corg/v6_api/blob/494d48dcc4097aba440d658994bef3e65eaa49b1/c2corg_api/scripts/initializedb.py#L42 right?

Just a question to understand how this works.

asaunier:
@tsauerwein is there something still required before merging the PR? Except approval :P 

Looks good to me. I have tested the branch in my API instance. At least it still works :)

tsauerwein:
I tested the migration this morning, and have to commit 1-2 small changes.

tsauerwein:
@arnaud-morvan As a reminder: Once this is merged, you will have to adapt the script `migrate_topoguide` to also create the schema `alembic`.

alembic_migration/README.md
tsauerwein: Yes, on the database that should be migrated.

PR #555, opened 2016-11-30
Include author for article

tsauerwein:
Needed for https://github.com/c2corg/v6_ui/pull/961#issuecomment-263886337


asaunier:
So now the name of the contributor that has created the article (whether personal or collaborative) is returned in an ``author`` attribute. Right?

That's probably not needed for collaborative articles (?) but at least it is consistent for all articles.
Looks good to me!

asaunier:
I guess it works also for migrated articles?

tsauerwein:
For all articles.

PR #558, opened 2016-12-01
Fix xreports diff of two documents

dkocich:
Rename `class Xreports` to `Xreport`
Rename `test_report.py` to `test_xreport.py`

Disable private/public check and always return public version of the xreport


asaunier:
Looks good to me

PR #562, opened 2016-12-03
Make sure routes and outings have a default geometry

tsauerwein:
Related to https://github.com/c2corg/v6_ui/issues/959

* If a route has no track geometry or main waypoint, the centroid of the convex hull of all associated waypoints is used.
* If an outing has no track geometry, the centroid of the convex hull of the default geometry of all associated routes is used.

This PR also checks that the last associated waypoint of a route and the last associated route of an outing, can not be disassociated.


asaunier:
Really cool, thanks!
I have tested, that works fine :)

tsauerwein:
Thanks for testing!

PR #571, opened 2016-12-06
Add author field to xreport detail

dkocich:
Add xreport's author to the API response (similar as for articles).

image


asaunier:
I guess that's the same than for outings, right?
Looks good to me then.
@tsauerwein OK for you?

Issue #520, opened 2016-11-22
Only moderators can remove an association from a collaborative document


tsauerwein:
See https://github.com/c2corg/v6_ui/issues/836#issuecomment-258503516
And: https://github.com/c2corg/v6_ui/pull/961#issuecomment-263911599

stef74:
to be homogeneous with UI I add blocking tag

PR #575, opened 2016-12-07
Check association permissions

tsauerwein:
* Moderators are allowed to do everything
* Normal users can add associations to outings if they are participant of the outing
* Normal users can add associations to reports, personal images and personal articles if they are creator of the document
* Normal users are not allowed to remove associations from collaborative documents (except associations to personal documents)

Closes https://github.com/c2corg/v6_api/issues/520

### DIFF
--- a/c2corg_api/tests/views/test_association.py
+++ b/c2corg_api/tests/views/test_association.py
@@ -136,7 +136,7 @@ class TestAssociationRest(BaseTestRest):
             'no rights to modify associations with outing {}'.format(
                 self.outing.document_id))
 
-    def test_add_association_wa_article_collab(self):
+    def test_add_association_wc_article_collab(self):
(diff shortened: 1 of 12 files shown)


c2corg_api/tests/views/test_association.py
```diff
@@ -133,6 +136,172 @@ def test_add_association_uo_no_rights(self):
             'no rights to modify associations with outing {}'.format(
                 self.outing.document_id))
 
+    def test_add_association_wc_article_collab(self):
+        """ Check that associations with articles can only be changed by
```
asaunier: only? :P

asaunier:
I think we can merge this PR now. Thanks!

tsauerwein:
Thanks for the review!

PR #586, opened 2016-12-12
pull in gunicorn alternative worker class dependencies

mfournier:
cf http://gunicorn-docs.readthedocs.io/en/stable/settings.html#worker-class


tsauerwein:
Ok, feel free to merge. Same for the UI PR.

Issue #534, opened 2016-11-24
es-syncer does not complain if no connection to Redis


tsauerwein:
If the es-syncer script is not able to connect to Redis, no error is shown.

mfournier:
So now we indeed have more logs, but the syncer still can't recover properly from a redis restart.

These are the last logs before I restarted the service this evening (note the timestamp dating back this morning):
```
syncer_1     | {"message": "Last update time: 2016-12-16 09:13:24.841674+00:00", "levelname": "INFO", "name": "c2corg_api.scripts.es.sync", "asctime": "2016-12-16 09:16:05,922", "funcName": "sync_es", "lineno": 26, "pathname": "/var/www/c2corg_api/scripts/es/sync.py", "stack_info": null, "module": "sync", "created": 1481879765.9222815, "threadName": "MainThread", "processName": "MainProcess", "getMessage": null}
syncer_1     | {"message": "Number of changed documents: 3", "levelname": "INFO", "name": "c2corg_api.scripts.es.sync", "asctime": "2016-12-16 09:16:05,941", "funcName": "sync_es", "lineno": 38, "pathname": "/var/www/c2corg_api/scripts/es/sync.py", "stack_info": null, "module": "sync", "created": 1481879765.9418766, "threadName": "MainThread", "processName": "MainProcess", "getMessage": null}
syncer_1     | {"message": "Sent 2 document(s) of type u", "levelname": "INFO", "name": "c2corg_api.scripts.es.sync", "asctime": "2016-12-16 09:16:19,535", "funcName": "create_search_documents", "lineno": 321, "pathname": "/var/www/c2corg_api/scripts/es/sync.py", "stack_info": null, "module": "sync", "created": 1481879779.535723, "threadName": "MainThread", "processName": "MainProcess", "getMessage": null}
syncer_1     | {"message": "Sent 1 document(s) of type i", "levelname": "INFO", "name": "c2corg_api.scripts.es.sync", "asctime": "2016-12-16 09:16:31,071", "funcName": "create_search_documents", "lineno": 321, "pathname": "/var/www/c2corg_api/scripts/es/sync.py", "stack_info": null, "module": "sync", "created": 1481879791.0716262, "threadName": "MainThread", "processName": "MainProcess", "getMessage": null}
[...]

PR #588, opened 2016-12-13
Make syncer more resilient

tsauerwein:
* Be more verbose in the logs
* In case the connection to Redis fails in the syncer, 3 retries are made. After that the syncer script crashes.
* In the API when pushing a message into the Redis queue: So far the Redis queue was created when the API instance was started. When the queue was deleted because of a problem with Redis, the API could no longer write into the queue. Now, the Redis queue is re-created.

Closes https://github.com/c2corg/v6_api/issues/534
Closes https://github.com/c2corg/v6_ui/issues/1169

ping @mfournier

### DIFF
--- a/c2corg_api/scripts/es/sync.py
+++ b/c2corg_api/scripts/es/sync.py
@@ -23,6 +23,7 @@ log = logging.getLogger(__name__)
 
 def sync_es(session, batch_size=1000):
     last_update, date_now = es_sync.get_status(session)
+    log.info('Last update time: {}'.format(last_update))
 
     if not last_update:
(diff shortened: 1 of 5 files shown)


asaunier:
OK for me but I let @mfournier give the final clearance :)

tsauerwein:
Merging for now, so that the changes can be tested.

mfournier:
This looks all good, and matches what we discussed on hipchat the other day, thanks a lot @tsauerwein !

The 3rd point you mention might indeed be the cause of the stalled sync we had last WE.

Issue #582, opened 2016-12-08
Check provided bbox in advanced-search


tsauerwein:
In the logs there are requests like `/outings?act=rock_climbing&bbox=NaN%2CNaN%2CNaN%2CNaN&pl=fr`, where the bbox is not correct. This makes ElasticSearch not happy.

Also related problem:

> elasticsearch.exceptions.RequestError: TransportError(400, 'search_phase_execution_exception', 'top latitude cannot equal bottom latitude: [44.86313139368782] vs [44.86313139368782]')

PR #593, opened 2016-12-19
Check and test NaN in BBOX (adv. search filter)

dkocich:
Close https://github.com/c2corg/v6_api/issues/582

I think this caused the error (in my tests). With this fix, it should be ok.

```
...
File "/home/davidk/git/c2c/v6_api/c2corg_api/tests/search/test_search_filters.py", line 413, in test_create_bbox_filter
    self.assertEqual(create_bbox_filter('NaN,NaN,NaN,NaN'), None)
  File "/usr/lib/python3.5/unittest/case.py", line 820, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib/python3.5/unittest/case.py", line 813, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: GeoBoundingBox(geom={'left': nan, 'right'[45 chars]xed') != None
-------------------- >> begin captured logging << --------------------
sqlalchemy.engine.base.Engine: INFO: BEGIN (implicit)
sqlalchemy.engine.base.Engine: INFO: ROLLBACK
--------------------- >> end captured logging << ---------------------
```

### DIFF
--- a/c2corg_api/search/search_filters.py
+++ b/c2corg_api/search/search_filters.py
@@ -321,6 +321,9 @@ def create_bbox_filter(query_term):
     xmin, ymin = transform(bbox3857[0], bbox3857[1])
     xmax, ymax = transform(bbox3857[2], bbox3857[3])
 
+    if xmin == xmax or ymin == ymax:
+        return None
+
     return GeoBoundingBox(
         geom={'left': xmin, 'bottom': ymin, 'right': xmax, 'top': ymax},
         type='indexed'
@@ -350,7 +353,10 @@ def parse_num(s):
         try:
             return int(s)
         except ValueError:
-            return float(s)
+            val = float(s)
+            if math.isnan(val):
+                return None
+            return val
     except ValueError:
         return None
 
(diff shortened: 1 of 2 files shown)


c2corg_api/search/search_filters.py
```diff
@@ -316,6 +316,9 @@ def create_bbox_filter(query_term):
     bbox3857 = list(map(parse_num, query_terms))
     if len(bbox3857) != 4 or not all(bbox3857):
         return None
+    for coord in bbox3857:
+        if math.isnan(float(coord)):
```
tsauerwein: Would it make sense to handle this check in `parse_num`?

tsauerwein:
Thanks for tackling this! The issue also talks about a 2nd problem:

> 'top latitude cannot equal bottom latitude: [44.86313139368782] vs [44.86313139368782]')

Do you want to handle this also in this PR?

dkocich:
It is added - it should be fine also for lon/lat now.

asaunier:
@dkocich @tsauerwein Shall we merge this PR?

c2corg_api/search/search_filters.py
```diff
@@ -347,6 +350,8 @@ def map_enum_to_int(enum_mapper, s):
 
 def parse_num(s):
     try:
+        if math.isnan(float(s)):
```
tsauerwein: Small optimization: Could you do it so that the string is only parsed once?

PR #597, opened 2017-01-05
Set topic link content for document without title

arnaud-morvan:
Relate https://github.com/c2corg/v6_forum/issues/68

### DIFF
--- a/c2corg_api/views/forum.py
+++ b/c2corg_api/views/forum.py
@@ -1,4 +1,5 @@
 import colander
+from urllib3.util.url import parse_url
 
 from c2corg_api.models import DBSession
 from c2corg_api.models.cache_version import update_cache_version_direct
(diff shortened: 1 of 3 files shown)


c2corg_api/views/forum.py
```diff
@@ -82,7 +82,8 @@ def collection_post(self):
         title = "{}_{}".format(locale.document_id, locale.lang)
         content = '<a href="{}">{}</a>'.format(
                 self.request.referer,
-                locale.title)
+                locale.title or "{}/{}".format(locale.document_id,
```
asaunier: OK for using ``document_id/lang`` as fallback even though it might not be very clear for users to have a link such as ``123456/fr`` :p 

Would it be possible to add the document type as well? For instance ``images/123456/fr`` ?

arnaud-morvan: Yes, I was already planning to do so, after discussion with tobias at lunch.

PR #606, opened 2017-01-17
Add Users to Xreports updatable associations

dkocich:
required later to update associations in Xreports https://github.com/c2corg/v6_ui/pull/1108

### DIFF
--- a/c2corg_api/models/association.py
+++ b/c2corg_api/models/association.py
@@ -381,5 +381,5 @@ updatable_associations = {
                    'outings', 'books', 'xreports'},
     AREA_TYPE: {'images'},
     BOOK_TYPE: {'routes', 'articles', 'images', 'waypoints'},
-    XREPORT_TYPE: {'routes', 'outings', 'articles', 'images'}
+    XREPORT_TYPE: {'routes', 'outings', 'articles', 'users', 'images'}
 }


c2corg_api/models/association.py
```diff
@@ -381,5 +381,6 @@ def _get_associations_to_add(new_associations, current_associations):
                    'outings', 'books', 'xreports'},
     AREA_TYPE: {'images'},
     BOOK_TYPE: {'routes', 'articles', 'images', 'waypoints'},
-    XREPORT_TYPE: {'routes', 'outings', 'articles', 'images'}
+    XREPORT_TYPE: {'routes', 'outings', 'articles', 'images', 'waypoints',
```
asaunier: Do we really need articles and waypoints here?

dkocich: I am not sure about waypoints - it is mentioned also in the issue (https://github.com/c2corg/v6_ui/pull/1108#issue-193748043). Articles can be probably also linked to Xreport from its own page.

mgerbaux: I do not think WP should be interesting (confusing witg I/A geolocalization). If needed, we can do **later** a join with outings or route associated.

Articles associations could be interesting.

Issue #465, opened 2016-11-02
Missing "delete document" tool


asaunier:

asaunier:
Can this issue really wait for 6.1? It means moderators would have no way to neutralize incorrect or inappropriate documents.

tsauerwein:
> It means moderators would have no way to neutralize incorrect or inappropriate documents.

Except doing it by hand in the database.

asaunier:
OK then, but is it really a good option:
* high risk of making damages in the DB (if incorrect or missing commands)
* quite a lot of actions to do: remove the archives and last version (both figures and locales and geometries), remove associations, remove history, invalidate cache, etc.

asaunier:
### Description

In this issue we will assume that we only delete "standard" documents: **waypoints, routes, outings, articles, xreports, books**. Since creating areas or maps is not accessible to the public, we might consider it is not needed to support those types (by the way maps are not "wiki" documents as the others).

**Deleting a user account is needed as well but should probably be considered in a dedicated PR** that would be an evolution of the current one. It's a bit more complicated since both the user account and the profile document should be removed. In addition changes should be done in the history of all edited documents. In v5, user account deleting was in fact done by merging the target account to a special account.

When removing a document, say a route, entries must be deleted in:
* main tables ``routes``, ``documents``, ``routes_locales``, ``documents_locales``, ``documents_geometries`` _[Note: delete entries for all available languages]_
* archives tables ``routes_archives``, ``documents_archives``, ``routes_locales_archives``, ``documents_locales_archives``, ``documents_geometries_archives``
* metadata tables ``documents_versions`` and ``history_metadata``
* associations tables ``associations``, ``association_log``, ``area_associations``, ``map_associations``
* comment table ``documents_topics`` + the comments should be removed in Discourse (for all langs)

Additional tasks:
* invalidate the API and UI caches for the deleted document
* update the ElasticSearch index
* for images: delete the related files in the images backend (that requires to get the list of filenames in the ``images`` and ``images_archives`` before deleting them). See https://github.com/c2corg/v6_images/issues/35

### Constraints

**It might be necessary to check for constraints before removing a document.**
For instance since routes should have at least one associated waypoint, it should not be possible to delete a waypoint if it's the only one associated to some routes. Ditto for outings: deleting a route should not be possible if outings are only associated to this route.

### Implementation

I am at the very beginning of the study but I have noticed Cornice supports DELETE requests. See for instance:
[...]

PR #610, opened 2017-01-20
Document deleting tool

asaunier:
Related to https://github.com/c2corg/v6_api/issues/465

Delete view:
* [x] Add document id to a special table where ES can figure out what doc should be removed from the index
* [x] when removing an image, update the feed entries that reference it
* [x] when removing an image, delete the files from the images storage (see https://github.com/c2corg/v6_images/issues/35 and https://github.com/c2corg/v6_api/blob/merge-documents/c2corg_api/views/image.py#L116-L137)
* [x] remove geoassociations

Unit tests:
* [x] make sure the existing doc/archives test are correct
* [x] check associations have been cleared
* [x] check cache_versions have been updated?
* [x] check the feed has been cleared
* [x] check comments have been cleared?
* [x] check image files have been removed?
* [ ] check ES index is updated => see https://github.com/c2corg/v6_api/issues/629

Add tests for
* [x] articles
* [x] books
* [x] images
* [x] outings
* [x] xreports

### DIFF
--- a/c2corg_api/scripts/es/sync.py
+++ b/c2corg_api/scripts/es/sync.py
@@ -3,6 +3,7 @@ from c2corg_api.models.area import Area
 from c2corg_api.models.association import AssociationLog, Association
 from c2corg_api.models.document import Document, DocumentGeometry
 from c2corg_api.models.document_history import DocumentVersion, HistoryMetaData
+from c2corg_api.models.es_sync import ESDeletedDocument
 from c2corg_api.models.outing import Outing, OUTING_TYPE
 from c2corg_api.models.route import Route, ROUTE_TYPE
 from c2corg_api.models.user import User
@@ -38,6 +39,13 @@ def sync_es(session, batch_size=1000):
     log.info('Number of changed documents: {}'.format(len(changed_documents)))
     if changed_documents:
         sync_documents(session, changed_documents, batch_size)
+
+    # get list of documents deleted since the last update
+    deleted_documents = get_deleted_documents(session, last_update)
+
+    log.info('Number of deleted documents: {}'.format(len(deleted_documents)))
+    if deleted_documents:
+        sync_deleted_documents(session, deleted_documents, batch_size)
 
     es_sync.mark_as_updated(session, date_now)
     log.info('Sync has finished')
@@ -234,6 +242,15 @@ def get_changed_outings_ro_uo(session, last_update):
         all()
 
 
+def get_deleted_documents(session, last_update):
+    """Returns the ids of documents deleted since the last update.
+    """
+    return session. \
+        query(ESDeletedDocument.document_id, ESDeletedDocument.type) . \
+        filter(ESDeletedDocument.deleted_at >= last_update). \
+        all()
+
+
 def sync_documents(session, changed_documents, batch_size):
     client = elasticsearch_config['client']
     batch = ElasticBatch(client, batch_size)
@@ -245,6 +262,24 @@ def sync_documents(session, changed_documents, batch_size):
                 docs = get_documents(
                     session, doc_type, batch_size, document_ids)
                 create_search_documents(doc_type, docs, batch)
+
+
+def sync_deleted_documents(session, deleted_documents, batch_size):
+    client = elasticsearch_config['client']
+    batch = ElasticBatch(client, batch_size)
+    index = elasticsearch_config['index']
+    n = 0
+    with batch:
+        for document_id, doc_type in deleted_documents:
+            batch.add({
+                '_index': index,
+                '_id': document_id,
+                '_type': doc_type,
+                'id': document_id,
+                '_op_type': 'delete'
+            })
+            n += 1
+        log.info('Removed {} document(s)'.format(n))
 
 
 def add_dependent_documents(session, docs_per_type):
--- a/c2corg_api/scripts/es/sync.py
+++ b/c2corg_api/scripts/es/sync.py
@@ -279,7 +279,7 @@ def sync_deleted_documents(session, deleted_documents, batch_size):
                 '_op_type': 'delete'
             })
             n += 1
-        log.info('Removed {} document(s)'.format(n))
+    log.info('Removed {} document(s)'.format(n))
 
 
 def add_dependent_documents(session, docs_per_type):
--- a/c2corg_api/tests/views/test_document_delete.py
+++ b/c2corg_api/tests/views/test_document_delete.py
@@ -1,17 +1,22 @@
 import datetime
 
+from c2corg_api.models.article import Article, ArchiveArticle
 from c2corg_api.models.association import Association
+from c2corg_api.models.book import Book, ArchiveBook
 from c2corg_api.models.document import (
     Document, DocumentLocale, DocumentGeometry, ArchiveDocument,
-    ArchiveDocumentLocale, ArchiveDocumentGeometry)
+    ArchiveDocumentLocale, ArchiveDocumentGeometry, UpdateType)
 from c2corg_api.models.document_history import DocumentVersion
 from c2corg_api.models.feed import update_feed_document_create
+from c2corg_api.models.image import Image, ArchiveImage
 from c2corg_api.models.outing import (
     Outing, OutingLocale, ArchiveOuting, ArchiveOutingLocale)
 from c2corg_api.models.route import (
     Route, RouteLocale, ArchiveRoute, ArchiveRouteLocale)
 from c2corg_api.models.waypoint import (
     Waypoint, WaypointLocale, ArchiveWaypoint, ArchiveWaypointLocale)
+from c2corg_api.models.xreport import (
+    Xreport, XreportLocale, ArchiveXreport, ArchiveXreportLocale)
 from c2corg_api.views.document import DocumentRest
 from c2corg_api.tests.views import BaseTestRest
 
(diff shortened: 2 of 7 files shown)


c2corg_api/views/document.py
```diff
@@ -296,6 +296,35 @@ def _put(
...
+    def _delete(self, clazz, schema, clazz_locale=None):
+        if not self.request.has_permission('moderator'):
+            raise HTTPForbidden('No permission to delete the document')
+        id = self.request.validated['id']
+
+        # get the current version of the document
+        document = self._get_document(clazz, id, clazz_locale=clazz_locale)
+        lang = document.locales[0].lang
+
+        associations = doc_associations.get_associations(document, lang, False)
```
tsauerwein: You are loading all associations to check if a waypoint is still the main waypoint of a route? This could be improved. You could directly query the routes table:

```
def _is_main_waypoint_of_route(waypoint):
    return DBSession.query(
        exists().
        where(Route.main_waypoint_id == waypoint.document_id)
    ).scalar()
```
Similar to https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/association.py#L161

c2corg_api/views/document_associations.py
```diff
@@ -303,3 +304,23 @@ def get_linked_xreports(document, lang):
...
+
+        # check if not a main waypoint
+        if 'all_routes' in associations and associations['all_routes'].total:
+            for route in associations['all_routes'].documents:
+                if route.main_waypoint_id == document.document_id:
+                    raise HTTPBadRequest(
+                        'This waypoint cannot be deleted because it is the '
+                        'main waypoint of route {}'.format(route.document_id))
+
+        # check if not the only waypoint of a route
```
tsauerwein: See also: https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/association.py#L169

c2corg_api/views/document.py
asaunier: OK thanks, but anyway we need to get the list of associations in order to update their ``cache_versions`` when the document has actually been removed, don't we?

tsauerwein: No, this can be done by calling a stored procedure: https://github.com/c2corg/v6_api/blob/494d48d/c2corg_api/models/cache_version.py#L41

c2corg_api/views/document.py
```diff
@@ -300,14 +301,18 @@ def _delete(self, clazz):
...
-
-        # get the current version of the document
         document = self._get_document(clazz, id)
 
         # check associations constraints
         doc_associations.check_before_deleting_document(document)
 
-        # TODO: image => get filenames
+        if document.type == IMAGE_TYPE:
+            filenames = DBSession.query(Image.filename). \
```
tsauerwein: I think you want to run this query on `ImageArchive`? Because there is only one entry in the `Image` table for a document id.

c2corg_api/views/document.py
```diff
@@ -311,40 +311,55 @@ def _delete(
...
         # Note: if an error occurs while deleting, SqlAlchemy will
         # automatically cancel all changes.
 
+        # Remove document from the cache system
+        DBSession.query(CacheVersion). \
+            filter(CacheVersion.document_id == id).delete()
+
+        # Remove associations and update cache of formerly associated docs
+        doc_associations.remove_associations(document)
+        update_cache_version(document)
```
tsauerwein: I guess you have to call `update_cache_version` before removing the associations. Otherwise the cache versions of the associations cannot be updated.

c2corg_api/views/document_delete.py
```diff
@@ -289,6 +290,5 @@ def _remove_associations(document_id):
         )).delete()
 
 
-def _notify_es_syncer(config):
-    # TODO
-    pass
+def _notify_es_syncer(document_id):
+    DBSession.add(ESDeletedDocument(document_id=document_id))
```
tsauerwein: You will also have to send a message to Redis if the database transaction is successful. See the existing function  `notify_es_syncer`.

asaunier:
I have rebased the PR to group and rename some commits to make them a bit more readable.
The first one remains a bit big because I have changed my mind, moving most of the code and tests to dedicated files and I had to group the change to keep only the final approach.

A few items are still not addressed in the TODO:
* testing the cache versions => perhaps only the cache version of the deleted doc? Or also the versions of formerly associated docs?
* testing the "removal" of comments (actually only the "associations" in the guidebook's topic table, comments should be removed separately from Discourse).
* testing ES index has actually been updated. I don't know how to test this yet.

asaunier:
@tsauerwein Do you mind having a look at this PR? Thanks :)

I'm currently working on the UI side. This will also allow to make real data testing.

c2corg_api/views/document_delete.py
```diff
@@ -0,0 +1,327 @@
...
+        _remove_archive_locale(archive_clazz_locale, document_id)
+        _remove_archive_geometry(document_id)
+        _remove_archive(archive_clazz, document_id)
+        _remove_locale(clazz_locale, document_id)
+        _remove_geometry(document_id)
+        _remove_figures(clazz, document_id)
+        _remove_from_feed(document_id)
+
+        if document_type == IMAGE_TYPE:
+            # Remove this image references from the feed
```
tsauerwein: "remove the references of this image ..."?

c2corg_api/views/document_delete.py
```diff
@@ -0,0 +1,327 @@
...
+        # Shift image references in the feed item
+        if item.image1_id == document_id:
+            item.image1_id = item.image2_id
+            item.image2_id = item.image3_id
+        elif item.image2_id == document_id:
+            item.image2_id = item.image3_id
+        item.image3_id = None
+        item.more_images = False
+        # FIXME what happens if the deleted image was the only image reference
+        # in the feed item => should the item be removed or replaced?
```
tsauerwein: If the change type is `'added_photos'` it would probably make sense to remove the entire feed entry.

c2corg_api/views/document_delete.py
```diff
@@ -0,0 +1,327 @@
...
+    if len(items):
+        DBSession.flush()
```
tsauerwein: Is the flush necessary here?

c2corg_api/scripts/es/sync.py
```diff
@@ -247,6 +264,24 @@ def sync_documents(session, changed_documents, batch_size):
...
+        for document_id, doc_type in deleted_documents:
+            batch.add({
+                '_index': index,
+                '_id': document_id,
+                '_type': doc_type,
+                'id': document_id,
+                '_op_type': 'delete'
+            })
+            n += 1
+        log.info('Removed {} document(s)'.format(n))
```
tsauerwein: Maybe move the log statement outside the `with batch:` block. Because the (last) request to ES is only send when the batch is closed. And only then you know if it was successful.

c2corg_api/tests/views/test_document_merge.py
```diff
@@ -261,6 +255,11 @@ def image_service_mock(url, request):
                 'target_document_id': self.image2.document_id
             }, 200)
             self.assertEqual(call['times'], 1)
+            self.assertIn('filenames=image1.1.jpg', call['request.body'])
+            self.assertIn('filenames=image1.jpg', call['request.body'])
```
tsauerwein: Thanks for fixing this test!

c2corg_api/tests/views/test_document_delete.py
```diff
@@ -0,0 +1,562 @@
...
+        self.session.flush()
+
+        self.image1.filename = 'image1.1.jpg'
+        self.session.flush()
+        DocumentRest.update_version(
+            self.image1, user_id,
+            'changed filename', [UpdateType.FIGURES], [])
+        self.session.flush()
+
+        # TODO test deleting SVG images?
```
tsauerwein: Handling this should be in the responsibility of the image service: https://github.com/c2corg/v6_images/blob/master/tests/inside/test_views.py#L27

c2corg_api/tests/views/test_document_delete.py
```diff
@@ -0,0 +1,562 @@
...
+    def test_non_existing_document(self):
+        self._delete(-9999999, 400)
+
+    def test_delete_main_waypoint(self):
+        """ Test that a main waypoint cannot be deleted.
+        """
+        response = self._delete(self.waypoint1.document_id, 400)
+        errors = response.json.get('errors')
+        self.assertEqual(len(errors), 1)
+        self.assertEqual(errors[0].get('name'), 'Bad Request')
```
tsauerwein: There is a convenience method to make testing for errors simpler. For example you could write this test as:

```python
self.assertErrorsContain(response.json, 'Bad Request', 'This waypoint cannot be deleted because it is a main waypoint.')
```

c2corg_api/tests/views/test_document_delete.py
```diff
@@ -0,0 +1,562 @@
...
+        )).count()
+        self.assertEqual(0, association_log_count)
+        association_count = self.session.query(TopoMapAssociation).filter(
+            TopoMapAssociation.document_id == document_id).count()
+        self.assertEqual(0, association_count)
+        association_count = self.session.query(AreaAssociation).filter(
+            AreaAssociation.document_id == document_id).count()
+        self.assertEqual(0, association_count)
+
+        # TODO check cache_versions have been updated
```
tsauerwein: For example when deleting `self.waypoint3` you could test that the cache version of the associated route `self.route3` has changed:

```python
        self.check_cache_version(self.route3.document_id, 2)

```

tsauerwein:
> check ES index is updated

You could do a test similar to this: https://github.com/c2corg/v6_api/blob/master/c2corg_api/tests/scripts/es/test_sync.py

c2corg_api/views/document_delete.py
asaunier: Yes, in feed entries you have references to images, see https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/feed.py#L152-L163

If you delete one of these images and don't update the feed entries that reference it, you'll get broken images/links in the remaining feed cards. I have tried to address this problem with https://github.com/c2corg/v6_api/blob/78e8fdc0445213a283a4eedf136a56afa0054369/c2corg_api/views/document_delete.py#L284-L305

asaunier: Oh, perhaps you meant my sentence was incorrect?

tsauerwein: > Oh, perhaps you meant my sentence was incorrect?

Yes. ;)

c2corg_api/views/document_delete.py
asaunier: I don't know. I wanted to make sure the changes done in the feed items were actually saved. If you think the flush is useless, I will remove it.

c2corg_api/tests/views/test_document_delete.py
asaunier: OK. Then I will remove this part if it does not need to be tested here.

asaunier:
> You could do a test similar to this: https://github.com/c2corg/v6_api/blob/master/c2corg_api/tests/scripts/es/test_sync.py

You mean the full file or adding some test to it?

tsauerwein:
> You mean the full file or adding some test to it?

The tests in this file are for the functions in [sync.py](https://github.com/c2corg/v6_api/blob/master/c2corg_api/scripts/es/sync.py). So you could add tests for the functions you've added to this file.

asaunier:
Argh!
When I try to delete in a real UI instance an outing with archived versions, I get the following error:
```
sqlalchemy.exc.IntegrityError: (psycopg2.IntegrityError) update or delete on table "documents" violates foreign key constraint "documents_archives_redirects_to_fkey" on table "documents_archives"
DETAIL:  Key (document_id)=(734972) is still referenced from table "documents_archives".
 [SQL: 'DELETE FROM guidebook.documents WHERE guidebook.documents.document_id = %(document_id_1)s'] [parameters: {'document_id_1': 734972}]
```

asaunier:
I don't really have time to add tests about ES index syncing when documents are deleted. By the way I don't really know how to simulate deleted docs in https://github.com/c2corg/v6_api/blob/master/c2corg_api/tests/scripts/es/test_sync.py

I have created a dedicated issue for this task https://github.com/c2corg/v6_api/issues/629 and will merge the PR soon if no one objects since @tsauerwein has approved it and since his last comments have been addressed.

asaunier:
I have also noticed that the image files are not actually deleted. See https://github.com/c2corg/v6_api/issues/630

Issue #596, opened 2017-01-01
Associated images sorting


asaunier:
As for now, it seems that the images listed in the a document's ``associations`` attribute are id-sorted. It would make more sense to **have them chronologically sorted using the EXIF date** info if available.

This is especially useful for outings to make sure associated images are shown in the order they have been taken. Which is not the case currently if several people upload there images or if a contributor uploads their image in whatever order (for instance when an upload has failed).

dkocich:
Related to UI issue https://github.com/c2corg/v6_ui/issues/1177

dkocich:
Note - Currently, there are around 150 images in DB which have bad `date_time` set (it is between years 4500 and 2017)...

desnoes:
fixed on demo

PR #612, opened 2017-01-24
Image sorting by date_time

dkocich:
Related to https://github.com/c2corg/v6_api/issues/596

Previously while browsing images ( https://www.camptocamp.org/images#offset=120 ) they were not sorted and while I was going through pages some of them had were newer/older date_time than others (some didn't show any date on detail view page).

I am not sure about sorting of linked images on View pages because I am not sure how exactly JS for image slider works and if it works correctly (also with latest changes related `date_time`).
It looks correctly ordered now when I click on the image and info icon.

Another thing to add and discuss
- for Outings/Xreport, it would be probably good to see images from the beginning to the end of a trip so the users can go through as time was passing.
- for Waypoint/Route it should be probably sorted by latest images first (opposite). Users probably want to look at WP to see current conditions rather than some historical photos.
- article?

### DIFF
--- a/c2corg_api/search/search_filters.py
+++ b/c2corg_api/search/search_filters.py
@@ -2,8 +2,6 @@ import math
 
 import pyproj
 import re
-
-from c2corg_api.models.image import IMAGE_TYPE
 from c2corg_api.models.outing import OUTING_TYPE
 from c2corg_api.models.xreport import XREPORT_TYPE
 from c2corg_api.search.mapping_types import reserved_query_fields
@@ -50,9 +48,6 @@ def build_query(url_params, meta_params, doc_type):
         if doc_type == OUTING_TYPE:
             search = search.sort(
                 {'date_end': {'order': 'desc'}}, {'id': {'order': 'desc'}})
-        elif doc_type == IMAGE_TYPE:
-            search = search.sort(
-                {'date_time': {'order': 'desc'}}, {'id': {'order': 'desc'}})
         elif doc_type == XREPORT_TYPE:
             search = search.sort(
               {'date': {'order': 'desc'}}, {'id': {'order': 'desc'}})
--- a/c2corg_api/search/search_filters.py
+++ b/c2corg_api/search/search_filters.py
@@ -2,6 +2,8 @@ import math
 
 import pyproj
 import re
+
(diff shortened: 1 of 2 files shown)


c2corg_api/search/search_filters.py
```diff
@@ -48,6 +50,9 @@ def build_query(url_params, meta_params, doc_type):
         if doc_type == OUTING_TYPE:
             search = search.sort(
                 {'date_end': {'order': 'desc'}}, {'id': {'order': 'desc'}})
+        elif doc_type == IMAGE_TYPE:
+            search = search.sort(
+                {'date_time': {'order': 'desc'}}, {'id': {'order': 'desc'}})
```
asaunier: desc or asc?
The oldest first because it makes more sense to show the images in the order they have been taken.

c2corg_api/views/document_associations.py
```diff
@@ -180,6 +180,7 @@ def get_linked_images(document, lang):
                 Association.child_document_id == Image.document_id,
                 Association.parent_document_id == document.document_id)
             ).
+        order_by(desc(Image.date_time)).
```
asaunier: ``asc`` I think.

asaunier:
Please note the image date sorting is needed only when listing the images associated to a document (eg. outings). Not in the advance search page => still sort on desc id here.

c2corg_api/views/document_associations.py
dkocich: This is where it can be set depending on the doc. type as I said in the first message.

dkocich:
>  image date sorting is needed only when listing the images associated to a document

ok

asaunier:
Sorry, I had not read your comments carefully enough!
I thought you were addressing issue https://github.com/c2corg/v6_api/issues/596 (I have added a link to your initial comment).

> Another thing to add and discuss

For outings and xreports, images should be sorted chronologically.
For other documents, I don't know. It could indeed make sense to show the newest first but using a chronological sorting is probably easier and more consistent!?

dkocich:
This PR was updated to sort all linked images chronologically on detail view pages. Maybe we can merge now and see how it looks on demo and if it is really better to sort images based on the doctype?

Issue #580, opened 2016-12-08
Validate geometries before storing


tsauerwein:
There are a couple of errors in the logs:
```
psycopg2.InternalError: LineString must have at least two points CONTEXT: SQL function "st_geomfromwkb" statement 1
```

We should make sure that the geometries are checked earlier.

dkocich:
Should we validate using [PostGIS](http://postgis.net/docs/ST_IsValid.html) DB or directly in python with [shapely](http://toblerity.org/shapely/manual.html#object.is_valid) / fiona ? Shapely is probably better.

asaunier:
An example of GPX file that makes the API crash is available at https://github.com/c2corg/v6_ui/issues/1143#issuecomment-265714053

I copy it here:
[2016-12-02.Pic.Blanc.du.Galibier.zip](https://github.com/c2corg/v6_api/files/701849/2016-12-02.Pic.Blanc.du.Galibier.zip)
(unzip before uploading it)

See my additional info at https://github.com/c2corg/v6_ui/issues/1143#issuecomment-272169192

UI sends the following ``geometry`` object:
```
geom: "{"type":"Point","coordinates":[714602.5181561863,5628074.236953052]}"
[...]

PR #613, opened 2017-01-25
Validate geometry type from geojson for each doc. type on API

dkocich:
related to https://github.com/c2corg/v6_api/issues/580

TODO
- As we discussed, it is probably better to validate geometry by geometry type stored in GeoJson before it is converted from GeoJson to WKB. Problem is that we are missing doctype in there https://github.com/c2corg/v6_api/blob/master/c2corg_api/ext/colander_ext.py#L47-L52

Right now this solution validates 2d geometries using Shapely and higher dimensions using separate DB query. If there is no geometry or geometry is taken from associated documents - nothing is validated.

Feel free to adjust...

**Possible problems...?**
- I am not sure about validation of associated geometries (now they are skipped). (example test https://github.com/c2corg/v6_api/blob/a4edbc11b90ca20053ae84ad944d7b0127b823f7/c2corg_api/tests/views/test_route.py#L870 )
- sometimes WPs have `geom` + `geom_detail` (LS, MLS) and now I validate only geom, which should contain `Point` goemetry

### DIFF
--- a/c2corg_api/ext/colander_ext.py
+++ b/c2corg_api/ext/colander_ext.py
@@ -27,8 +27,8 @@ class Geometry(SchemaType):
     srid
         The SRID of the geometry should also match the column definition.
     """
-    def __init__(self, geometry_type='GEOMETRY', srid=-1, map_srid=-1):
-        self.geometry_type = geometry_type.upper()
+    def __init__(self, geometry_types=['GEOMETRY'], srid=-1, map_srid=-1):
+        self.geometry_types = [t.upper() for t in geometry_types]
         self.srid = int(srid)
 
     def serialize(self, node, appstruct):
@@ -53,16 +53,26 @@ class Geometry(SchemaType):
         if cstruct is null or cstruct == '':
             return null
         try:
-            return wkbelement_from_geojson(cstruct, self.srid)
-        except Exception:
-            raise Invalid(node, 'Invalid geometry: %r' % cstruct)
+            geojson = json.loads(cstruct)
+            geom_type = geojson['type'].upper()
+            allowed_types = self.geometry_types
+            if geom_type in allowed_types:
+                return wkbelement_from_geojson(geojson, self.srid)
+            else:
+                raise Invalid(
+                    node, 'Invalid geometry type. Expected: %s. Got: %s.'
+                    % (allowed_types, geom_type))
+        except Invalid as exc:
+            raise exc
+        except:
+            raise Invalid(node, 'Invalid geometry: %s' % cstruct)
 
     def cstruct_children(self, node, cstruct):
         return []
 
 
 def wkbelement_from_geojson(geojson, srid):
-    geometry = wkb.dumps(json.loads(geojson), big_endian=False)
+    geometry = wkb.dumps(geojson, big_endian=False)
     return from_wkb(geometry, srid)
 
 
--- a/c2corg_api/tests/views/test_waypoint.py
+++ b/c2corg_api/tests/views/test_waypoint.py
@@ -660,6 +660,35 @@ class TestWaypointRest(BaseDocumentTestRest):
             feed_change.area_ids, [self.area1.document_id]
         )
 
+    def test_post_wrong_geom_type(self):
+        body = {
+            'document_id': 1234,
+            'version': 2345,
+            'geometry': {
+                'document_id': 5678, 'version': 6789,
+                'geom': '{"type": "LineString", "coordinates": '
+                        '[[635956, 5723604], [635960, 5723610]]}',
+                'geom_detail':
+                    '{"type": "Point", "coordinates": [635956, 5723604]}'
+            },
+            'waypoint_type': 'summit',
+            'elevation': 3779,
+            'locales': [{
+                'id': 3456, 'version': 4567,
+                'lang': 'en', 'title': 'Mont Pourri',
+                'access': 'y'}
+            ],
+            'associations': {
+                'waypoint_children': [
+                    {'document_id': self.waypoint2.document_id}
+                ]
+            }
+        }
+        errors = self.post_wrong_geom_type(body)
+        self.assertEqual(
+            errors[0]['description'], "Invalid geometry type. Expected: "
+            "['POINT']. Got: LINESTRING.")
+
     def test_put_wrong_document_id(self):
         body = {
             'document': {
--- a/c2corg_api/models/area.py
+++ b/c2corg_api/models/area.py
@@ -1,6 +1,6 @@
 from c2corg_api.models import schema, Base
 from c2corg_api.models.document import (
-    ArchiveDocument, Document, geometry_schema_overrides,
+    ArchiveDocument, Document, get_geometry_schema_overrides,
     schema_document_locale, schema_attributes)
 from c2corg_api.models.enums import area_type
 from c2corg_api.models.schema_utils import restrict_schema, \
@@ -81,7 +81,7 @@ schema_area = SQLAlchemySchemaNode(
         'locales': {
             'children': [schema_document_locale]
         },
-        'geometry': geometry_schema_overrides
+        'geometry': get_geometry_schema_overrides(['POLYGON', 'MULTIPOLYGON'])
     })
 
 schema_create_area = get_create_schema(schema_area)
--- a/c2corg_api/models/document.py
+++ b/c2corg_api/models/document.py
@@ -299,7 +299,7 @@ class _DocumentGeometryMixin(object):
                 spatial_index=self.__name__ != 'ArchiveDocumentGeometry'),
             info={
                 'colanderalchemy': {
-                    'typ': colander_ext.Geometry('POINT', srid=3857)
+                    'typ': colander_ext.Geometry(['POINT'], srid=3857)
                 }
             }
         )
@@ -314,7 +314,7 @@ class _DocumentGeometryMixin(object):
                 spatial_index=self.__name__ != 'ArchiveDocumentGeometry'),
             info={
                 'colanderalchemy': {
-                    'typ': colander_ext.Geometry('GEOMETRY', srid=3857)
+                    'typ': colander_ext.Geometry(['GEOMETRY'], srid=3857)
                 }
             }
         )
@@ -449,6 +449,21 @@ geometry_schema_overrides = {
         }
     }
 }
+
+
+def get_geometry_schema_overrides(geometry_types):
+    return {
+        # whitelisted attributes
+        'includes': ['version', 'geom', 'geom_detail'],
+        'overrides': {
+            'version': {
+                'missing': None
+            },
+            'geom_detail': {
+                'typ': colander_ext.Geometry(geometry_types, srid=3857)
+            }
+        }
+    }
 
 
 def get_available_langs(document_id):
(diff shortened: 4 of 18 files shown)


c2corg_api/views/document.py
```diff
@@ -493,6 +506,53 @@ def _check_versions(self, document, document_in):
...
+    """
+    doc_type = document.type
+
+    if doc_type in [OUTING_TYPE, ROUTE_TYPE, AREA_TYPE, MAP_TYPE]:
+        g_field = str(document.geometry.geom_detail)
+    elif doc_type in [IMAGE_TYPE, USERPROFILE_TYPE,
+                      XREPORT_TYPE, WAYPOINT_TYPE]:
+        g_field = str(document.geometry.geom)
+
+    if g_field == 'None':
```
asaunier: Do you mean ``g_field is None``?

c2corg_api/views/document.py
```diff
@@ -493,6 +506,53 @@ def _check_versions(self, document, document_in):
...
+            result = DBSession.execute(q)
+            geom_type = result.fetchone()[0]
+        except:
+            log.warning('Error with geometry %s %s',
+                        doc_type, g_field, exc_info=True)
+            return False
+
+    if doc_type in [IMAGE_TYPE, USERPROFILE_TYPE, XREPORT_TYPE, WAYPOINT_TYPE]:
+        if geom_type.upper() in ['POINT']:
+            return True
```
asaunier: A shorter syntax would be for instance:
```
geom_type = geom_type.upper()
if doc_type in [IMAGE_TYPE, USERPROFILE_TYPE, XREPORT_TYPE, WAYPOINT_TYPE]:
  return geom_type == 'POINT'
elif doc_type in [OUTING_TYPE, ROUTE_TYPE]:
  return geom_type in ['LINESTRING', 'MULTILINESTRING']
elif doc_type in [AREA_TYPE, MAP_TYPE]:
  return geom_type == 'POLYGON'
return False
```

By the way do you really need to check for ``MAP_TYPE``? In UI they are not even a document type and it's not possible to edit them. Is it possible in the API?

c2corg_api/views/document.py
```diff
@@ -229,6 +237,11 @@ def _put(
         # get the current version of the document
         document = self._get_document(clazz, id, clazz_locale=clazz_locale)
 
+        if document.geometry is not None:
```
asaunier: What about simply ``if document.geometry``?

c2corg_api/views/document.py
```diff
@@ -493,6 +506,53 @@ def _check_versions(self, document, document_in):
...
+        g_field = str(document.geometry.geom_detail)
+    elif doc_type in [IMAGE_TYPE, USERPROFILE_TYPE,
+                      XREPORT_TYPE, WAYPOINT_TYPE]:
+        g_field = str(document.geometry.geom)
+
+    if g_field == 'None':
+        return True
+
+    try:
+        s = wkb.loads(g_field, hex=True)
```
asaunier: What is ``s``? Perhaps use a more explicit variable name?

c2corg_api/views/document.py
```diff
@@ -493,6 +506,53 @@ def _check_versions(self, document, document_in):
...
+            return False
+
+    if doc_type in [IMAGE_TYPE, USERPROFILE_TYPE, XREPORT_TYPE, WAYPOINT_TYPE]:
+        if geom_type.upper() in ['POINT']:
+            return True
+    elif doc_type in [OUTING_TYPE, ROUTE_TYPE]:
+        if geom_type.upper() in ['LINESTRING', 'MULTILINESTRING']:
+            return True
+    elif doc_type in [AREA_TYPE, MAP_TYPE]:
+        if geom_type.upper() in ['POLYGON']:
```
tsauerwein: I think it can also be `MultiPolygon`?

c2corg_api/views/document.py
```diff
@@ -493,6 +506,53 @@ def _check_versions(self, document, document_in):
...
+    if g_field == 'None':
+        return True
+
+    try:
+        s = wkb.loads(g_field, hex=True)
+        geom_type = s.type
+    except Exception as e:
+        logging.exception(e)
+        try:
+            q = "SELECT GeometryType(ST_GeomFromEWKT('" + g_field + "'));"
```
tsauerwein: Instead of making a request to the database, couldn't you parse the WKB to GeoJSON and then check the `type` property? See https://github.com/c2corg/v6_api/blob/master/c2corg_api/ext/colander_ext.py#L69

And also: http://norvig.com/21-days.html#answers

tsauerwein:
> As we discussed, it is probably better to validate geometry by geometry type stored in GeoJson before it is converted from GeoJson to WKB. Problem is that we are missing doctype in there https://github.com/c2corg/v6_api/blob/master/c2corg_api/ext/colander_ext.py#L47-L52

Please check https://github.com/c2corg/v6_api/commit/a12c4a651a5908dfab6a6a6a6c573a0d172e8d12
This demonstrates an idea how you could set valid geometry types per document type.

c2corg_api/views/document.py
dkocich: Yes I could and I fixed it :)

dkocich:
@tsauerwein I rebased on your branch and added some changes in the two last commits. Maybe you can quickly check? Thanks.

c2corg_api/ext/colander_ext.py
```diff
@@ -53,7 +53,10 @@ def deserialize(self, node, cstruct):
         if cstruct is null or cstruct == '':
             return null
         try:
-            return wkbelement_from_geojson(cstruct, self.srid)
+            geom_type = json.loads(cstruct)['type'].upper()
+            allowed_types = node.typ.geometry_types
```
tsauerwein: `self.geometry_types` does not contain the overwritten types?

c2corg_api/ext/colander_ext.py
```diff
@@ -53,7 +53,10 @@ def deserialize(self, node, cstruct):
         if cstruct is null or cstruct == '':
             return null
         try:
-            return wkbelement_from_geojson(cstruct, self.srid)
+            geom_type = json.loads(cstruct)['type'].upper()
```
tsauerwein: This way you are calling `json.loads` two times on `cstruct`: Here and in `wkbelement_from_geojson`. Could you do it so that it is called only once?

c2corg_api/ext/colander_ext.py
```diff
@@ -53,7 +53,10 @@ def deserialize(self, node, cstruct):
...
+            allowed_types = node.typ.geometry_types
+            if geom_type in allowed_types:
```
tsauerwein: Could you raise an `Invalid` exception if it is not a valid type? Something like "Expected: ... Got: ...".

c2corg_api/views/document.py
```diff
@@ -186,6 +191,11 @@ def _create_document(
         document = schema.objectify(document_in)
         document.document_id = None
 
+        if document.geometry:
```
tsauerwein: Is this check still needed here?

c2corg_api/views/document.py
```diff
@@ -229,6 +239,11 @@ def _put(
         # get the current version of the document
         document = self._get_document(clazz, id, clazz_locale=clazz_locale)
 
+        if document.geometry:
```
tsauerwein: Is this check still needed here?

c2corg_api/ext/colander_ext.py
dkocich: it is adjusted to load JSON only once but I had to adjust few other tests because I moved json loading function out of `wkbelement_from_geojson`.

c2corg_api/ext/colander_ext.py
```diff
@@ -53,16 +53,24 @@ def deserialize(self, node, cstruct):
...
+            geojson = json.loads(cstruct)
+            geom_type = geojson['type'].upper()
+            allowed_types = self.geometry_types
+            if geom_type in allowed_types:
+                return wkbelement_from_geojson(geojson, self.srid)
+            else:
+                raise Invalid(node)
+        except:
+            raise Invalid(
+                node, 'Invalid geometry with expected type: %s. Got %s'
```
tsauerwein: There are two error cases: The geometry is invalid because of a wrong type or because of a malformed GeoJSON (`wkbelement_from_geojson` fails). Right now, you get the same error message for the two cases. For example, you would get the error message "Invalid geometry with expected type: [POINT]. Got POINT" when sending `{"type": "POINT", some invalid geojson}`. This could be improved. For example:

```python
        try:
            geojson = json.loads(cstruct)
            geom_type = geojson['type'].upper()
            allowed_types = self.geometry_types
            if geom_type in allowed_types:
                return wkbelement_from_geojson(geojson, self.srid)
            else:
                raise Invalid(
                    node, 'Invalid geometry type. Expected: %s. Got: %s.'
                    % (allowed_types, geom_type))
        except:
              raise Invalid(node, 'Invalid geometry: %r' % cstruct)
```

c2corg_api/tests/views/test_waypoint.py
```diff
@@ -660,6 +660,34 @@ def test_post_success(self):
...
+            'associations': {
+                'waypoint_children': [
+                    {'document_id': self.waypoint2.document_id}
+                ]
+            }
+        }
+        errors = self.post_wrong_geom_type(body)
+        self.assertEqual(
+            errors[0]['description'], "Invalid geometry with expected type: "
+            "['POINT']. Got MULTIPOINT")
```
tsauerwein: Is this test correct? The type of `geom_detail` is `Point` but the error message says "Got MULTIPOINT"?

c2corg_api/views/document.py
```diff
@@ -186,6 +191,10 @@ def _create_document(
         document = schema.objectify(document_in)
         document.document_id = None
 
+        if not validate_geometry_type(document):
```
tsauerwein: Is this check still needed here?

c2corg_api/views/document.py
```diff
@@ -229,6 +238,10 @@ def _put(
         # get the current version of the document
         document = self._get_document(clazz, id, clazz_locale=clazz_locale)
 
+        if not validate_geometry_type(document):
```
tsauerwein: Same question: Why is the check still needed?

c2corg_api/views/document.py
```diff
@@ -493,6 +506,55 @@ def _check_versions(self, document, document_in):
                 raise HTTPConflict('version of geometry has changed')
 
 
+def validate_geometry_type(document):
```
tsauerwein: Is this function still needed?

c2corg_api/views/document.py
dkocich: there are many cases of `None` geometry in our tests which were not passing so it is better to do have if statement somewhere - I moved it into the `validate_geometry_type` function so it isn't duplicated at least.

tsauerwein:
Thanks for adding the tests! :+1:

c2corg_api/views/document.py
tsauerwein: My question was more about `validate_geometry_type`. Why do you still need to call this function?

c2corg_api/views/document.py
dkocich: no :) it is deleted now

dkocich:
> Thanks for adding the tests! :+1:

I also added one more (`test_post_corrupted_geojson_geom`) in `test_route.py` to test second exception case

tsauerwein:
Very nice! Thanks for your work, this is now looking very good.

Before merging, you could maybe squash/restructure the commits a bit. For example there could be a commit "Validate geometry type in deserialize" and "Add and adapt tests for geometry validation".

dkocich:
Thanks for reviews and help. It is squashed in 2 commits now.

tsauerwein:
Please feel free to merge.

Issue #386, opened 2016-09-06
Admin tools: merging documents


asaunier:
TODO
- [x] set `redirects_to` from merged documents to the target document id
- [x] transfer associations from source to target document
- [x] remove the merged document from the feed?

tsauerwein:
Closing this ticket, the UI part is tracked in https://github.com/c2corg/v6_ui/issues/1322

PR #619, opened 2017-02-01
Support merging documents

tsauerwein:
Still to do:

* [x] Delete files for images (requires https://github.com/c2corg/v6_images/issues/35)
* [x] Make sure no new associations can be created with redirected documents
* [x] Do not try to transfer an association if the association already exists with the target document

Not handled: Merging user accounts.

Related to https://github.com/c2corg/v6_api/issues/386

### DIFF
--- a/c2corg_api/models/cache_version.py
+++ b/c2corg_api/models/cache_version.py
@@ -39,9 +39,15 @@ class CacheVersion(Base):
 
 
 def update_cache_version(document):
+    update_cache_version_full(document.document_id, document.type)
+
+
+def update_cache_version_full(document_id, type):
+    """ Update the cache version of the given document + associated documents.
+    """
     DBSession.execute(
         text('SELECT guidebook.update_cache_version(:document_id, :type)'),
-        {'document_id': document.document_id, 'type': document.type}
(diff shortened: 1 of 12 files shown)


asaunier:
Comments should probably be managed as well when a document is merged. Either by moving the comments to the target doc, or by removing them (needed when deleting a document as well).

tsauerwein:
> Comments should probably be managed as well when a document is merged.

It would be possible to make a request to the Discourse API to move comments from one topic to another topic.

I am wondering what we should do if the source and target document do not have the same locales. E.g. the source document has a translation in English and German, and the target document translations in French and Spanish. How would you transfer the comments?

Maybe it's a rare case, and we could simply warn the moderator if it occurs. Otherwise we could let the moderators manually take care of merging the comments. What do you think?

tsauerwein:
Besides the open question on how to handle comments, this PR would be ready to review.

fjacon:
@tsauerwein  for now (V6.1), we do not take care of comments : we keep only comments from the target document.

tsauerwein:
Thanks for the review!

Issue #605, opened 2017-01-17
Service listing all recent changes


asaunier:
See the UI issue at https://github.com/c2corg/v6_ui/issues/1323

tsauerwein:
What data would be required?

| Time | User | Type | Titel | Changed Locales |
| ------- | ----- |------- | ------ | --------------------- |
| 21/12/2016 00:35:04 | [Frédéric Bunoz](https://www.camptocamp.org/profiles/288) | waypoint | [Mont Blanc](https://www.camptocamp.org/waypoints/37355/fr/mont-blanc) | [fr](https://www.camptocamp.org/waypoints/history/37355/fr) - [en](https://www.camptocamp.org/waypoints/history/37355/en) |

For a table like this, the following data would be needed:

```
{
  "changes": [{
      "user": {
        "user_id": "...",
        "name": "...",
        ...
      },
      "comment": "...",
      "written_at": "...",
      "document": {
        "document_id": "...",
        "title": "...",
        "title_prefix": "...",
        "lang": "..."
      },
      "changed_langs": ["en", "fr"]
    }, ...
  ],
  pagination_token": "..."
}
```

asaunier:
@stef74 What do you think?

celinestern:
In V5 (http://www.pre-prod.dev.camptocamp.org/routes/whatsnew) we had:
Nom	Date	Auteur	Nature de la révision	Commentaire

Same data in V6 should be ok

stef74:
@asaunier ok with your proposal (as in v5), confirmed by @celinestern

stef74:
@fjacon why extra ?

PR #620, opened 2017-02-02
Browse latest changes on the API

dkocich:
Related to https://github.com/c2corg/v6_api/issues/605

Right now I use http://localhost:6548/document/changes to access this data which are used to render a table of latest changes...

~~TODO - I should add:~~
~~- pagination to load batches of data on UI~~
~~- maybe cache, authentication for moderators or possibility to search by document type.~~

```javascript
{
    documents: [{
        user: {
            username: "dkocich",
            lang: "en",
            user_id: 811849,
            name: "dkocich",
            forum_username: "dkocich"
        },
        written_at: "2017-01-24T15:00:57.852404+01:00",
        version_id: 1302526,
        lang: "en",
        document: {
            document_id: 826224,
            type: "x",
            description: null,
            summary: null,
            title: "asfasfasfas"
        },
        comment: "creation"
    },
    ...
    ],
    count: 500
}
```

### DIFF
--- a/c2corg_api/tests/views/test_document_changes.py
+++ b/c2corg_api/tests/views/test_document_changes.py
@@ -1,0 +1,140 @@
+from c2corg_api.models.document import DocumentGeometry
+from c2corg_api.models.document_history import DocumentVersion, HistoryMetaData
+from c2corg_api.models.route import Route, RouteLocale
+from c2corg_api.models.waypoint import Waypoint, WaypointLocale
+from c2corg_api.tests.views import BaseTestRest
+from c2corg_api.tests.views.test_feed import get_document_ids
+from c2corg_api.views.document import DocumentRest
+
+
+class TestChangesDocumentRest(BaseTestRest):
+
+    def setUp(self):
+        super(TestChangesDocumentRest, self).setUp()
+        self._prefix = '/documents/changes'
+
+        contributor_id = self.global_userids['contributor']
+
+        self.waypoint1 = Waypoint(
+            waypoint_type='summit', elevation=2000,
+            geometry=DocumentGeometry(
+                geom='SRID=3857;POINT(635956 5723604)'),
+            locales=[
+                WaypointLocale(
+                    lang='fr', title='Dent de Crolles',
+                    description='...',
+                    summary='La Dent de Crolles')
+            ])
+        self.session.add(self.waypoint1)
+        self.session.flush()
+        DocumentRest.create_new_version(self.waypoint1, contributor_id)
+        self.session.flush()
+
+        self.waypoint2 = Waypoint(
+            waypoint_type='summit', elevation=4985,
+            geometry=DocumentGeometry(
+                geom='SRID=3857;POINT(635956 5723604)'),
+            locales=[
+                WaypointLocale(
+                    lang='en', title='Mont Blanc',
+                    description='...',
+                    summary='The heighest point in Europe')
+            ])
+        self.session.add(self.waypoint2)
+        self.session.flush()
+        DocumentRest.create_new_version(self.waypoint2, contributor_id)
+        self.session.flush()
+
+        self.waypoint3 = Waypoint(
+            waypoint_type='summit', elevation=4985,
+            redirects_to=self.waypoint1.document_id,
+            geometry=DocumentGeometry(
+                geom='SRID=3857;POINT(635956 5723604)'),
+            locales=[
+                WaypointLocale(
+                    lang='en', title='Mont Blanc',
+                    description='...',
+                    summary='The heighest point in Europe')
+            ])
+        self.session.add(self.waypoint3)
+        self.session.flush()
+        DocumentRest.create_new_version(self.waypoint3, contributor_id)
+        self.session.flush()
+
+        self.route1 = Route(
+            activities=['skitouring'], elevation_max=1500, elevation_min=700,
+            main_waypoint_id=self.waypoint1.document_id,
+            locales=[
+                RouteLocale(
+                    lang='fr', title='Mont Blanc du ciel',
+                    description='...', summary='Ski')
+            ])
+        self.session.add(self.route1)
+        self.session.flush()
+        DocumentRest.create_new_version(self.route1, contributor_id)
+        self.session.flush()
+
+        version_count = self.session.query(DocumentVersion).count()
+        self.assertEqual(4, version_count)
+
+        hist_meta_count = self.session.query(HistoryMetaData).count()
+        self.assertEqual(4, hist_meta_count)
+
+    def test_get_changes(self):
+        response = self.app.get(self._prefix, status=200)
+        body = response.json
+
+        self.assertIn('total', body)
+        self.assertIn('pagination_token', body)
+        self.assertIn('feed', body)
+
+        feed = body['feed']
+        self.assertEqual(4, len(feed))
+        self.assertEqual(4, body['total'])
+
+        # check that the change for the route (latest change) is listed first
+        latest_change = feed[0]
+
+        self.assertEqual(
+            self.route1.document_id, latest_change['document']['document_id'])
+
+    def test_get_changes_paginated(self):
+        response = self.app.get(
+            self._prefix + '?limit=2', status=200)
+        body = response.json
+
+        document_ids = get_document_ids(body)
+        self.assertEqual(2, len(document_ids))
+        self.assertEqual(document_ids, [self.route1.document_id,
+                                        self.waypoint3.document_id])
+        pagination_token = body['pagination_token']
+
+        # last 2 changes
+        response = self.app.get(
+            self._prefix + '?limit=2&token=' + pagination_token, status=200)
+        body = response.json
+
+        document_ids = get_document_ids(body)
+        self.assertEqual(2, len(document_ids))
+        self.assertEqual(
+            document_ids,
+            [self.waypoint2.document_id, self.waypoint1.document_id])
+        pagination_token = body['pagination_token']
+
+        # empty response
+        response = self.app.get(
+            self._prefix + '?limit=2&token=' + pagination_token, status=200)
+        body = response.json
+
+        feed = body['feed']
+        self.assertEqual(0, len(feed))
+
+    def test_get_changes_pagination_invalid_format(self):
+        response = self.app.get(
+            self._prefix + '?token=invalid-token', status=400)
+        self.assertError(response.json['errors'], 'token', 'invalid format')
+
+    def test_get_changes_userid_invalid_format(self):
+        response = self.app.get(
+            self._prefix + '?u=invalid-user_id', status=400)
+        self.assertError(response.json['errors'], 'u', 'invalid u')
--- a/c2corg_api/views/document_changes.py
+++ b/c2corg_api/views/document_changes.py
@@ -1,0 +1,141 @@
+from c2corg_api.models import DBSession
+from c2corg_api.models.document import Document, ArchiveDocumentLocale
+from c2corg_api.models.document_history import DocumentVersion, HistoryMetaData
+from c2corg_api.models.user import User
+from c2corg_api.views import cors_policy
+from c2corg_api.views.feed import get_params
+from c2corg_api.views.validation import validate_simple_token_pagination,\
+  validate_user_id_not_required
+
+from sqlalchemy.sql.expression import desc
+
+from cornice.resource import resource, view
+from sqlalchemy.orm import joinedload, load_only
+
+
+@resource(path='/documents/changes', cors_policy=cors_policy)
+class ChangesDocumentRest(object):
+    """ Unique class for returning history of changes in documents.
+    """
+
+    def __init__(self, request):
+        self.request = request
+
+    @view(validators=[validate_simple_token_pagination,
+                      validate_user_id_not_required])
+    def get(self):
+        """Get the public document changes feed.
+
+        Request:
+            `GET` `/document/changes[&limit=...][&token=...][&u=]`
+
+        Parameters:
+
+            `limit=...` (optional)
+            How many entries should be returned (default: 30).
+            The maximum is 50.
+
+            `token=...` (optional)
+            The pagination token. When requesting a feed, the response includes
+            a `pagination_token`. This token is to be used to request the next
+            page.
+
+            `u=...` (optional)
+            Changes made by one user.
+
+            For more information about "continuation token pagination", see:
+            http://www.servicedenuages.fr/pagination-continuation-token (fr)
+        """
+        user_id = self.request.validated.get('u')
+        lang, token_id, _, limit = get_params(self.request, 30)
+
+        changes = get_changes_of_feed(token_id, limit, user_id)
+
+        doc_ids = []
+        for change in changes:
+            doc_ids.append(change.id)
+
+        return load_feed(doc_ids, limit, user_id)
+
+
+def get_changes_of_feed(token_id, limit, user_id=None):
+    query = DBSession. \
+        query(HistoryMetaData.id, HistoryMetaData.user_id). \
+        order_by(HistoryMetaData.id.desc())
+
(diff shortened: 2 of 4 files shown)


asaunier:
It is not ordered by the association (then no need to implement it now) but I thought of a useful (?) evolution: would it be possible to have an optional parameter to get the changes made by a given user?

dkocich:
@asaunier it should work... http://localhost:6548/document/changes?u=113594 returns changes of one user

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,154 @@
...
+def serialize_change(change):
+    return {
+        'written_at': change.history_metadata.written_at.isoformat(),
+        'lang': change.lang,
+        'document': {
+            'document_id': change.document.document_id,
+            'type': change.document.type,
+            'title': change.document_locales_archive.title,
+            'summary': change.document_locales_archive.summary,
+            'description': change.document_locales_archive.description
```
asaunier: I don't think we need the description (it can be very big, see for instance a route description) and probably not the summary either (especially if it has not changed from the previous version).

Here is the list of data requested by the client (see https://github.com/c2corg/v6_api/issues/605#issuecomment-275209675):
* title
* data
* author
* type of change (creation/update)
* comment about the change
Perhaps the quality would make sense as well.

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+
+@resource(path='/document/changes', cors_policy=cors_policy)
+class ChangesDocumentRest(object):
+    """ Unique class for returning history of a changes in documents.
+    """
+
+    def __init__(self, request):
+        self.request = request
+
+    # validate_pagination, validate_token_pagination, validate_user_id
```
tsauerwein: ?

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @view(validators=[validate_token_pagination,
+                      validate_user_id_not_required])
+    def get(self):
+        """Get the public document changes feed (available to moderators).
```
tsauerwein: public or only available to moderators?

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+            Changes made by one user.
+
+            For more information about "continuation token pagination", see:
+            http://www.servicedenuages.fr/pagination-continuation-token (fr)
+        """
+        user_id = self.request.validated.get('u')
+        lang, token_id, token_time, limit = get_params(self.request)
+
+        if token_time is None:
+            token_time = datetime.datetime.now()
```
tsauerwein: Why can `token_time` be `None`?

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def get_changes_of_feed(token_id, token_time, limit, user_id=None):
+    query = DBSession. \
+        query(HistoryMetaData.id, HistoryMetaData.user_id,
+              HistoryMetaData.written_at). \
+        order_by(HistoryMetaData.written_at.desc())
+
+    # pagination filter
+    if token_id is not None and token_time:
```
tsauerwein: It should be enough to use `HistoryMetaData.id` as pagination token. No need to also use `HistoryMetaData.written_at` because the time is set automatically when inserting new rows.

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
+import datetime
+from operator import or_, and_
```
asaunier: Shouldn't it rather be
``from sqlalchemy.sql.expression import or_, and_``
?
As in https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/feed.py#L16

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+  validate_user_id_not_required
+
+from sqlalchemy.sql.expression import desc
+
+from cornice.resource import resource, view
+from sqlalchemy.orm import joinedload, load_only
+from urllib import parse as urllib_parse
+
+
+@resource(path='/document/changes', cors_policy=cors_policy)
```
asaunier: Please use ``documents`` instead of ``document``, to be consistent with the other service using ``waypoints``, ``routes``.
I think we use ``document`` for history service but it should be changed, as discussed this week with @tsauerwein .

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+class ChangesDocumentRest(object):
+    """ Unique class for returning history of a changes in documents.
```
asaunier: probably a typo here at "a changes"

c2corg_api/views/document_changes.py
asaunier: Do you really need this comment?

c2corg_api/views/document_changes.py
asaunier: Why moderators?
Anyway I see no access restriction to moderators?

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+    def get(self):
+        """Get the public document changes feed (available to moderators).
+
+        Request:
+            `GET` `/document/changes[&limit=...][&token=...][&u=]`
+
+        Parameters:
+
+            `limit=...` (optional)
+            How many entries should be returned (default: 10).
```
asaunier: 10 is probably too low. Perhaps 30 as for the advanced search?

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+        doc_ids = []
+        for ch in changes:
+            doc_ids.append(ch.id)
+
+        return load_feed(doc_ids, limit, user_id)
+
+
+def get_changes_of_feed(token_id, token_time, limit, user_id=None):
+    query = DBSession. \
+        query(HistoryMetaData.id, HistoryMetaData.user_id,
```
asaunier: Do you really need ``user_id`` and ``written_at``? Those data are not used in the calling function.

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+        user_id = self.request.validated.get('u')
+        lang, token_id, token_time, limit = get_params(self.request)
+
+        if token_time is None:
+            token_time = datetime.datetime.now()
+
+        changes = get_changes_of_feed(token_id, token_time, limit, user_id)
+
+        doc_ids = []
+        for ch in changes:
```
asaunier: What about using ``change`` as variable name instead ``ch``? It would be more explicit.

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+            doc_ids.append(ch.id)
+
+        return load_feed(doc_ids, limit, user_id)
+
+
+def get_changes_of_feed(token_id, token_time, limit, user_id=None):
```
asaunier: Do you really need that function? Since it is used only be ``load_feed``, you could move its code to ``load_feed``?
And its query could be used as ``subquery`` in the ``load_feed`` query.

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+            'quality': change.document.quality
+        },
+        'user': {
+            'user_id': change.history_metadata.user_id,
+            'name': change.history_metadata.user.name,
+            'username': change.history_metadata.user.username,
+            'lang': change.history_metadata.user.lang
+        },
+        'version_id': change.id,
+        'history_metadata_id': change.history_metadata.id,
```
asaunier: This attribute is probably useless (and not used in the UI part anyway).

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+            'type': change.document.type,
+            'quality': change.document.quality
+        },
+        'user': {
+            'user_id': change.history_metadata.user_id,
+            'name': change.history_metadata.user.name,
+            'username': change.history_metadata.user.username,
+            'lang': change.history_metadata.user.lang
+        },
+        'version_id': change.id,
```
asaunier: Perhaps a duplicate of the ``version`` attribute above?

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,160 @@
...
+        .options(load_only('document_locales_archive'),
+                 joinedload('document_locales_archive').load_only(
+                    ArchiveDocumentLocale.title)) \
+        .order_by(desc(DocumentVersion.id)) \
+        .filter(DocumentVersion.history_metadata_id.in_(doc_ids)) \
+        .limit(limit) \
+        .all()
+
+    if not doc_changes:
+        return {'feed': []}
```
asaunier: Shouldn't you return ``total: 0`` as well? It would be more consistent with the case where the feed is actually not empty.

c2corg_api/views/document_changes.py
dkocich: I wasn't sure during development - so it should be completely public to everyone?

c2corg_api/views/document_changes.py
tsauerwein: > Do you really need that function? Since it is used only be load_feed, you could move its code to load_feed?

I am not in favor of inlining the code. Like this `get` is easier to understand.

c2corg_api/views/document_changes.py
asaunier: Yes, it's not sensitive data anyway.

c2corg_api/views/feed.py
```diff
@@ -129,11 +129,14 @@ def get(self):
...
-def get_params(request):
+def get_params(request, custom_limit=False):
     lang = request.validated.get('lang')
     token_id = request.validated.get('token_id')
     token_time = request.validated.get('token_time')
-    limit = request.validated.get('limit')
+    if not request.validated.get('limit') and custom_limit:
+        limit = 30
+    else:
+        limit = request.validated.get('limit')
```
tsauerwein: What about making this:

```python
def get_params(request, default_page_limit=DEFAULT_PAGE_LIMIT):
    lang = request.validated.get('lang')
    token_id = request.validated.get('token_id')
    token_time = request.validated.get('token_time')
    limit = request.validated.get('limit')
    limit = min(
        default_page_limit if limit is None else limit,
MAX_PAGE_LIMIT)
...
```
?

This would make the signature of the function a bit clearer.

c2corg_api/tests/views/test_document_changes.py
```diff
@@ -0,0 +1,142 @@
...
+
+        self.assertIn('total', body)
+        self.assertIn('pagination_token', body)
+        self.assertIn('feed', body)
+
+        feed = body['feed']
+        self.assertEqual(4, len(feed))
+        self.assertEqual(4, body['total'])
+
+        # check that the change for the outing (latest change) is listed first
```
tsauerwein: outing?

c2corg_api/tests/views/test_document_changes.py
```diff
@@ -0,0 +1,142 @@
...
+        response = self.app.get(
+            self._prefix + '?limit=2', status=200)
+        body = response.json
+
+        document_ids = get_document_ids(body)
+        self.assertEqual(2, len(document_ids))
+        self.assertEqual(document_ids, [self.route1.document_id,
+                                        self.waypoint3.document_id])
+        pagination_token = body['pagination_token']
+        print('token je ', pagination_token)
```
tsauerwein: forgotten debug stmt

c2corg_api/tests/views/test_document_changes.py
```diff
@@ -0,0 +1,142 @@
...
+            self._prefix + '?limit=2&token=' + pagination_token, status=200)
+        body = response.json
+
+        document_ids = get_document_ids(body)
+        self.assertEqual(2, len(document_ids))
+        self.assertEqual(
+            document_ids,
+            [self.waypoint2.document_id, self.waypoint1.document_id])
+        pagination_token = body['pagination_token']
+        print('token je ', pagination_token)
```
tsauerwein: forgotten debug stmt

c2corg_api/views/document_changes.py
```diff
@@ -61,7 +61,7 @@ def get(self):
 def get_changes_of_feed(token_id, limit, user_id=None):
     query = DBSession. \
         query(HistoryMetaData.id, HistoryMetaData.user_id). \
-        order_by(HistoryMetaData.written_at.desc())
+        order_by(HistoryMetaData.id.desc(), HistoryMetaData.written_at.desc())
```
tsauerwein: Do you need to order by the two columns? Wouldn't `HistoryMetaData.id` be enough?

c2corg_api/tests/views/test_document_changes.py
```diff
@@ -0,0 +1,142 @@
...
+                RouteLocale(
+                    lang='fr', title='Mont Blanc du ciel',
+                    description='...', summary='Ski')
+            ])
+        self.session.add(self.route1)
+        self.session.flush()
+        DocumentRest.create_new_version(self.route1, contributor_id)
+        self.session.flush()
+
+        # check that the feed entry is removed
```
tsauerwein: Does this comment make sense here?

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,142 @@
...
+
+def get_changes_of_feed(token_id, limit, user_id=None):
+    query = DBSession. \
+        query(HistoryMetaData.id, HistoryMetaData.user_id). \
+        order_by(HistoryMetaData.id.desc(), HistoryMetaData.written_at.desc())
+
+    # pagination filter
+    if token_id is not None:
+        query = query.filter(
+            and_(HistoryMetaData.id < token_id))
```
tsauerwein: Why `and_`?

c2corg_api/views/validation.py
```diff
@@ -162,7 +162,8 @@ def validate_token_pagination(request, **kwargs):
     Validate token pagination parameters (limit and token) for feed.
     """
     check_get_for_integer_property(request, 'limit', False)
-    validate_token(request)
+    simple_token = True if request.path == '/documents/changes' else False
```
tsauerwein: Checking for the request path is a bit fragile. If you change the route path, you'd have to remember to change it here.

Looking at `validate_token` I am wondering if it would not make sense to create a new function `validate_simple_token`. The checks for the simple and *complex* token have little in common after all.

And for `validate_token_pagination` to avoid checking the request path, you could either create a new function or have a signature like `validate_token_pagination(simple_token, request, **kwargs)` and use the validator like:

```python
@view(validators=[functools.partial(validate_token_pagination, True), ...
```

dkocich:
I fixed comments and created new functions for simple token validation. Can you take a look today? Many thanks. :)

c2corg_api/views/document_changes.py
```diff
@@ -0,0 +1,141 @@
...
+            page.
+
+            `u=...` (optional)
+            Changes made by one user.
+
+            For more information about "continuation token pagination", see:
+            http://www.servicedenuages.fr/pagination-continuation-token (fr)
+        """
+        user_id = self.request.validated.get('u')
+        lang, token_id, token_time, limit = get_params(self.request, 30)
```
tsauerwein: Minor: You could write `lang, token_id, _, limit = ...` to indicate that you are not interested in the value of `token_time`. This is a common pattern in Haskell, but you also find it in Python code.

dkocich: Maybe __ is better?
> ..The issue is that “_” is commonly used as an alias for the gettext() function,... from http://docs.python-guide.org/en/latest/writing/style/#create-an-ignored-variable

tsauerwein: Good point! But we are not using gettext in the API project.

PR #623, opened 2017-02-08
Handle route versions without activity

tsauerwein:
`activities` is a required field, so it should not be empty. But old versions might have no activities, so we include the fields for all activities in that case.

See https://github.com/c2corg/v6_ui/pull/1476#discussion_r100052264

### DIFF
--- a/c2corg_api/tests/views/test_outing.py
+++ b/c2corg_api/tests/views/test_outing.py
@@ -171,6 +171,16 @@ class TestOutingRest(BaseDocumentTestRest):
     def test_get_version(self):
         self.get_version(self.outing, self.outing_version)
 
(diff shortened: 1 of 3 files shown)


asaunier:
What about outings? Perhaps there is the same problem (including with paragliding)?

dkocich:
I found that previous error message in logs. I am not sure if there was anything similar for outings. We should maybe check it. :)

tsauerwein:
> What about outings? Perhaps there is the same problem (including with paragliding)?

Good point, there are indeed some (22):

```
select count(*)
from guidebook.outings_archives
where activities is null or array_length(activities, 1) is null;
```

tsauerwein:
Branch updated to also handle outing versions.

PR #624, opened 2017-02-14
Invite outing participants to forum topic

arnaud-morvan:
Relate https://github.com/c2corg/v6_ui/issues/1066

### DIFF
--- a/c2corg_api/views/forum.py
+++ b/c2corg_api/views/forum.py
@@ -1,10 +1,14 @@
 import colander
 from urllib3.util.url import parse_url
 
+from c2corg_common import document_types
+
 from c2corg_api.models import DBSession
+from c2corg_api.models.association import Association
 from c2corg_api.models.cache_version import update_cache_version_direct
 from c2corg_api.models.document import DocumentLocale
 from c2corg_api.models.document_topic import DocumentTopic
+from c2corg_api.models.user import User
 from c2corg_api.security.discourse_client import get_discourse_client
 
 from cornice.resource import resource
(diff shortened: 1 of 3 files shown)


arnaud-morvan:
@tsauerwein : can you take a look.
Note that I haven't switched to bennylope/pydiscourse master because the topic/invite request was not working for me with an email.

c2corg_api/views/forum.py
```diff
@@ -102,9 +109,39 @@ def collection_post(self):
...
             DBSession.flush()
 
+            if locale.type == document_types.OUTING_TYPE:
+                self.invite_participants(client, locale, topic_id)
+
         return response
+
+    def invite_participants(self, client, locale, topic_id):
+        try:
+            forum_usernames = [r for (r, ) in (
```
tsauerwein: `r` like `forum_useRRRname`?

c2corg_api/views/forum.py
```diff
@@ -102,9 +109,39 @@ def collection_post(self):
...
             update_cache_version_direct(locale.document_id)
             DBSession.flush()
 
+            if locale.type == document_types.OUTING_TYPE:
+                self.invite_participants(client, locale, topic_id)
+
         return response
+
+    def invite_participants(self, client, locale, topic_id):
+        try:
```
tsauerwein: You are using two `try ... except`, the first one to avoid that calling `self.invite_participants` makes the request fail, and the 2nd to invite the other participants if the invitation for one participant fails?

I am ok with the 2nd `try ... except`. But I think it would make more sense to put the first `try ... except` around the call to `self.invite_participants(client, locale, topic_id)`.

arnaud-morvan:
All done, it should be good now

PR #626, opened 2017-02-16
Skip query if no values for `in()`

tsauerwein:
When running the tests of current master, a warning was shown:

```
/home/tsauerwein/projects/c2corg/v6_api/.build/venv/lib/python3.4/site-packages/sqlalchemy/sql/default_comparator.py:161: 
SAWarning: The IN-predicate on "documents_versions.history_metadata_id" was invoked with an empty sequence. This results in a contradiction, which nonetheless can be expensive to evaluate.  Consider alternative strategies for improved performance.
  'strategies for improved performance.' % expr)
```


tsauerwein:
Thanks for the review!

Issue #625, opened 2017-02-15
Whatsnew feed returns unwanted data


asaunier:
* [x] User creations are reported => we don't want this.
* [x] The total number of feed elements are returned with every API requests, which is useless. Only once is useful (when there is no token?). Perhaps it would be simpler not to provide this info. See https://github.com/c2corg/v6_api/blob/e62ee25aac632b7abff648b56482df7ffc1f256b/c2corg_api/views/document_changes.py#L76-L81

Related to https://github.com/c2corg/v6_api/issues/605 and https://github.com/c2corg/v6_api/pull/620

asaunier:
Perhaps we don't need outings changes either (only collaborative doc changes?)?
@desnoes @stef74 @celinestern @loicperrin @fbunoz What do you think?

desnoes:
@asaunier : if it is realated to the latest changes, only collaborative doc are interesting => don't need outings changes.

asaunier:
@dkocich I see you have checked the todo list items above, but I see no related PR?

PR #628, opened 2017-02-17
Fix recently changed docs API

dkocich:
still WIP

related to https://github.com/c2corg/v6_api/issues/625

It is required to rewrite `get_changes_of_feed` to filter Outings+Users earlier to get a correct list of document IDs. If document types are filtered in `load_feed`, then API don't have to return exactly 30 documents...

Other issues to fix (or planned to add)
- how to load `title_prefix` when it is available only for some doc. types? related to https://github.com/c2corg/v6_ui/pull/1466#issuecomment-280615906
- add filtering on doc. type in the future

### DIFF
--- a/c2corg_api/tests/views/test_document_changes.py
+++ b/c2corg_api/tests/views/test_document_changes.py
@@ -1,6 +1,10 @@
+import datetime
+
 from c2corg_api.models.document import DocumentGeometry
 from c2corg_api.models.document_history import DocumentVersion, HistoryMetaData
+from c2corg_api.models.outing import Outing
 from c2corg_api.models.route import Route, RouteLocale
+from c2corg_api.models.user_profile import UserProfile
 from c2corg_api.models.waypoint import Waypoint, WaypointLocale
 from c2corg_api.tests.views import BaseTestRest
 from c2corg_api.tests.views.test_feed import get_document_ids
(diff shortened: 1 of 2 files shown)


asaunier:
Keep in mind that a possible future evolution (not needed now) would be to filter on a document type. For instance to get only the changes in routes.
Which means that the code should be designed to be able to specify what document types to keep or to exclude.

dkocich:
API should filter outings and user profiles now and they shouldn't appear in the feed.

> Which means that the code should be designed to be able to specify what document types to keep or to exclude.

Ok.

c2corg_api/views/document_changes.py
```diff
@@ -50,19 +52,22 @@ def get(self):
         lang, token_id, _, limit = get_params(self.request, 30)
 
         changes = get_changes_of_feed(token_id, limit, user_id)
-        doc_ids = [change.id for change in changes]
+        doc_ids = [change.history_metadata_id for change in changes]
 
+        print(doc_ids)
```
asaunier: Please remove this debug print

c2corg_api/tests/views/test_document_changes.py
```diff
@@ -74,30 +78,59 @@ def setUp(self):
...
+        response = self.app.get(self._prefix + '?token=0', status=200)
+        body = response.json
+
+        self.assertIn('total', body)
+        self.assertNotIn('pagination_token', body)
+        self.assertIn('feed', body)
+
+        feed = body['feed']
+        self.assertEqual(0, len(feed))
+        self.assertEqual(0, body['total'])
```
tsauerwein: `total` should be total number of changes? Is `0` correct in this case?

dkocich: We removed `total` from UI completely at the end. I deleted it now on API too so it returns only empty `{feed:[]}` which is now same as for the classic `/feed`

tsauerwein: Isn't it a difference between "You've reached the end of the feed" vs. "There are no changes" (mainly when filtering on a user-id)?

asaunier:
Is the PR ready to be reviewed/merged?
@tsauerwein is it OK for you?

tsauerwein:
> @tsauerwein is it OK for you?

See my comment/question above about removing `total`:

> Isn't it a difference between "You've reached the end of the feed" vs. "There are no changes" (mainly when filtering on a user-id)?

asaunier:
What do you recommend?

I don't think we need to make complicated things about this tool.
What about using a generic message such as "No (more) changes"?

asaunier:
Anyway that kind of message is not managed by the API and should be adapted in the UI.

@dkocich Since @tsauerwein seems OK with my last suggestion, please merge the PR, thanks!

PR #632, opened 2017-02-20
Add errors to request.errors in documents/delete validation

arnaud-morvan:
Relate https://github.com/c2corg/v6_ui/pull/1504
Avoid cors errors when request validation fails.

### DIFF
--- a/c2corg_api/views/document_delete.py
+++ b/c2corg_api/views/document_delete.py
@@ -29,7 +29,6 @@ from c2corg_api.views import cors_policy, restricted_json_view
 from c2corg_api.views.image import delete_all_files_for_image
 from c2corg_api.views.validation import validate_id
 from cornice.resource import resource
-from pyramid.httpexceptions import HTTPNotFound, HTTPBadRequest
 from sqlalchemy.sql.expression import or_, and_, exists
 from sqlalchemy.sql.functions import func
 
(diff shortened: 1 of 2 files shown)


c2corg_api/views/document_delete.py
```diff
@@ -43,26 +42,52 @@ def validate_document(request, **kwargs):
...
+    if document_type not in (WAYPOINT_TYPE,
+                             ROUTE_TYPE,
+                             OUTING_TYPE,
+                             IMAGE_TYPE,
+                             ARTICLE_TYPE,
+                             BOOK_TYPE,
+                             XREPORT_TYPE):
+        request.errors.add(
+            'querystring',
+            '{}'.format(document_id),
```
tsauerwein: `name` is usually the name of the field that contains some invalid data, and not the value of that field. In this case I think we could use `document_id`.

c2corg_api/views/document_delete.py
```diff
@@ -196,7 +221,6 @@ def _get_models(document_type):
         return Book, None, ArchiveBook, None
     if document_type == XREPORT_TYPE:
         return Xreport, XreportLocale, ArchiveXreport, ArchiveXreportLocale
-    raise HTTPBadRequest('Unsupported type when deleting document')
```
tsauerwein: Do you maybe want to use an assertion here?

Issue #578, opened 2016-12-07
Advanced-search fails if offset is too big


tsauerwein:
When making a request to an advanced-search view with an offset > 10000 (for [example](http://localhost:6544/routes?bbox=457507%252C5594424%252C559627%252C5730176&offset=30000)), ElasticSearch complains:

> Result window is too large, from + size must be less than or equal to: [10000] but was [30030]. See the scroll api for a more efficient way to request large data sets. This limit can be set by changing the [index.max_result_window] index level parameter.

A simple solution could be to hide the "go to last page" button in the UI, because it does not really make sense to flip through thousand of pages.

asaunier:
> A simple solution could be to hide the "go to last page" button in the UI, because it does not really make sense to flip through thousand of pages.

@tsauerwein Can we close the issue? I think the solution you have suggested above has been implemented in the UI.

PR #638, opened 2017-02-21
Prevent requests to ES if offset+limit is too large

tsauerwein:
Closes https://github.com/c2corg/v6_api/issues/578


c2corg_api/views/document.py
```diff
@@ -65,6 +68,12 @@ def _collection_get(self, doc_type, documents_config):
             'lang': validated.get('lang')
         }
 
+        if meta_params['offset'] + meta_params['limit'] > ES_MAX_RESULT_WINDOW:
+            # ES does not process requests where offset + limit is greater
+            # than 1000, see:
```
asaunier: 1000 => 10000

Issue #608, opened 2017-01-18
Add a language preferences to the feed


asaunier:
So that users may see only content in the selected languages (in combination with the regions and activity preferences).

TODO:

- [x] store languages prefs. See for instance https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/user.py#L101-L103
- [x] add feed filtering on language
- [x] add a lang/locales list attribute in the feed table: eg. https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/feed.py#L145 => this attribute is updated whenever a new locale is created for the given document (for instance in order to advertise uploaded images for a document for which a filter-matching locale has been created in the meantime). See for instance the case of users associating (performed also when updating an outing)

PR #642, opened 2017-02-24
[WIP] Add a language preferences to the feed - API part

dkocich:
related to issue https://github.com/c2corg/v6_api/issues/608

STILL WIP!

Few notes to look at later
- I wonder why activities were updatable only for some document types? ([code here](https://github.com/c2corg/v6_api/blob/3e23b418cc0ffae0fe6a1757cc000cd39a0f3631/c2corg_api/models/feed.py#L253))
-

### DIFF
--- a/c2corg_api/models/enums.py
+++ b/c2corg_api/models/enums.py
@@ -126,6 +126,8 @@ book_type = enum(
     'book_type', attributes.book_types)
 feed_change_type = enum(
     'feed_change_type', attributes.feed_change_types)
+lang = enum(
+    'lang', attributes.default_langs)
 event_type = enum(
     'event_type', attributes.event_types)
 author_status = enum(
--- a/c2corg_api/models/feed.py
+++ b/c2corg_api/models/feed.py
@@ -5,7 +5,8 @@ from c2corg_api.models.area import Area, AREA_TYPE
 from c2corg_api.models.area_association import AreaAssociation
 from c2corg_api.models.article import ARTICLE_TYPE
 from c2corg_api.models.association import Association
-from c2corg_api.models.document import Document, UpdateType
+from c2corg_api.models.book import BOOK_TYPE
+from c2corg_api.models.document import Document, UpdateType, DocumentLocale
 from c2corg_api.models.enums import feed_change_type
 from c2corg_api.models.image import Image, IMAGE_TYPE
 from c2corg_api.models.outing import OUTING_TYPE
@@ -13,6 +14,7 @@ from c2corg_api.models.route import ROUTE_TYPE
 from c2corg_api.models.user import User
 from c2corg_api.models.user_profile import USERPROFILE_TYPE
 from c2corg_api.models.utils import ArrayOfEnum
+from c2corg_api.models.xreport import XREPORT_TYPE
 from c2corg_api.views.validation import association_keys
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.dialects.postgresql.array import ARRAY
@@ -137,6 +139,10 @@ class DocumentChange(Base):
     activities = Column(
         ArrayOfEnum(enums.activity_type), nullable=False, server_default='{}')
 
+    # list of languages (available locales of a document)
+    languages = Column(
+        ArrayOfEnum(enums.lang), nullable=False, server_default='{}')
+
     # For performance reasons, areas and users are referenced in simple integer
     # arrays in 'feed_document_changes', no PK-FK relations are set up.
     # To prevent inconsistencies, triggers are used.
@@ -178,6 +184,7 @@ class DocumentChange(Base):
         copy.document_id = self.document_id
         copy.document_type = self.document_type
         copy.change_type = self.change_type
+        copy.languages = self.languages
         copy.activities = self.activities
         copy.area_ids = self.area_ids
         if copy.document_type == OUTING_TYPE:
@@ -198,6 +205,8 @@ def update_feed_document_create(document, user_id):
     # users can be queried
     DBSession.flush()
 
+    languages = [document.available_langs]
+
     activities = []
     if document.type in [ARTICLE_TYPE, OUTING_TYPE, ROUTE_TYPE]:
         activities = document.activities
@@ -217,6 +226,7 @@ def update_feed_document_create(document, user_id):
         document_id=document.document_id,
         document_type=document.type,
         activities=activities,
+        languages=languages,
         area_ids=area_ids,
         user_ids=user_ids
     )
@@ -227,6 +237,7 @@ def update_feed_document_create(document, user_id):
 def update_feed_document_update(document, user_id, update_types):
     """Update the feed entry for a document:
 
+    - update `languages` if the locales have changed.
     - update `area_ids` if the geometry has changed.
     - update `activities` if figures have changed.
     - update `user_ids` if the document is an outing and the participants
@@ -250,9 +261,13 @@ def update_feed_document_update(document, user_id, update_types):
         update_areas_of_changes(document)
 
     # updates activities
-    if document.type in [ARTICLE_TYPE, OUTING_TYPE, ROUTE_TYPE] and \
+    if document.type in [ARTICLE_TYPE, OUTING_TYPE, ROUTE_TYPE,
+                         BOOK_TYPE, XREPORT_TYPE] and \
             UpdateType.FIGURES in update_types:
         update_activities_of_changes(document)
+
+    # update languages
+    update_languages_of_changes(document)
 
     # update users_ids/participants (only for outings)
     if document.type != OUTING_TYPE:
@@ -445,6 +460,24 @@ def update_activities_of_changes(document):
         DocumentChange.__table__.update().
         where(DocumentChange.document_id == document.document_id).
         values(activities=document.activities)
+    )
+
+
+def update_languages_of_changes(document):
+    """Update the langs of all feed entries of the given document.
+    """
+
+    languages = DBSession. \
+        query(func.array_agg(DocumentLocale.lang,
+                             type_=postgresql.ARRAY(String))). \
+        filter(DocumentLocale.document_id == document.document_id). \
+        group_by(DocumentLocale.document_id). \
+        scalar()
+
+    DBSession.execute(
+        DocumentChange.__table__.update().
+        where(DocumentChange.document_id == document.document_id).
+        values(languages=languages)
     )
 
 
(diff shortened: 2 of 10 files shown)


c2corg_api/models/enums.py
```diff
@@ -126,6 +126,8 @@ def enum(name, types):
     'book_type', attributes.book_types)
 feed_change_type = enum(
     'feed_change_type', attributes.feed_change_types)
+default_lang = enum(
+    'default_lang', attributes.langs_priority)
```
asaunier: Using ``attributes.default_langs`` would be more consistent I think.
https://github.com/c2corg/v6_common/blob/1e3fe0991ecc8acbb0d9234f96eb71413089aa7f/c2corg_common/attributes.py#L2

In addition what about renaming the enumeration to ``langs`` instead? I don't really understand why it is "default".

c2corg_api/models/feed.py
```diff
@@ -137,6 +139,10 @@ class DocumentChange(Base):
     activities = Column(
         ArrayOfEnum(enums.activity_type), nullable=False, server_default='{}')
 
+    # enums.default_lang ... OR ... ARRAY(String)
+    lang_ids = Column(
```
asaunier: What about naming this attribute ``langs`` instead? Those are not really ids.
Or ``available_langs``?

c2corg_api/models/feed.py
```diff
@@ -198,6 +205,9 @@ def update_feed_document_create(document, user_id):
     # users can be queried
     DBSession.flush()
 
+    lang_ids = []
```
asaunier: Is this line really needed since the variable is redefined the next line?

c2corg_api/models/feed.py
```diff
@@ -250,10 +262,13 @@ def update_feed_document_update(document, user_id, update_types):
         update_areas_of_changes(document)
 
     # updates activities
-    if document.type in [ARTICLE_TYPE, OUTING_TYPE, ROUTE_TYPE] and \
+    if document.type in [ARTICLE_TYPE, OUTING_TYPE, ROUTE_TYPE,
+                         BOOK_TYPE, XREPORT_TYPE] and \
```
asaunier: :+1:

c2corg_api/models/feed.py
```diff
@@ -448,6 +463,24 @@ def update_activities_of_changes(document):
...
+def update_languages_of_changes(document):
+    """Update the langs of all feed entries of the given document.
+    """
+
+    lang_ids = DBSession. \
+        query(func.array_agg(DocumentLocale.lang,
+                             type_=postgresql.ARRAY(String))). \
+        filter(DocumentLocale.document_id == document.document_id). \
+        group_by(DocumentLocale.document_id). \
+        scalar()
```
asaunier: Perhaps it's possible to use ``subquery()`` instead to avoid having 2 distinct SQL queries?

c2corg_api/scripts/migration/init_feed.py
```diff
@@ -15,6 +15,7 @@ def migrate(self):
             self.session_target.execute(SQL_DOCUMENT_CREATE_CHANGE)
             self.session_target.execute(SQL_AREAS_FOR_CHANGES)
             self.session_target.execute(SQL_ACTIVITIES_FOR_CHANGES)
+            self.session_target.exececute(SQL_LANGUAGES_FOR_CHANGES)
```
asaunier: There's a typo in ``exececute``.

asaunier: But do you really need to do this? We won't run the migration again!

c2corg_api/views/__init__.py
```diff
@@ -177,12 +180,21 @@ def set_best_locale(documents, preferred_lang, expunge=True):
 
 
 def get_best_locale(available_locales, preferred_lang):
-    if preferred_lang in available_locales:
-        best_locale = available_locales[preferred_lang]
-    else:
+
+    print(' LOOOOOOOOOOOOOOOOOOOOOOOG ', preferred_lang, available_locales)
+
+    if type(preferred_lang) is list:
```
asaunier: Do we really have the case of several preferred languages?

c2corg_api/views/feed.py
```diff
@@ -182,24 +191,46 @@ def create_personal_filter(user):
...
-        if area_filter is not None and activity_filter is not None:
-            area_activity_filter = and_(area_filter, activity_filter)
         elif area_filter is not None:
-            area_activity_filter = area_filter
+            area_activity_language_filter = area_filter
         elif activity_filter is not None:
-            area_activity_filter = activity_filter
+            area_activity_language_filter = activity_filter
+        elif language_filter is not None:
+            area_activity_language_filter = language_filter
```
asaunier: All those conditions seem a bit complicated.
Why don't you use the following approach:
```
filters = []
if area_filter is not None:
    filters.append(area_filter)
if activity_filter is not None:
    filters.append(activity_filter)
if language_filter is not None:
    filters.append(language_filter)
if len(filters):
    area_activity_language_filter = and_(filters)
```
Perhaps you don't even need the ``is not None`` explicit conditions?

c2corg_api/views/validation.py
```diff
@@ -57,6 +57,28 @@ def validate_lang_(lang, request):
...
+        elif lang in default_langs:
+            validated.append(lang)
+
+        if validated:
+            request.validated['lang'] = validated
+
+        # if lang in default_langs:
+        #     request.validated['lang'] = lang
+        else:
+            request.errors.add('querystring', 'lang', 'invalid lang')
```
asaunier: This means that if you pass an incorrect lang in the comma-separated list, (eg. ``fr,de,ru``), the incorrect lang will be ignored but no error will be thrown, right?
Shouldn't it throw an error whenever an invalid lang is detected?

c2corg_api/views/validation.py
```diff
@@ -90,7 +112,7 @@ def validate_preferred_lang_param(request, **kwargs):
     is correct ("...?pl=...").
     """
     lang = request.GET.get('pl')
-    validate_lang_(lang, request)
+    validate_pref_lang_(lang, request)
```
asaunier: Are you sure this change is needed? Do we really have cases where we pass a list of langs as ``pl``?

c2corg_api/views/__init__.py
dkocich: I think it would be good to allow users to select more languages in the feed. I think that you too.
image
It would be good later to allow users to change a priority of the best locale which shows in the feed (right now it is handled by the `get_best_locale` [function](https://github.com/c2corg/v6_api/blob/179b1d45baf224c1fd80a937f8c83714b689d0b4/c2corg_api/views/__init__.py#L179) which returns the best language based on the `lang_priority` [attribute](https://github.com/c2corg/v6_common/blob/1e3fe0991ecc8acbb0d9234f96eb71413089aa7f/c2corg_common/attributes.py#L3)).

c2corg_api/scripts/migration/init_feed.py
dkocich: Ok, so we probably won't need this (IMO I would keep it in the code just for completeness).

c2corg_api/views/__init__.py
asaunier: I don't see the place where ``get_best_locale`` is called with a list of preferred languages?

c2corg_api/scripts/migration/init_feed.py
asaunier: I disagree with that: it's useless. The migration script will no longer be executed and we will no longer apply changes to the migration script when changing the DB model in the future => please remove those changes.

On the other hand don't you need to add an alembic script for the need language attributes in the feed?

c2corg_api/models/feed.py
```diff
@@ -198,6 +205,8 @@ def update_feed_document_create(document, user_id):
     # users can be queried
     DBSession.flush()
 
+    languages = [document.available_langs]
```
tsauerwein: Isn't `document.available_langs` already a list? And are you sure `document.available_langs` is set at this point? I am wondering if this covered by your tests?

c2corg_api/models/feed.py
```diff
@@ -250,10 +261,14 @@ def update_feed_document_update(document, user_id, update_types):
...
 
     # updates activities
-    if document.type in [ARTICLE_TYPE, OUTING_TYPE, ROUTE_TYPE] and \
+    if document.type in [ARTICLE_TYPE, OUTING_TYPE, ROUTE_TYPE,
+                         BOOK_TYPE, XREPORT_TYPE] and \
             UpdateType.FIGURES in update_types:
         update_activities_of_changes(document)
 
+    # update languages
+    update_languages_of_changes(document)
```
tsauerwein: You could do:

```python
if UpdateType.LANG in update_types:
    update_languages_of_changes(document)
```

c2corg_api/tests/models/test_feed.py
```diff
@@ -66,13 +68,30 @@ def test_check_area_ids_on_create_invalid(self):
...
+        change = DocumentChange(
+            user=self.user1, change_type='created', document=self.waypoint,
+            document_type=WAYPOINT_TYPE,
+            area_ids=[], user_ids=[], languages=[]
+        )
+        try:
+            self.session.add(change)
+            self.session.flush()
+        except Exception as exc:
+            self.assertTrue('Invalid languages: ru' in exc.orig.pgerror)
```
tsauerwein: Where does `'ru'` come from?

c2corg_api/views/validation.py
```diff
@@ -57,6 +57,27 @@ def validate_lang_(lang, request):
             request.errors.add('querystring', 'lang', 'invalid lang')
 
 
+def validate_pref_lang_(lang, request):
+    """Checks given langs are one of the available langs.
+    """
+    if lang is not None:
+        validated = []
+        if ',' in lang:
```
tsauerwein: This check is not needed. For example `'en'.split(',')` returns `['en']`.

c2corg_api/views/__init__.py
```diff
@@ -177,7 +177,12 @@ def set_best_locale(documents, preferred_lang, expunge=True):
 
 
 def get_best_locale(available_locales, preferred_lang):
-    if preferred_lang in available_locales:
+    if type(preferred_lang) is list:
```
tsauerwein: Where are you passing a list as value for `preferred_lang`?

Issue #217, opened 2016-04-05
Create a protect/unprotect document webservice


asaunier:
(only moderators may edit protected documents)

UI issue: https://github.com/c2corg/v6_ui/issues/239

asaunier:
Protected documents are already supported in the API:
* a ``protected`` attribute is already available in the models: https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/document.py#L51
* protection status was migrated from v5
* only moderators can edit a protected document: https://github.com/c2corg/v6_api/blob/8091cc20a144f4f71fe7bf93623f0751608ec832/c2corg_api/views/document.py#L243-L244
* if a document is protected, the ``info`` webservice returns a ``protected`` flag (example?)
* if not, the  ``info`` webservice does not return this flag (eg. https://api.camptocamp.org/waypoints/37737/fr/info)
* the ``protected`` attribute is provided (either true or false) when listing documents: https://api.camptocamp.org/waypoints

TODO:
* [ ] create a service to set the ``protected`` attribute to true or false (only for moderators)
* [ ] add a unit test about this new service

PR #644, opened 2017-03-15
Add protect/unprotect document service

asaunier:
Related to https://github.com/c2corg/v6_api/issues/217

Inspired from the user block/unblock service https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/user_block.py


asaunier:
@arnaud-morvan OK for merging the PR?

Issue #645, opened 2017-03-16
Delete a locale from a document


asaunier:
Note:
* check if the locale to delete is the only one available => if yes, delete the whole document (with the same constraints and effects as the "delete document" tool) => use the standard deleting tool
* if other locales are available => only remove the given locale (+ archived version of the locale and history metadata)

desnoes:
It seems it is not possible to deletec more than one linguistic version of a document.
Cf. experiment detailed in https://github.com/c2corg/v6_ui/issues/1555

PR #648, opened 2017-03-29
Add locale deletion service

asaunier:
Related to https://github.com/c2corg/v6_api/issues/645

TODO
* [x] create a ``DELETE documents/delete/{document_id}/{lang}`` service similar to https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document_delete.py#L145
* [x] validate document id and lang exist and check if other locales are available
* [x] if no other locale => use the existing ``DELETE documents/delete/{document_id}`` logic (factorize code)
* if other locales exist => only remove the locale:
 * [x] _remove_history_metadata (of the locale only => add a lang filter)
 * [x] _remove_archive_locale (of the given locale only)
 * [x] _remove_locale (of the given locale only)
 * [x] update cache
 * [x] update ES index
 * [ ] ~~update feed to update the lang list~~ => not relevant until  https://github.com/c2corg/v6_api/pull/642 is actually merged
* [x] ~~remove comments related to the locale (note: when deleting the whole document, only the references to the comments are removed, not the comments themselves)~~ => already managed by ``_remove_locale``

### DIFF
--- a/c2corg_api/tests/views/test_document_delete.py
+++ b/c2corg_api/tests/views/test_document_delete.py
@@ -7,7 +7,8 @@ from c2corg_api.models.association import Association, AssociationLog
 from c2corg_api.models.book import Book, ArchiveBook
 from c2corg_api.models.document import (
     Document, DocumentLocale, DocumentGeometry, ArchiveDocument,
-    ArchiveDocumentLocale, ArchiveDocumentGeometry, UpdateType)
+    ArchiveDocumentLocale, ArchiveDocumentGeometry, UpdateType,
+    get_available_langs)
 from c2corg_api.models.document_history import DocumentVersion
 from c2corg_api.models.document_topic import DocumentTopic
 from c2corg_api.models.feed import (
@@ -368,7 +369,7 @@ class TestDocumentDeleteRest(BaseTestRest):
             self._prefix + str(self.waypoint1.document_id), {},
             headers=headers, status=403)
 
-    def test_non_existing_document(self):
+    def test_delete_non_existing_document(self):
         self._delete(-9999999, 400)
 
     def test_delete_main_waypoint(self):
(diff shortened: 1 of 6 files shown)


tsauerwein:
> if other locales exist => only remove the locale:

The cache version would also have to be updated. And regarding ElasticSearch: The locales of a document are stored in `available_locales` in ES. If a locale is deleted, the document would have to be updated. The easiest would probably be to create a new document version, so that the syncer script would be noticed automatically.

asaunier:
PR updated.

There is probably a problem when removing entries in tables ``documents_versions`` and ``history_metadata``!?

```
FAIL: test_delete_locale_waypoint (c2corg_api.tests.views.test_document_delete.TestDocumentDeleteRest)
  File "/home/travis/build/c2corg/v6_api/c2corg_api/tests/views/test_document_delete.py", line 685, in test_delete_locale_waypoint
    self._test_delete_locale(self.waypoint3.document_id, 'fr')
  File "/home/travis/build/c2corg/v6_api/c2corg_api/tests/views/test_document_delete.py", line 682, in _test_delete_locale
    initial_other_locale_versions_count)
nose.proxy.AssertionError: 0 != 1
```

> And regarding ElasticSearch: The locales of a document are stored in available_locales in ES. If a locale is deleted, the document would have to be updated. The easiest would probably be to create a new document version, so that the syncer script would be noticed automatically.

Creating a new document version would be an option but when deleting a locale we have no other difference that would justify a new version (the new version would remain the same for remaining locales) as in https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document_merge.py#L142-L147

I was considering adding an optional ``lang`` attribute to https://github.com/c2corg/v6_api/blob/ea3e5dcbc47f2a4eea469a8467e8ff0c1d2392d7/c2corg_api/models/es_sync.py#L39-L51 to make the syncer know only a locale should be deleted.
But I am not sure it is that possible to pass a "delete" request to ES for a locale only in https://github.com/c2corg/v6_api/blob/ea3e5dcbc47f2a4eea469a8467e8ff0c1d2392d7/c2corg_api/scripts/es/sync.py#L267-L282

Perhaps one should rather trigger a standard "index" request as in https://github.com/c2corg/v6_api/blob/3817d41971c444d9023a4b14e62071d8d930a1dd/c2corg_api/search/mapping.py#L122-L135 but with ``available_locales`` set to the remaining languages (as suggested by @tsauerwein ?).

tsauerwein:
> Perhaps one should rather trigger a standard "index" request as in https://github.com/c2corg/v6_api/blob/3817d41971c444d9023a4b14e62071d8d930a1dd/c2corg_api/search/mapping.py#L122-L135 but with available_locales set to the remaining languages

Yes, that would be better.

asaunier:
> Yes, that would be better.

That's the next (and hopefully last) point to do with this PR.

Concerning the ``DocumentVersion`` and ``HistoryMetaData`` cleaning I have made sure that only metadata related to a single locale were removed.

Here is an example to understand how it works...

Say I have created a document in FR, then created the EN version, then changed some figure (an history metadata is created, linked to both locales), and finally made a change in the EN locale:
```
c2corg_alex=# select lang, history_metadata_id from guidebook.documents_versions where document_id = 842149;
 lang | history_metadata_id 
------+---------------------
 en   |             1295016
 fr   |             1295015
 en   |             1295015
 en   |             1295014
 fr   |             1295013
(5 rows)
```
History metadata whose id is 1295015 is linked twice:
```
c2corg_alex=# select history_metadata_id, lang, count(*) over (partition by history_metadata_id) from guidebook.documents_versions where document_id = 842149;
 history_metadata_id | lang | count 
---------------------+------+-------
             1295013 | fr   |     1
             1295014 | en   |     1
             1295015 | fr   |     2
             1295015 | en   |     2
             1295016 | en   |     1
(5 rows)
```
I can use the above query as a subquery and filter on the lang and number of related locales to make sure I get only ids for history metadata linked to only one locale:
```
c2corg_alex=# select history_metadata_id from (select history_metadata_id, lang, count(*) over (partition by history_metadata_id) as cnt from guidebook.documents_versions where document_id = 842149) as t where t.lang = 'en' and t.cnt = 1;
 history_metadata_id 
---------------------
             1295014
             1295016
(2 rows)

c2corg_alex=# select history_metadata_id from (select history_metadata_id, lang, count(*) over (partition by history_metadata_id) as cnt from guidebook.documents_versions where document_id = 842149) as t where t.lang = 'fr' and t.cnt = 1;
 history_metadata_id 
---------------------
             1295013
(1 row)
```

This is what https://github.com/c2corg/v6_api/blob/fe65168ea312f6d8042dd869b14667a3ba38394a/c2corg_api/views/document_delete.py#L313-L324 does.

asaunier:
A few explanations about what this PR does:
* creates a new ``es_deleted_locales`` table to store references to deleted locales
* the syncer uses the table to figure out what documents have changed because some locales have been removed and reindex them (as if changes have been done)
* factorize the ``DeleteDocumentRest`` and ``DeleteDocumentLocaleRest`` services (because they share common functions and behaviours)

asaunier:
I have added a test (making sure that deleting the only locale of a document fully removes the document) and fixed a conflict with https://github.com/c2corg/v6_api/pull/651

The PR is ready for review.

asaunier:
@arnaud-morvan OK for merging?

arnaud-morvan:
I've tested the locale is removed from elastic search (and it good).
On another side, the syncer make an enormous number of requests on the database for each changed document, but this should not be related to this PR.

c2corg_api/tests/views/test_document_delete.py
```diff
@@ -643,3 +644,118 @@ def image_service_mock(url, request):
...
+        self.assertEqual(0, count)
+
+        # but not for other locales
+        other_locale_versions_count = self.session. \
+            query(DocumentVersion). \
+            filter(DocumentVersion.document_id == document_id). \
+            filter(DocumentVersion.lang != lang).count()
+        self.assertEqual(other_locale_versions_count,
+                         initial_other_locale_versions_count)
+
```
asaunier: Do we really need to test a simple ``DBSession.add(ESDeletedLocale(...))``?

arnaud-morvan: This is not strictly needed, but match with the tests policy we have used on the API, so I've noted that is not done. If you think it's not necessary, feel free to merge.

asaunier:
@arnaud-morvan Thanks for reviewing!

I prefer not adding the test about ``ESDeletedLocale``: we have not done so for ``ESDeletedDocument`` and this table is actually not real data but a way to make ES know the document has changed.

PR #650, opened 2017-04-06
Fix document delete order

arnaud-morvan:
Relate https://github.com/c2corg/v6_ui/issues/1552


asaunier:
Please merge your PR soon. It will probably create a conflict with https://github.com/c2corg/v6_api/pull/648 and I would like to fix it.

PR #651, opened 2017-04-06
Remove waypoint from route_archive.main_waypoint

arnaud-morvan:
Relate https://github.com/c2corg/v6_ui/issues/1552


arnaud-morvan:
@asaunier : This may create conflicts too

Issue #653, opened 2017-04-12
Not possible to delete more than one linguistic version of a document


asaunier:
See https://github.com/c2corg/v6_ui/issues/1555#issuecomment-293480695

@desnoes said

> I cannot delete this version of a route : http://www.demov6.camptocamp.org/routes/828490/en/chamechaude-essai-en. Same thing with the italian version : http://www.demov6.camptocamp.org/routes/828490/it/essi-it
>
> WP: same pb with http://www.demov6.camptocamp.org/waypoints/828492/en/essai-en
>
> I did the following experiment:
>
> i've created a WP (http://www.demov6.camptocamp.org/waypoints/828493/fr/toto-sommet) in french.
i've created an english version of this document and then deleted it successfully.
i've created an italian version, and then tried to delete. It was not possible (error message).
It seems it is not possible to delete more than one linguistic version of a document.

@arnaud-morvan replied

> This may be due to the ESDeletedLocale primary key, this table is not purged.

=> The primary key of ``ESDeletedLocale`` should indeed be ``document_id`` + ``lang`` instead of ``document_id``.

PR #652, opened 2017-04-11
Update avalanche slopes values

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1519

Requires https://github.com/c2corg/v6_common/pull/55 and https://github.com/c2corg/v6_ui/pull/1593

Tested successfully in my UI/API instances.

PR #654, opened 2017-04-12
Fix ESDeletedLocale primary key

asaunier:
Related to https://github.com/c2corg/v6_api/issues/653

~~Not tested yet because my DB is currently in the status of this other (unmerged) alembic script https://github.com/c2corg/v6_api/pull/652
Should probably be updated the latter is merged.~~


asaunier:
Updated and tested successfully (it is now possible to remove several locales of the same document).
Please review.

asaunier:
Thanks @arnaud-morvan !

Issue #658, opened 2017-05-01
Empty point geom should not be accepted when creating a route


asaunier:
The home feed is broken in the UI demo because the API feed request returns a 500 error.
At least the anonymous feed, my personal feed works OK, perhaps because my filters strip a compromised document out of the feed.

Here are the errors reported by the demo API logs (I have added some line breaks and blank lines to make them a bit more readable):
```
{"message": "Unexpected error: Out of range float values are not JSON compliant", "levelname": "ERROR", "name": "c2corg_api.views", "asctime": "2017-05-01 07:54:30,397", "funcName": "catch_all_error_handler", "lineno": 60, "pathname": "/var/www/c2corg_api/views/__init__.py", "stack_info": null, "module": "__init__", "created": 1493625270.3972888, "threadName": "MainThread", "processName": "MainProcess", "getMessage": null, "exc_info": "Traceback (most recent call last):\n

File \"/var/www/.build/venv/lib/python3.4/site-packages/c2corg_common/utils/caching.py\", line 51, in get_or_create_multi\n    should_cache_fn=should_cache_fn)\n
File \"/var/www/.build/venv/lib/python3.4/site-packages/dogpile/cache/region.py\", line 937, in get_or_create_multi\n    new_values = creator(*keys_to_get)\n
File \"/var/www/.build/venv/lib/python3.4/site-packages/c2corg_common/utils/caching.py\", line 113, in fn\n    raise CreatorException(exc)\nc2corg_common.utils.caching.CreatorException: Out of range float values are not JSON compliant\n\n

During handling of the above exception, another exception occurred:\n\nTraceback (most recent call last):\n

File \"/var/www/.build/venv/lib/python3.4/site-packages/pyramid/tweens.py\", line 22, in excview_tween\n    response = handler(request)\n
File \"/var/www/.build/venv/lib/python3.4/site-packages/pyramid_tm/__init__.py\", line 109, in tm_tween\n    reraise(*exc_info)\n
File \"/var/www/.build/venv/lib/python3.4/site-packages/pyramid_tm/compat.py\", line 15, in reraise\n    raise value\n
File \"/var/www/.build/venv/lib/python3.4/site-packages/pyramid_tm/__init__.py\", line 88, in tm_tween\n    response = handler(request)\n
File \"/var/www/.build/venv/lib/python3.4/site-packages/pyramid/router.py\", line 158, in handle_request\n    view_name\n
[...]

PR #659, opened 2017-05-01
Improve detection of invalid geometries

asaunier:
Fixes https://github.com/c2corg/v6_api/issues/658

### DIFF
--- a/c2corg_api/ext/colander_ext.py
+++ b/c2corg_api/ext/colander_ext.py
@@ -6,8 +6,8 @@ from colander import (null, Invalid, SchemaType)
 from geoalchemy2 import WKBElement
 from geomet import wkb
 from geoalchemy2.compat import buffer, bytes
-import json
 import geojson
+from geojson.validation import is_polygon, checkListOfObjects
 
 
 class Geometry(SchemaType):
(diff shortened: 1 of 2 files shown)


asaunier:
Gasp, according to Travis it seems it detects too many "invalid" geometries :(

c2corg_api/ext/colander_ext.py
```diff
@@ -53,11 +53,16 @@ def deserialize(self, node, cstruct):
...
             return null
         try:
-            geojson = json.loads(cstruct)
-            geom_type = geojson['type'].upper()
+            geojson_data = json.loads(cstruct)
+            geom_type = geojson_data['type'].upper()
             allowed_types = self.geometry_types
             if geom_type in allowed_types:
-                return wkbelement_from_geojson(geojson, self.srid)
+                validation = geojson.is_valid(geojson_data)
```
tsauerwein: I think this should be:
```
validation = geojson.is_valid(geojson.loads(cstruct))
```

Though it would be a pity that the string is parsed twice. It would also be good to test if 3D/4D geometries still work.

asaunier: Thanks! I have made the change you suggested and run the routes unittests: they indeed pass except the 3D/4D tests.

tsauerwein: It seems that `geojson.is_valid` only accepts 2D and 3D geometries, not 4D:
https://github.com/frewsxcv/python-geojson/blob/master/geojson/validation.py#L29

:-/

asaunier: > It seems that geojson.is_valid only accepts 2D and 3D geometries, not 4D:

In that case (testing a Point geometry), is it really a problem?

According to my tests the problem is rather that 4D multilinestrings are detected as MultiPoint (?!) whose points are indeed rejected because not compatible with the 4D:
https://github.com/frewsxcv/python-geojson/blob/master/geojson/validation.py#L32-L33

I will soon upgrade my PR with "custom" validation.

asaunier:
I have updated the PR with commit d27ecb7 which adds a custom geometry validation function replacing ``geojson.is_valid`` that is not compatible with some geometries (for instance 4D multilinestrings).

tsauerwein:
:+1:

asaunier:
Thanks @tsauerwein !

PR #662, opened 2017-05-07
Add support for slacklining

tsauerwein:
See https://github.com/c2corg/v6_ui/issues/1543

This PR adapts the API to support slacklining:
- Adds `slacklining` as new activity
- Adds `slackline_spot` as new waypoint type
- Adds new fields to the routes model (`slackline_type`, `slackline_height`)
- Adds new fields to the routes_local model (`slackline_anchor1`, `slackline_anchor2`)
- Adds new fields to the waypoints model (`slackline_types`, `slackline_length_min`, `slackline_length_max`)

To make it possible to filter routes on the slackline type, I had to adapt the ES mapping. When applying this PR, a script has to be run to change the mapping, see https://github.com/c2corg/v6_api/tree/highline/es_migration

Still todo:

* [x] For the import, ignore changes of the [C2C account](https://www.camptocamp.org/profiles/2/fr) user in the feed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,7 @@ test: .build/venv/bin/nosetests template .build/dev-requirements.timestamp .buil
 
 .PHONY: lint
 lint: .build/venv/bin/flake8
-	.build/venv/bin/flake8 c2corg_api
+	.build/venv/bin/flake8 c2corg_api es_migration
 
 .PHONY: install
 install: .build/requirements.timestamp template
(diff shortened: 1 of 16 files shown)


tsauerwein:
Thanks for the review!

Waiting with the merge until the import is ready (see https://github.com/c2corg/v6_ui/issues/1543).

fbunoz:
I don't know the API code, then, I can't review your code.
But is this PR can be used as an example of adding new fields in the API ?

tsauerwein:
> But is this PR can be used as an example of adding new fields in the API ?

Yes! Also on how to make new fields searchable.

Issue #663, opened 2017-05-07
API is very slow if there is no filter


cbeauchesne:
It's not always the case, I would say 10%, but when it happens, the API is very slow if no filter is used.

To reproduce, simply go here https://www.camptocamp.org/outings, and unselect "filter on map" (which is checked by default, and so, request is filtered).

You will see that filtered request is quite instant, but the non-filtered one can take up to 20 seconds.

tsauerwein:
Good observation!

If no filters are given, the documents are [directly loaded from the database](https://github.com/c2corg/v6_api/blob/a300ced/c2corg_api/views/document.py#L82). For outings, the documents are [ordered](https://github.com/c2corg/v6_api/blob/5fa6347/c2corg_api/views/document_listings.py#L37) by `date_end` and the document id. Currently there is no index for this kind of query, this should explain the bad performance.

cbeauchesne:
Can we consider that as a bug? Maybe a dirty fix could be a fake filter that always returns TRUE?, and thus use optimized code?

asaunier:
> If no filters are given, the documents are directly loaded from the database. For outings, the documents are ordered by date_end and the document id. Currently there is no index for this kind of query, this should explain the bad performance.

@tsauerwein Thanks for the hint!

With PR #664 I suggest an Alembic script that adds an index on outing's ``date_end`` attribute.
Since we use Alembic for the DB creation as well, I guess it is not needed to explicitely define the index as in http://docs.sqlalchemy.org/en/latest/core/constraints.html#indexes ?

I have also find this post http://stackoverflow.com/a/40321047 saying that defining the primary key as (``document_id``, ``date_end``) should be 
> significantly (2-4 times) faster then an index on a date column

On the other hand the contributor says later that
>  an ordinary index significantly improves performance and should be sufficient

cbeauchesne:
Seems not to happen anymore

PR #664, opened 2017-05-08
[WIP] Add index on outing date_end attribute

asaunier:
Related to https://github.com/c2corg/v6_api/issues/663

### DIFF
--- a/alembic_migration/versions/8bd0478eea2a_add_index_on_outing_date_end.py
+++ b/alembic_migration/versions/8bd0478eea2a_add_index_on_outing_date_end.py
@@ -1,0 +1,23 @@
+"""Add index on outing date_end
+
+Revision ID: 8bd0478eea2a
+Revises: 91b1beed9a1c
+Create Date: 2017-05-08 19:43:33.817581
+
+"""
+from alembic import op
+
+
+# revision identifiers, used by Alembic.
+revision = '8bd0478eea2a'
+down_revision = '91b1beed9a1c'
+branch_labels = None
+depends_on = None
+
+def upgrade():
+    op.create_index(
+        'outing_date_end_idx', 'outings', ['date_end'], schema='guidebook')
+
+
+def downgrade():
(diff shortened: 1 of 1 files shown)


alembic_migration/versions/8bd0478eea2a_add_index_on_outing_date_end.py
```diff
@@ -0,0 +1,23 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '8bd0478eea2a'
+down_revision = '91b1beed9a1c'
+branch_labels = None
+depends_on = None
+
+def upgrade():
+    op.create_index(
+        'outing_date_end_idx', 'outings', ['date_end'], schema='guidebook')
```
tsauerwein: I would say I should be more something like:
```
Index(
            'ix_guidebook_òutings_date_end_and_document_id',
            date_end.desc(), document_id, postgresql_using='btree'),
```

But I don't have a database with enough data at hand to test.

Also, don't you also want to add the index to the model file?

asaunier: See also https://github.com/c2corg/v6_api/issues/663#issuecomment-299944457

> I would say it should be more something like:

Similar to the feed index on date desc + id:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/feed.py#L166-L174

tsauerwein: Yes, but maybe test if the index is used (on a database with many outings).

cbeauchesne:
@asaunier is this PR still useful ? Do you think we should merge, or close ?

asaunier:
I don't remember but it seems WIP and as you say in the related issue, it seems the problem (slow loading when no filter) no longer happens (?).
Perhaps we can close the PR.

cbeauchesne:
Ok, I close it. I confirm that I did not noticed any slowness on this kind of request since a while. Thank you

Issue #670, opened 2017-05-23
Support document changes reverting


asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1068
The UI PR is https://github.com/c2corg/v6_ui/pull/1101

Problems we have when editing an archived version of a document (to revert to this archived version or use it as base for new changes):
* [ ] the API returns a version conflict when the UI sends the versions of an edited archived version. For instance:
> Conflict : version of locale 'fr' has changed
* [ ] associations are not managed when editing an archived version and are not passed within the editing request to the API. It seems the API requires this attribute for routes and outings. For instance when editing a route:
> associations.waypoints : at least one waypoint required

asaunier:
>  the API returns a version conflict when the UI sends the versions of an edited archived version. For instance

The API indeed checks that the versions (of the geometry, locales and figures data) match the current versions of the document and throw a "Conflict" error if not:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/waypoint.py#L217-L219
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document.py#L246

A possibility would be to disable this check for moderators and restrict the revert tool to moderators (which would by the way avoid the UI problems described at https://github.com/c2corg/v6_ui/pull/1101#issuecomment-303716041).

@tsauerwein Do you think skipping the version checks for moderators would cause problems? I guess that for instance if a moderator makes a standard edition at the same time than another user, the conflict with the latter user's changes will not be reported and those changes will be silently be archived.

Another option would be to provide the current document versions (as well as the current ``protected`` attribute and perhaps the associated users or author id? see https://github.com/c2corg/v6_ui/pull/1101#issuecomment-303688922) in addition to the ``previous_version_id``, ``next_version_id`` and ``version`` attributes provided as well as the standard ``document`` object. See for instance:
https://api.camptocamp.org/routes/46867/fr/334485

Of course the latter approach requires more work, in the API as well as in the UI, even though the results might be cleaner.

asaunier:
> associations are not managed when editing an archived version and are not passed within the editing request to the API. It seems the API requires this attribute for routes and outings. For instance when editing a route

For instance when route editing request is sent, the API validates the change against some criterias, including associated waypoints, see ``validate_required_associations``:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/route.py#L124
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/route.py#L69-L82

[...]

PR #671, opened 2017-06-02
Add a document change reverting service

asaunier:
Implementing the alternate approach described at https://github.com/c2corg/v6_api/issues/670#issuecomment-303770491

UI part: https://github.com/c2corg/v6_ui/pull/1669

### DIFF
--- a/c2corg_api/views/document_revert.py
+++ b/c2corg_api/views/document_revert.py
@@ -2,27 +2,21 @@ import logging
 
 from c2corg_api import DBSession
 from c2corg_api.models.area import AREA_TYPE, Area, ArchiveArea
-from c2corg_api.models.area_association import update_areas_for_document
 from c2corg_api.models.article import ARTICLE_TYPE, Article, ArchiveArticle
 from c2corg_api.models.book import BOOK_TYPE, Book, ArchiveBook
-from c2corg_api.models.cache_version import update_cache_version
 from c2corg_api.models.document import (
-    Document, DocumentLocale, ArchiveDocumentLocale, UpdateType)
+    Document, DocumentLocale, ArchiveDocumentLocale)
 from c2corg_api.models.document_history import DocumentVersion
-from c2corg_api.models.feed import update_feed_document_update
 from c2corg_api.models.image import IMAGE_TYPE, Image, ArchiveImage
 from c2corg_api.models.outing import (
     OUTING_TYPE, Outing, OutingLocale, ArchiveOuting, ArchiveOutingLocale)
 from c2corg_api.models.route import (
     ROUTE_TYPE, Route, RouteLocale, ArchiveRoute, ArchiveRouteLocale)
-from c2corg_api.models.topo_map import MAP_TYPE
-from c2corg_api.models.topo_map_association import update_maps_for_document
 from c2corg_api.models.waypoint import (
     WAYPOINT_TYPE, Waypoint, WaypointLocale, ArchiveWaypoint,
     ArchiveWaypointLocale)
 from c2corg_api.models.xreport import (
     XREPORT_TYPE, Xreport, XreportLocale, ArchiveXreport, ArchiveXreportLocale)
-from c2corg_api.search.notify_sync import notify_es_syncer
 from c2corg_api.views import cors_policy, restricted_json_view
 from c2corg_api.views.area import update_associations
 from c2corg_api.views.document import DocumentRest
@@ -33,9 +27,8 @@ from colander import (
     MappingSchema, SchemaNode, Integer, String, required, OneOf)
 from cornice.resource import resource
 from cornice.validators import colander_body_validator
-from pyramid.httpexceptions import HTTPBadRequest, HTTPConflict
+from pyramid.httpexceptions import HTTPBadRequest
 from sqlalchemy.orm import joinedload, contains_eager
-from sqlalchemy.orm.exc import StaleDataError
 from sqlalchemy.orm.util import with_polymorphic
 from sqlalchemy.sql.expression import exists, and_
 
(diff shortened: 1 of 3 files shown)


asaunier:
I'm trying to use a similar approach for the reverting service then for the updating one, see method ``_put()`` at https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document.py#L229

But I'm quite struggling with document's ``version`` attributes.
Since we want to save some old archive data as a new version of the document, we cannot check that the ``version`` values are up-to-date => in https://github.com/c2corg/v6_api/blob/056d1edeb98abb9bdf40d69e8487dbfb253a34ad/c2corg_api/views/document_revert.py#L83 I have then omitted the version check done when usually making an update.
But then the system expects up-to-date ``version`` values else we get integrity errors existing values are used when trying to save the old versions:
> sqlalchemy.exc.IntegrityError: (psycopg2.IntegrityError) duplicate key value violates unique constraint "uq_documents_locales_archives_document_id_version_lang"
DETAIL:  Key (version, document_id, lang)=(1, 6, en) already exists.
 [SQL: 'INSERT INTO guidebook.documents_locales_archives (version, title, summary, description, type, document_id, lang) VALUES (%(version)s, %(title)s, %(summary)s, %(description)s, %(type)s, %(document_id)s, %(lang)s) RETURNING guidebook.documents_locales_archives.id'] [parameters: {'description': '...', 'document_id': 6, 'lang': 'en', 'type': 'w', 'version': 1, 'summary': 'The highest point in Europe', 'title': 'Mont Blanc'}]

If I set the document's ``version`` by hand to the latest values, as in https://github.com/c2corg/v6_api/blob/056d1edeb98abb9bdf40d69e8487dbfb253a34ad/c2corg_api/views/document_revert.py#L110 SqlAlchemy tries to update the document (!?) but gets an error because the version already exists.

As for now I have not been able to figure out how to manage those ``version`` numbers in order to save an archive as the new document version. I have read that those numbers were automatically incremented by SqlAlchemy https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document.py#L262-L265

@tsauerwein would you have an idea? :)

tsauerwein:
> @tsauerwein would you have an idea? :)

I will try to take a look tomorrow!

asaunier:
One thing I don't understand is how SqlAlchemy automatically updates the ``version`` attribute as stated here https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document.py#L262-L265

In the first commits of this PR I have tried to:
* get the current version of the document: ``document``
* get the archive version of the document I want to revert to: ``document_in``
* update the current version with the archive: ``document.update(document_in)`` as in https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document.py#L252

Problems:
* ``version`` attributes of the archive document are outdated and cause integrity errors because entries with the given version already exist in the database.
* if I try to change the ``document_in``'s ``version`` attributes by hand to sync them with the ``version`` attributes of ``document`` (current version of the document), SqlAlchemy detects such changes and tries to update the entries in the table (and gets integrity errors because entries with the same ``version`` already exist). Anyway entries in the database should not be updated at this point but I have tried to use ``with DBSession.no_autoflush`` with no success.

With commit cded858 (that does not work either as for now) I have tried to use another approach:
* I instanciate a new document object (``new_document``), that SqlAlchemy does not know yet.
* I add properties to this object from ``document_in`` (the document version to revert to) and set the ``version`` attribute by hand
* eventually I update the document with my new object: ``document.update(new_document)``
See https://github.com/c2corg/v6_api/blob/cded858868df4c0a82c41165c8f640b155dea8b5/c2corg_api/views/document_revert.py#L111-L133

As for now I get errors because of ``null`` values. Probably because my code is incorrect and does not transfer all expected properties.

tsauerwein:
I quickly hacked something together to make your tests pass:
https://github.com/c2corg/v6_api/compare/revert-service-tmp

The idea is:
- Get the current document.
- Get the document at the given version for a locale.
- Update the current document with the old document.
- Before storing, set the versions back to the current versions. I think this was the point you were struggling with.
- Rest like with `put`.

In `put`, the versions of the document in the database and of the document that was passed in have to be the same. Otherwise it would mean that you are trying to update the document from an outdated version. But for the revert this is exactly what we want to do. So, we pretend that we are updating from the latest version, while we are actually using some old version.

I hope this makes sense.

asaunier:
Thanks @tsauerwein !
I have updated the PR with your commit and changes I have done based upon it.

For instance I have added a ``after_update`` callback that can be used depending on the document type to apply additional changes (for instance updating route titles after reverting changes to their main waypoint), as it is done for ``_put``:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document.py#L280-L281
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/waypoint.py#L231-L232

I guess I should do the same for the ``before_update`` callback, used for routes and outings:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/route.py#L133-L135
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/outing.py#L105-L107

As for now I have ignored it because it mainly relies on associations to update the default geometry. But it adds some more complexity and I wonder if this callback is really needed since associations are not changed when reverting!?

tsauerwein:
> As for now I have ignored it because it mainly relies on associations to update the default geometry.

When reverting to an older version, the default geometry of that old version would be used. If the associations have changed (and the route/outing has no `geom_detail`), the old default geometry might no longer be correct. But I am not sure if this is a case worth considering.

asaunier:
@arnaud-morvan @fjacon This PR about a document reverting service is ready to be reviewed.

In short:
* moderators (only) see a "restore this version" button when viewing an archived version of a document, see PR https://github.com/c2corg/v6_ui/pull/1669
* if the button is clicked a special ``documents/revert`` service is called with params ``document_id``, ``lang``, ``version_id``
* the service makes a few checks and then creates a new version of the document using the data (figures and/or locales for the passed ``lang`` and/or geometry) of the passed ``version_id`` (no data is removed). Associations are not updated.

Please review :)

asaunier:
A possible improvement to manage the (rare?) case where a route is reverted to a version whose main waypoint is no longer associated. See https://github.com/c2corg/v6_ui/pull/1669#issuecomment-307324376

c2corg_api/views/document_revert.py
```diff
@@ -0,0 +1,236 @@
...
+            filter(DocumentLocale.lang == lang). \
+            options(joinedload('geometry')). \
+            options(contains_eager(locales_type_eager, alias=locales_type))
+        return document_query.first()
+
+    def _get_after_update(self, document_type):
+        if document_type == WAYPOINT_TYPE:
+            return update_linked_route_titles
+        if document_type == ROUTE_TYPE:
+            return update_title_prefix
```
arnaud-morvan: Does not this handle the routes main waypoint title_prefix that you're saying is not handled ?

c2corg_api/views/document_revert.py
```diff
@@ -0,0 +1,236 @@
...
+            document_id, lang, version_id, archive_clazz, archive_locale_clazz)
+
+        old_versions = document.get_versions()
+        user_id = self.request.authenticated_userid
+
+        # update the document with the input document
+        document.update(document_in)
+
+        # now set back the versions to the versions of the current document
+        # because we basing our changes on the lastest version
```
tsauerwein: wording: "...we **are** basing..."

c2corg_api/views/document_revert.py
```diff
@@ -0,0 +1,236 @@
...
+        document.version = old_versions['document']
+        document.locales[0].version = old_versions['locales'][lang]
+        if document.geometry:
+            document.geometry.version = old_versions['geometry']
+
+        # TODO? support 'before_update' callbacks as in _put()
```
tsauerwein: I think it should be relatively save to skip `before_update`. It is only used to set the default geometry for routes and outings without geometry. The only inconsistency that could occur is when the associated waypoints and routes have changed between the latest and the *revert version*. If you leave a comment explaining this issue it should be good.

c2corg_api/views/document_revert.py
```diff
@@ -0,0 +1,236 @@
...
+                update_areas_for_document(document, reset=True)
+
+            if document.type != MAP_TYPE and UpdateType.GEOM in update_types:
+                update_maps_for_document(document, reset=True)
+
+            after_update = self._get_after_update(document_type)
+            if after_update:
+                after_update(document, update_types, user_id=user_id)
+
+            update_cache_version(document)
```
tsauerwein: You also have to notify ES:
https://github.com/c2corg/v6_api/blob/a300ced/c2corg_api/views/document.py#L299

tsauerwein: And I think you should also call `update_feed_document_update`.

c2corg_api/views/document_revert.py
```diff
@@ -0,0 +1,236 @@
...
+        except StaleDataError:
+            raise HTTPConflict('concurrent modification')
+
+        # when flushing the session, SQLAlchemy automatically updates the
+        # version numbers in case attributes have changed. by comparing with
+        # the old version numbers, we can check if only figures or only locales
+        # have changed.
+        (update_types, changed_langs) = document.get_update_type(old_versions)
+
+        if update_types:
```
tsauerwein: This part looks very similar to the code in `DocumentRest._put`. Would it be possible to refactor a bit?

c2corg_api/views/document_revert.py
asaunier: Yes it does: if the reverting of a route changes the main waypoint id, the route's ``title_prefix`` is updated accordingly (as expected). What I meant is that if the waypoint was deassociated in the meantime, it does not restore the association (the reverting tool changes only the document, not its associations). Then the route might be prefixed by the name of a WP that is not associated to the route.

asaunier:
Thanks @arnaud-morvan @tsauerwein for your reviews!
I will see if I can make the changes suggested by @tsauerwein

asaunier:
I have updated the PR to implement the suggested changes. Commit a8e02e0 factorizes the common code between document updating and document reverting.

PR #673, opened 2017-06-21
Add a disable_comments attribute to outings and xreports

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1515
Requires https://github.com/c2corg/v6_common/pull/57

### DIFF
--- a/alembic_migration/versions/ff894861149d_add_column_disable_comments.py
+++ b/alembic_migration/versions/ff894861149d_add_column_disable_comments.py
@@ -1,0 +1,41 @@
+"""Add column disable_comments
+
+Revision ID: ff894861149d
+Revises: bacd59c5806a
+Create Date: 2017-06-21 14:34:08.421642
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'ff894861149d'
+down_revision = 'bacd59c5806a'
+branch_labels = None
+depends_on = None
+
+def upgrade():
+    op.add_column(
+        'outings',
+        sa.Column('disable_comments', sa.Boolean()),
+        schema='guidebook')
+    op.add_column(
+        'outings_archives',
+        sa.Column('disable_comments', sa.Boolean()),
+        schema='guidebook')
+    op.add_column(
+        'xreports',
+        sa.Column('disable_comments', sa.Boolean()),
+        schema='guidebook')
+    op.add_column(
+        'xreports_archives',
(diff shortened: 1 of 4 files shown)


asaunier:
The unit tests keep failing with those changes, reporting that:
> .....................E....E..E..............................................................EEEEEEEEEE........................................................................................E/home/travis/build/c2corg/v6_api/.build/venv/lib/python3.4/site-packages/sqlalchemy/orm/scoping.py:106: SAWarning: At least one scoped session is already present.  configure() can not affect sessions that have already been created.
  warn('At least one scoped session is already present.

I have probably forgotten something when adding those simple boolean attributes ``disable_comments`` but I cannot figure out what :/

asaunier:
I think I have found what is going wrong. If I create an outing with the standard UI (no ``disable_comments``) attribute sent to the API, the API fails because:
> sqlalchemy.exc.IntegrityError: (psycopg2.IntegrityError) null value in column "disable_comments" violates not-null constraint

Probably I should
* either remove the ``nullable=False`` setting in the alembic script
* or add a ``disable_comments`` attribute to all outings/xreports test data in the unit tests (and also add the ``nullable=False`` statement to the outings/xreports models as well).

I guess option 1 is easier.

asaunier:
055cd15 implements option 1 above.

Not sure it's enough though, since 5 tests fail in my instance, saying for instance:
```
======================================================================
FAIL: test_put_success_lang_only (c2corg_api.tests.views.test_xreport.TestXreportRest)
  File "/home/alex/v6_api/c2corg_api/tests/views/test_xreport.py", line 588, in test_put_success_lang_only
    body, self.xreport1, user='moderator')
  File "/home/alex/v6_api/c2corg_api/tests/views/__init__.py", line 873, in put_success_lang_only
    self.assertEquals(body.get('version'), document.version)
nose.proxy.AssertionError: 2 != 1
```

asaunier:
> Not sure it's enough though, since 5 tests fail in my instance, saying for instance:

That was because of the ``server_default='false'`` I had added to the definition of the new attribute ``disable_comments`` => the default value was used by the editing unit tests (this attribute is not provided explicitely) and it made SqlAlchemy think a new version of the "figure" part of the document was needed, even though only locales were actually changed.

I have removed this default behaviour, since it is OK to have the field default to ``null`` (same effect actually), as it is already the case for other boolean attributes such as ``outing.partial_trip`` for instance.

The UI works fine combined to this new API version (tested in my instance).

The PR is ready for reviewing!

PR #674, opened 2017-06-23
Anonymous xreports

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1507 (approach 2)
Requires https://github.com/c2corg/v6_common/pull/58

TODO:
* [x] Add attribute ``anonymous`` to xreports
* [x] Use a dedicated user account when ``anonymous: true``
* [x] Tests


asaunier:
This PR is ready to be reviewed!

What it does:
* add an ``anonymous`` attribute to the xreport model
* add the possibility to create document anonymously: in that case the current user account is replaced by a given account (account 2 is "Camptocamp Association" in demo/prod databases). Any document type could be concerned even though I have enabled this feature for xreports only at the moment.

asaunier:
@arnaud-morvan Thanks for reviewing!

PR #675, opened 2017-07-10
Remove cable_car from public_transportation types

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1214

### DIFF
--- a/alembic_migration/versions/d4d360ef69bc_update_public_transportation_types.py
+++ b/alembic_migration/versions/d4d360ef69bc_update_public_transportation_types.py
@@ -1,0 +1,165 @@
+"""Update public_transportation_types
+
+Revision ID: d4d360ef69bc
+Revises: 6d64a7fbdb8b
+Create Date: 2017-07-10 16:17:32.295441
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+from alembic_migration.extensions import drop_enum
+from sqlalchemy.sql.expression import and_, any_, cast
+from sqlalchemy.sql.functions import func
+from sqlalchemy.sql.schema import Table, MetaData
+
+from c2corg_api.models.utils import ArrayOfEnum
+
+
+# revision identifiers, used by Alembic.
+revision = 'd4d360ef69bc'
+down_revision = '6d64a7fbdb8b'
+branch_labels = None
+depends_on = None
+
+def upgrade():
+    old_options = ('train', 'bus', 'service_on_demand', 'boat', 'cable_car')
+    new_options = ('train', 'bus', 'service_on_demand', 'boat')
+
+    old_type = sa.Enum(
+        *old_options, name='public_transportation_type', schema='guidebook')
+    new_type = sa.Enum(
+        *new_options, name='public_transportation_type_', schema='guidebook')
+    new_type.create(op.get_bind(), checkfirst=False)
+
+    # Create new column with temporary name
+    op.add_column(
+        'waypoints',
+        sa.Column('public_transportation_types_', ArrayOfEnum(new_type)),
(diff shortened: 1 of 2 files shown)


asaunier:
The alembic script does not work yet because:
```
sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) cannot cast type guidebook.public_transportation_type[] to guidebook.public_transportation_type_[]
LINE 1: ...ET lift_access=true, public_transportation_types_=CAST(array...
                                                             ^
 [SQL: "UPDATE guidebook.waypoints SET lift_access=%(lift_access)s, public_transportation_types_=CAST(array_remove(guidebook.waypoints.public_transportation_types, %(array_remove_1)s) AS guidebook.public_transportation_type_[]) WHERE guidebook.waypoints.waypoint_type = 'access' AND 'cable_car' = ANY (guidebook.waypoints.public_transportation_types)"] [parameters: {'lift_access': True, 'array_remove_1': 'cable_car'}]
```

Perhaps I have to remove the ``cable_car`` entries from the tables and only then set the new type instead of doing both at the same time.

asaunier:
> Perhaps I have to remove the cable_car entries from the tables and only then set the new type instead of doing both at the same time.

No more luck with 1d03a9c

And with no cast I get
```
sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) column "public_transportation_types_" is of type guidebook.public_transportation_type_[] but expression is of type guidebook.public_transportation_type[]
LINE 1: ...debook.waypoints SET public_transportation_types_=guidebook....
                                                             ^
HINT:  You will need to rewrite or cast the expression.
 [SQL: "UPDATE guidebook.waypoints SET public_transportation_types_=guidebook.waypoints.public_transportation_types WHERE guidebook.waypoints.waypoint_type = 'access'"]
```

alembic_migration/versions/d4d360ef69bc_update_public_transportation_types.py
```diff
@@ -0,0 +1,165 @@
...
+        })
+    )
+    op.execute(
+        archives.update(). \
+        where(archives.c.public_transportation_types != None). \
+        values({
+            'public_transportation_types_': cast(func.string_to_array(
+                archives.c.public_transportation_types_str, ','), ArrayOfEnum(new_type))
+        })
+    )
```
asaunier: Perhaps the 3 update queries above could be replaced by only 2 like this:
```
    op.execute(
        waypoints.update(). \
        where(waypoints.c.public_transportation_types != None). \
        values({
            'public_transportation_types_str': func.array_to_string(
                func.array_remove(waypoints.c.public_transportation_types, 'cable_car'), ',')
        })
    )
    op.execute(
        waypoints.update(). \
        where(waypoints.c.public_transportation_types != None). \
        values({
            'public_transportation_types_': cast(func.string_to_array(
                waypoints.c.public_transportation_types_str, ','), ArrayOfEnum(new_type))
        })
    )
```
=> we try to remove ``cable_car`` from all entries without testing first if ``waypoints.c.public_transportation_types`` contains it and directly make the ``array_to_string``.
No clue what is more efficient. Anyway the tool is run only once and concerns only ~2000 entries, so I don't think it's really important.

asaunier:
@arnaud-morvan The PR is ready to be reviewed.

What is does:
* main goal: remove type ``cable_car`` from the ``public_transportation_type`` datatype in postgresql
* create temporary cols in tables ``waypoints`` and ``waypoints_archives`` to perform the migration from the former type to the new one. Eventually the temp/old cols are removed.
* set ``lift_access: True`` for entries that previously add  ``cable_car`` in their ``public_transportation_types``

PR #676, opened 2017-07-18
Fix transfer encoding of sent emails

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1730

I have not figured out how to easily set the message encoding when defining the ``Message`` at https://github.com/c2corg/v6_api/blob/de9e3a730f299b5fcc98183edaee04de7027e1a5/c2corg_api/emails/email_service.py#L55-L60

but I have noticed it seems possible to do so for ``Attachment`` and that message bodies could be defined directly as ``Attachment`` (and are converted to ``Attachment`` anyway if not already the case):
https://github.com/Pylons/pyramid_mailer/blob/master/pyramid_mailer/message.py#L201-L216

This PR explicitely uses an ``Attachment`` object as body of the message and set the charset when creating this object.

@gberaudo I know it's been a while since you have written the email service code, but does this change make sense to you?

### DIFF
--- a/c2corg_api/emails/email_service.py
+++ b/c2corg_api/emails/email_service.py
@@ -1,7 +1,7 @@
 import transaction  # NOQA
 
 from pyramid_mailer import get_mailer
-from pyramid_mailer.message import Message
+from pyramid_mailer.message import Message, Attachment
 from functools import lru_cache
 from c2corg_common.attributes import default_langs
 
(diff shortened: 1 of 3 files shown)


asaunier:
Thanks @gberaudo !
Not sure the change is enough though, since the problem can still be reproduced on the demo with this branch, see https://github.com/c2corg/v6_ui/issues/1730#issuecomment-316112272

By the way would you have an idea why we have those weird replacements in the email bodies with hotmail? For instance ``https://www.demov6.camptocamp.org/auth`` becomes ``https://www.demov6.campto=3Damp.org/a=uth``

gberaudo:
Can this be linked to the `quoted-printable` encoding?

`=3D` looks like the quoted form of `=` (see https://fr.wikipedia.org/wiki/Quoted-Printable).
But it looks strange to have it at this place in the URL.

asaunier:
I have updated the PR and changed ``transfer_encoding`` to ``base64`` instead of the default ``quoted-printable``. I have also removed the UTF8 charset setting.
The branch may be tested on the demo. I have tested with gmail and hotmail address and the received messages are OK in both cases.

PR #682, opened 2017-08-18
Add SSO rest service

arnaud-morvan:
How does it works :

There id graphical interface to add new SSO keys, connect directly to the postgresql server , example:
```
psql -d c2corg -c "INSERT INTO users.sso_key("domain", "key") VALUES ('pro-montagne', 'mykey');"
```

Some external service, with an SSO key to create/authenticate users in c2c domain.

Some external service named *example.com* send a POST request to `/sso_sync` with:
- sso_key: the_example_dot_com_sso_key
- external_id: 999
- email: newuser@external.domain.net
- username: newuser (mandatory only for not existing users)
- name: New User (defaults to username)
- forum_username: NewUser (defaults to username)
- lang: fr  (mandatory only for not existing users)

The C2C API server checks the SSO key and the source domain of the request (using DNS request).
It the server is authorized, it will search for the user by `external domain` and `external_id`.
If not found, it will search for it by `email`.
If not found, it will create it, returning an error for already used `username` or `forum_username`.
If everything go fine, the C2C server returns an authentication url with a token valid for only this user and for the next 10 minutes.

The external service server insert, in response to the browser, an iframe to that url.

So the browser will send a GET request to `/sso_login`.

The C2C api server will check the token and authenticate the user, returning a redirection the forum.

The browser send a request to the forum.

The browser is now authenticated everywhere.

Remaining TODO:
* Put the documentation somewhere ? I suppose in the wiki.
* check that the sso_login request has the sso_key.domain as referer ?

### DIFF
--- a/c2corg_api/views/sso.py
+++ b/c2corg_api/views/sso.py
@@ -1,0 +1,281 @@
+from base64 import b64encode
+from datetime import datetime, timedelta
+from os import urandom
+from pytz import utc
+from urllib.parse import urlencode
+import logging
+
+import colander
+from cornice.resource import resource, view
+from cornice.validators import colander_body_validator
+from pydiscourse.exceptions import DiscourseClientError
+from pyramid.httpexceptions import (
+    HTTPForbidden,
+    HTTPFound,
+    HTTPInternalServerError,
+)
+
+from c2corg_common.attributes import default_langs
+
+from c2corg_api.models import DBSession
+from c2corg_api.models.document import DocumentLocale
+from c2corg_api.models.sso import (
+    SsoKey,
+    SsoExternalId,
+)
+from c2corg_api.models.user import User
+from c2corg_api.models.user_profile import UserProfile
+from c2corg_api.security.discourse_client import get_discourse_client
+from c2corg_api.security.roles import log_validated_user_i_know_what_i_do
+from c2corg_api.views import (
+    cors_policy,
+    json_view,
+)
+from c2corg_api.views.user import (
+    validate_unique_attribute,
+    validate_forum_username,
+)
+
+CONST_EXPIRE_AFTER_MINUTES = 10
+
+log = logging.getLogger(__name__)
+
+
+class SsoSyncSchema(colander.MappingSchema):
+    sso_key = colander.SchemaNode(colander.String())
+    external_id = colander.SchemaNode(colander.String())
+    email = colander.SchemaNode(colander.String(),
+                                missing=None)
+    username = colander.SchemaNode(colander.String(),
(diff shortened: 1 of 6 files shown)


arnaud-morvan:
Note that for now I check that the sso_sync request client (pro-montagne server) has the IP address given by DNS for that domain. This may be blocking in case it is behind a reverse proxy. We may had a list of authorized hosts for each SSO key, or disable the test. This should be tested behind HAProxy in production infrastructure.

c2corg_api/views/sso.py
```diff
@@ -0,0 +1,282 @@
...
+                               validator=colander.OneOf(default_langs),
+                               missing=None)
+    groups = colander.SchemaNode(colander.String(),
+                                 missing=None)
+
+sso_sync_schema = SsoSyncSchema()
+
+
+def sso_sync_validator(request, **kwargs):
+    if 'sso_key' not in request.json:
```
asaunier: Shouldn't you rather manipulate ``request.validated`` instead of ``request.json``? See for instance https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document_merge.py#L29

c2corg_api/views/sso.py
```diff
@@ -0,0 +1,282 @@
...
+        """
+        request = self.request
+        sso_key = request.sso_key
+        sso_external_id = request.sso_external_id
+        user = request.sso_user
+
+        if user is None:
+            # create new user
+            user = User(
+                username=request.json['username'],
```
asaunier: ``request.json`` => ``request.validated``

(Actually I think you should do the change for all occurences of ``request.json`` in this file)

c2corg_api/views/sso.py
```diff
@@ -0,0 +1,282 @@
...
+            colander_body_validator,
+            sso_sync_validator,
+        ])
+    def post(self):
+        """
+        Synchronize user details and return authentication url.
+        Important: Email addresses need to be validated by external site.
+        """
+        request = self.request
+        sso_key = request.sso_key
```
asaunier: Don't you need to get the attributes from ``request.json`` or ``request.validated``?

asaunier:
@arnaud-morvan Thanks for the changes you made after my comments.
It looks good to me now, except that there are a few formatting problems reported by Travis:
> c2corg_api/views/sso.py:80:80: E501 line too long (80 > 79 characters)
c2corg_api/views/sso.py:99:80: E501 line too long (93 > 79 characters)
c2corg_api/tests/views/test_sso.py:250:9: F841 local variable 'body' is assigned to but never used

PR #686, opened 2017-08-23
Use 8bit transfer_encoding for sent emails

asaunier:
Related to https://github.com/c2corg/v6_ui/issues/1763


asaunier:
After testing we have figured out that:
* emails to free.fr address are still not delivered
* "=" problems in messages to hotmail addresses are back

=> not a solution.

Issue #683, opened 2017-08-23
Outing's information: add GPS attribute


desnoes:
Add a boolean "has_track" attribute set to true when the `geom_detail` is set (no need to transfer the whole geometry when listing cards).

desnoes:
see also : https://github.com/c2corg/v6_ui/issues/1549

asaunier:
Extending this to route documents (actually: any document type with a ``geom_detail`` attribute) would make sense as well, and probably won't add extra work. That's what I did in PR #687

PR #687, opened 2017-08-25
[WIP] Add geom.has_geom_detail attribute

asaunier:
Related to https://github.com/c2corg/v6_api/issues/683

Just for the recorsd because it does not work yet since
```
DocumentGeometry.has_geom_detail = column_property(
    DocumentGeometry.geom_detail is not None,
    deferred=True)
```
does not work as expected and causes the following error:
```
    DocumentGeometry.geom_detail is not None)
  File "<string>", line 2, in column_property
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/sqlalchemy/orm/properties.py", line 120, in __init__
    self._orig_columns = [expression._labeled(c) for c in columns]
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/sqlalchemy/orm/properties.py", line 120, in <listcomp>
    self._orig_columns = [expression._labeled(c) for c in columns]
  File "<string>", line 1, in <lambda>
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/sqlalchemy/sql/elements.py", line 4144, in _labeled
    return element.label(None)
AttributeError: 'bool' object has no attribute 'label'
```

(ditto with ``deferred=True``)

``column_property`` seems to expect ``Column`` parameter and not support a boolean:
http://docs.sqlalchemy.org/en/latest/orm/mapping_columns.html#sqlalchemy.orm.column_property


asaunier:
I have tried for instance to add the following to my model class:
```
    @declared_attr
    def has_geom_detail(self):
        return column_property(self.geom_detail)
```
=> this works: a "has_geom_detail" attribute is returned containing the value of "geom_detail".

Now if I try to get a boolean result:
```
    column_property(self.geom_detail is not None)
```
causes an error:
AttributeError: 'bool' object has no attribute 'label'
=> it seems column_property's initial arguments are expected to be columns.

On the other hand
```
    column_property(self.geom_detail != None)
```
causes no error, but no attribute is return though.

No more luck with a cast:
```
    column_property(cast(self.geom_detail), Boolean))
```
=> "cannot cast type geometry to boolean"

In SQL the following statement seems to return the expected results:
```
select case when geom_detail is not null then True else False end from documents_geometries limit 10;
```
but the matching SqlAlchemy syntax I have written:
```
    column_property(case([(self.geom_detail != None, True)], else_=False))
```
(see http://docs.sqlalchemy.org/en/latest/core/sqlelement.html?highlight=case#sqlalchemy.sql.expression.case ) causes no error but no attribute is returned either.

asaunier:
Commit bd80b9e does not provide a solution but at least a new "has_geom_detail" attribute is returned (at the moment containing "geom_detail" instead of a boolean). Eg.
http://c2corgv6-demo.gis.internal:6545/outings
```
...
geometry": {
        "geom": "{\"type\": \"Point\", \"coordinates\": [736653.5809197543, 5863845.5846177135]}",
        "version": 1,
        "has_geom_detail": "{\"type\": \"LineString\", \"coordinates\": [[736384.6787817676, 5863351.873003268], [736585.3259810163, 5864087.579400512], [736795.5278088005, 5863342.3183747325], [737091.7212934055, 5863734.058144693]]}"
      },
...
```
The thing is now to return a boolean instead but I have not figured out how to do it yet.
Either I get errors because the arguments for ``column_property`` are not valid or the attribute is not provided in the JSON output (with no explicit reason).

asaunier:
Using 
```
DocumentGeometry.has_geom_detail = column_property(
    DocumentGeometry.geom_detail)
```
instead of 
```
@declared_attr
def has_geom_detail(self):
    return column_property(self.geom_detail)
```
has exactly the same output in the JSON and changing it to 
```
DocumentGeometry.has_geom_detail = column_property(
    DocumentGeometry.geom_detail != None)
```
aslo prevent the attribute from being returned in the JSON output (but with no explanation).

asaunier:
Recap:
* Using geom_detail attribute as is
```
DocumentGeometry.has_geom_detail = column_property(
    DocumentGeometry.geom_detail)
```
=> OK: 
```
"geometry": {
        "version": 1,
        "geom": "{\"coordinates\": [736653.5809197543, 5863845.5846177135], \"type\": \"Point\"}",
        "has_geom_detail": "{\"coordinates\": [[736384.6787817676, 5863351.873003268], [736585.3259810163, 5864087.579400512], [736795.5278088005, 5863342.3183747325], [737091.7212934055, 5863734.058144693]], \"type\": \"LineString\"}"
      },
```
* Standard geom attribute:
```
DocumentGeometry.has_geom_detail = column_property(
    DocumentGeometry.geom)
```
=> OK:
```
"geometry": {
        "geom": "{\"coordinates\": [736653.5809197543, 5863845.5846177135], \"type\": \"Point\"}",
        "has_geom_detail": "{\"coordinates\": [736653.5809197543, 5863845.5846177135], \"type\": \"Point\"}",
        "version": 1
      },
```
* Using "IS NOT NULL" test:
```
DocumentGeometry.has_geom_detail = column_property(
    DocumentGeometry.geom_detail != None)
```
=> KO (no error but no ``has_geom_detail`` attribute)
```
"geometry": {
        "version": 1,
        "geom": "{\"type\": \"Point\", \"coordinates\": [736653.5809197543, 5863845.5846177135]}"
      },
```
* Using a postgis transformation as in https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/document.py#L404-L406
```
DocumentGeometry.has_geom_detail = column_property(
    func.ST_Transform(DocumentGeometry.geom_detail, 4326))
```
=> KO:
```
"geometry": {
        "version": 1,
        "geom": "{\"type\": \"Point\", \"coordinates\": [736653.5809197543, 5863845.5846177135]}"
      },
```
* Using attribute ``version``:
```
DocumentGeometry.has_geom_detail = column_property(
    DocumentGeometry.version)
```
=> KO:
```
"geometry": {
        "version": 1,
        "geom": "{\"type\": \"Point\", \"coordinates\": [736653.5809197543, 5863845.5846177135]}"
      },
```
* Using the document id:
```
DocumentGeometry.has_geom_detail = column_property(
    DocumentGeometry.document_id)
```
=> OK:
```
"geometry": {
        "geom": "{\"coordinates\": [736653.5809197543, 5863845.5846177135], \"type\": \"Point\"}",
        "version": 1,
        "has_geom_detail": 842194
      },
```
(OK with the ``document_id``, KO with ``version`` - both are integers)

I have tried adding ``__add_properties__ = ('has_geom_detail')`` as suggested by @pvalsec to the ``DocumentGeometry`` model but it has no good/bad effect.

asaunier:
Perhaps a syntax that could help:
https://github.com/c2corg/v6_api/blob/9f97ed5b643e6097680e767876cfd1b70c86f3bb/c2corg_api/models/feed.py#L57-L62
```
User.has_area_filter = column_property(
    select([func.count(FilterArea.area_id) > 0]).
    where(FilterArea.user_id == User.id).
    correlate_except(FilterArea),
    deferred=True
)
```

tsauerwein:
It seems that ColanderAlchemy does not recognize `column_property` as a proper column. We could use a workaround as proposed in my last fixup commit.

asaunier:
> It seems that ColanderAlchemy does not recognize column_property as a proper column. We could use a workaround as proposed in my last fixup commit.

Thanks @tsauerwein for the workaround! It works :)

Just one thing: the new ``has_geom_detail`` attribute is now made available for all documents with a geometry, both when listing or viewing a document, not only for routes/outings (only document types with an actual ``geom_detail`` attribute). The "v6_common" config https://github.com/c2corg/v6_common/pull/60 seems to be ignored.

I guess https://github.com/c2corg/v6_api/blob/5bf5108158a7f7e4aed5ff54295d27a24e483c9e/c2corg_api/views/__init__.py#L131-L136 adds the new attribute to any document with a ``geometry``.

No big deal but should we:
* cancel the ``v6_common`` change (useless?)?
* add a test in  https://github.com/c2corg/v6_api/blob/5bf5108158a7f7e4aed5ff54295d27a24e483c9e/c2corg_api/views/__init__.py#L131-L136 to add the attribute only when ``geom_detail`` exists?

c2corg_api/views/__init__.py
```diff
@@ -128,12 +128,15 @@ def to_json_dict(obj, schema, with_special_locales_attrs=False):
...
-    if 'geometry' in obj_dict and hasattr(obj, 'geometry'):
-        geometry_dict = obj_dict['geometry']
-        geometry = obj.geometry
-        for attr in geometry_special_attributes:
-            if hasattr(geometry, attr):
-                geometry_dict[attr] = getattr(geometry, attr)
+
+    #geometry attribute has_geom is only required for outings:'o' and routes:'r'
+    if 'geometry' in obj_dict and hasattr(obj, 'geometry') \
+        and (obj.type == 'o' or obj.type=='r'):
```
asaunier: You can rather use ``OUTING_TYPE`` and ``ROUTE_TYPE`` here, with
```
from c2corg_api.models.outing import OUTING_TYPE
from c2corg_api.models.route import ROUTE_TYPE
```

julsbreakdown:
Yes Indeed thank you for the info

tsauerwein:
> Just one thing: the new has_geom_detail attribute is now made available for all documents with a geometry, both when listing or viewing a document, not only for routes/outings (only document types with an actual geom_detail attribute). The "v6_common" config c2corg/v6_common#60 seems to be ignored.

Good point! I haven't thought of that. The check `if hasattr(geometry, attr)` always loads the `has_geom_detail` attribute.

We could do the following:

- Keep `geometry.has_geom_detail` in common, because the attributes are also used to generate the query. If the attribute is not included in the query to load all documents, n queries would have to be made to load the values for each document (to verify).
- In [to_json_dict](https://github.com/c2corg/v6_api/blob/8844ae9/c2corg_api/views/__init__.py#L105) add a parameter `with_special_geometry_attrs=False` and call the function [here](https://github.com/c2corg/v6_api/blob/8844ae9dc72faa083e608cd8ecadbc46f859ab88/c2corg_api/views/document_listings.py#L140) with the parameter set to `True`.

c2corg_api/models/utils.py
```diff
@@ -13,7 +13,10 @@ def copy_attributes(obj_from, obj_to, attributes):
     Copies the given attributes from `obj_from` to `obj_to` (shallow copy).
     """
     for attribute in attributes:
-        if hasattr(obj_from, attribute):
+        # has_geom_attribute isnt a real column in a table
+        # it is only defined in the models
+        # we avoid looking for it here making travis break
```
tsauerwein: Do the tests pass locally, but not on Travis?

julsbreakdown: sorry by travis I mean all the tests. 
They broke locally and on travis before

tsauerwein: `copy_attributes` is called in `DocumentGeometry.to_archive/update`:
https://github.com/c2corg/v6_api/blob/8844ae9dc72faa083e608cd8ecadbc46f859ab88/c2corg_api/models/document.py#L340-L346

Any reason `has_geom_detail` must be listed in `DocumentGeometry._ATTRIBUTES`?

julsbreakdown: I don't see any reason indeed, i will removed it.

julsbreakdown:
I've applied the changes but now we are losing the linestring info (geom_detail), it is no longer displayed on the map / available for download. See the print screen 
capture du 2017-09-21 15-03-40
capture du 2017-09-21 15-04-44

requirements.txt
```diff
@@ -21,7 +21,7 @@ python-json-logger==0.1.5
 redis==2.10.5
 requests==2.11.1
 setuptools==28.8.0
-Shapely==1.5.17
+Shapely==1.6.1
```
asaunier: Just curious: why do we need to upgrade Shapely?

c2corg_api/views/__init__.py
```diff
@@ -126,7 +128,13 @@ def to_json_dict(obj, schema, with_special_locales_attrs=False):
             for attr in locale_special_attributes:
                 if hasattr(locale, attr):
                     locale_dict[attr] = getattr(locale, attr)
-
+    if (with_special_geometry_attrs and obj.type in(ROUTE_TYPE, OUTING_TYPE)):
```
asaunier: Do you really need the parenthesis for the condition? A blank space after ``in`` is recommended I think.
Perhaps you could write:
```
if with_special_geometry_attrs and obj.type in (ROUTE_TYPE, OUTING_TYPE):
```

c2corg_api/models/utils.py
```diff
@@ -13,7 +13,7 @@ def copy_attributes(obj_from, obj_to, attributes):
     Copies the given attributes from `obj_from` to `obj_to` (shallow copy).
     """
     for attribute in attributes:
-        if hasattr(obj_from, attribute):
+        if hasattr(obj_from, attribute) and attribute:
```
tsauerwein: Why is the `... and attribute` needed?

c2corg_api/views/document_schemas.py
```diff
@@ -105,7 +105,9 @@ def get_all_fields(fields, activities, field_list_type):
     """
     fields_list = [
         fields.get(activity).get(field_list_type) for activity in activities
+        if fields.get(activity)
```
tsauerwein: Why is this new check required?

asaunier:
I have tested the branch in my instance and as far as I can see it works really good :+1: 

A few comments/questions:
* Before merging the PR it would be nice to squash the commits, for instance using ``git rebase -i --autosquash e8984e0f357e80f81dcd2f37f08472cf4dd86995``. You can also do it when merging the PR (option "Squash and merge" in the dropdown menu right near the "Merge pull request" button) but it's a good practise to do it before merging I think.
*  You have cancelled the changes in the v6_common part. Then I get we can close this PR https://github.com/c2corg/v6_common/pull/60 On the other hand I think @tsauerwein has recommended to keep ``geometry.has_geom_detail`` in v6_common's listing config, to prevent making additional queries, see https://github.com/c2corg/v6_api/pull/687#issuecomment-330884983
* It would probably make sense to add a few simple unit tests. For instance in https://github.com/c2corg/v6_api/blob/master/c2corg_api/tests/views/test_route.py you could check that the list actually have ``has_geom_detail`` attributes, and that it's true for ``self.route``. Test data are set in https://github.com/c2corg/v6_api/blob/master/c2corg_api/tests/views/test_route.py#L1170 and listing tests can be done here https://github.com/c2corg/v6_api/blob/master/c2corg_api/tests/views/test_route.py#L37 I think.

requirements.txt
julsbreakdown: Shapely was giving an error with the travis build, Changing the version made the build pass

c2corg_api/views/__init__.py
julsbreakdown: Not need, I added it before because it was a multiline condition, I will remove it

c2corg_api/views/document_schemas.py
julsbreakdown: I was getting AttributeError: 'NoneType' object has no attribute 'get', for the paragliding activity

c2corg_api/models/utils.py
julsbreakdown: Not required anymore

c2corg_api/views/document_schemas.py
tsauerwein: Do you still have the test-case that produced the error? I would like to avoid that this fix is hiding another problem.

julsbreakdown: Simply by going to http://0.0.0.0:6553/routes. I think the same happens here http://c2corgv6-demo.gis.internal:6555/routes. I havent got access to this server ATM, so I cant be sure

tsauerwein: Ok, I see. `paragliding` is not listed in [c2corg_common/fields_route.py](https://github.com/c2corg/v6_common/blob/9a7f0cb814f88a4f7bdbd7443db6cc2e4ada2641/c2corg_common/fields_route.py). But if I remember correctly, there are no paragliding routes, @asaunier?

So, this might be an issue with your test data.

julsbreakdown: The data I use is a dump from june 2017.
I will see with Alex (he isn't working today)

asaunier: @tsauerwein Indeed, there are no paragliding routes, only paragliding outings. Because with a paraglider there is no predefined trajectories, you go up using an hiking/climbing/skitouring/... route and you go down by the airs.
When creating/editing a route you cannot select the paragliding activity (but you can for outings).

asaunier: http://c2corgv6-demo.gis.internal:6555/routes is my instance on the dev server, it's not a good idea to make tests on it since it might be broken because of work in progress, in addition the ES indexing is a bit broken, I have not spend time fixing it yet.

julsbreakdown:
I think everything is covered => merging

c2corg_api/views/document_schemas.py
tsauerwein: Were you able to find out why there was a route with paragliding as an activity?

asaunier: > Were you able to find out why there was a route with paragliding as an activity?

There are probably none. In my dev instance (same data than @julsbreakdown 's?) or in prod:
```
c2corg=# select count(*) from guidebook.routes where 'paragliding' = any(activities);
 count 
-------
     0
(1 row)
```

Issue #684, opened 2017-08-23
Outing's information: add associated photos attribute


desnoes:
Add an attribute for counting the number of photos associated to the outing. This number will be used by UI to display the number of photos in the outing's card.

asaunier:
Probably a duplicate of https://github.com/c2corg/v6_api/issues/338

asaunier:
It would make sense to extend this data to any document type, not only outings. It probably don't add extra work.

asaunier:
The issue about the ratings in the outings lists should probably be considered along with the current one. See https://github.com/c2corg/v6_api/issues/685

There are 2 main options:

## Option 1: "joins"

We add joins in the documents listing queries in order to also retrieve data from associated documents (a bit like the associated areas, already returned in the lists => TODO: investigate how the associated areas are managed in the doc lists), eg. number of associated images (nb of associated routes could be useful for WP lists, or nb of associated outings for routes lists?) or ratings of associated routes (for outings lists).

Pros:
* no change in the DB structure
* data are always up-to-date (cache?) even though new images are associated or removed

Cons:
* adds quite a lot of complexity to database queries?!
* impact on performances?

## Option 2 "stored metadata"

Metadata attribute are added to the documents table (but not to archived versions + no new version is created when the metadata are updated).

Pros:
* no additional complexity to code/queries when listing docs (only new attributes must be added to the list of "listing attributes", eg. https://github.com/c2corg/v6_common/blob/master/c2corg_common/fields_outing.py#L43)
* not much impact on performances

Cons:
* requires changes in the database structure, especially in the Document model https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/document.py#L70 (or in specific models for document type specific data such as outing ratings). TODO: make sure metadata are not lost when editing a document
* triggers must be set up in order to update the stored metadata when associated docs changes (eg. added or removed images, updated rating in an associated route)

asaunier:
Concerning option 1
> TODO: investigate how the associated areas are managed in the doc lists),

Joins are added here:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document_listings.py#L103-L113

See also:
* https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document_listings.py#L122-L123
* https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document_listings.py#L189-L197
and
* https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/document.py#L154-L155
[...]

PR #689, opened 2017-09-12
Add associated images counter in outings lists

asaunier:
Related to https://github.com/c2corg/v6_api/issues/684

The chosen solution is to make an extra DB query to count the number of images associated to the listed documents (eg. in the advanced search pages or in associations section of document pages). The counts are then added to the listed documents objects before being returned by the API.

As for now, **the ``img_count`` attribute is available only in outings lists** but could be added for other document types if needed (currently disabled for those types since not asked by #684 and to avoid the additional DB query if not required).

Please note that ``img_count`` is available also for outings in the feed service, even though it is not really needed since the feed already returns a subset of associated images if any.
It would be possible to remove the ``img_count`` attribute in the feed (for instance to make sure the extra DB request is not done) by adding
```
document_config.include_img_count = False
```
after https://github.com/c2corg/v6_api/blob/master/c2corg_api/views/feed.py#L381
The problem is that listed items are cached and reused both in the standard documents lists and in the feed items. Which mean that an outing document might have or not have the ``img_count`` attribute whether called first in the feed or in the standard lists. I have then preferred to keep the attribute in both cases.

### DIFF
--- a/c2corg_api/views/document_listings.py
+++ b/c2corg_api/views/document_listings.py
@@ -1,10 +1,12 @@
 from c2corg_api.caching import cache_document_listing
 from c2corg_api.models import DBSession
 from c2corg_api.models.area import schema_listing_area
+from c2corg_api.models.association import Association
 from c2corg_api.models.cache_version import get_document_id, \
     get_cache_keys
 from c2corg_api.models.document import (
     set_available_langs)
+from c2corg_api.models.image import IMAGE_TYPE
 from c2corg_api.models.outing import Outing
 from c2corg_api.models.xreport import Xreport
 from c2corg_api.models.user import User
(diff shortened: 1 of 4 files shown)


c2corg_api/views/document_listings.py
```diff
@@ -195,3 +201,16 @@ def _set_areas_for_documents(documents, lang):
...
             to_json_dict(m, schema_listing_area) for m in document._areas
         ]
+
+
+def _set_img_count_for_documents(documents, document_ids):
+    res = DBSession.query(
+        Association.parent_document_id, func.count('*')). \
+        filter(and_(
+            Association.parent_document_id.in_(document_ids),
+            Association.child_document_type == 'i')). \
```
tsauerwein: You could use `IMAGE_TYPE`.

asaunier: Thanks @tsauerwein !

PR #690, opened 2017-09-13
Add a langs filter to the activity feed

asaunier:
Related to https://github.com/c2corg/v6_api/issues/608
Replace https://github.com/c2corg/v6_api/pull/642

UI part issue https://github.com/c2corg/v6_ui/issues/1382 and PR https://github.com/c2corg/v6_ui/pull/1789

TODO
* [x] Add langs to user preferences
* [x] Fix ``langs`` attributes definitions (enum or foreign key?)
* [x] Add a ``langs`` attribute to the feed table
* [x] Filter feed service output according to the langs pref 
* [x] Unit tests

### DIFF
--- a/c2corg_api/models/feed.py
+++ b/c2corg_api/models/feed.py
@@ -5,7 +5,8 @@ from c2corg_api.models.area import Area, AREA_TYPE
 from c2corg_api.models.area_association import AreaAssociation
 from c2corg_api.models.article import ARTICLE_TYPE
 from c2corg_api.models.association import Association
-from c2corg_api.models.document import Document, UpdateType
+from c2corg_api.models.book import BOOK_TYPE
+from c2corg_api.models.document import Document, DocumentLocale, UpdateType
 from c2corg_api.models.enums import feed_change_type
 from c2corg_api.models.image import Image, IMAGE_TYPE
 from c2corg_api.models.outing import OUTING_TYPE
@@ -13,12 +14,13 @@ from c2corg_api.models.route import ROUTE_TYPE
 from c2corg_api.models.user import User
 from c2corg_api.models.user_profile import USERPROFILE_TYPE
 from c2corg_api.models.utils import ArrayOfEnum
+from c2corg_api.models.xreport import XREPORT_TYPE
 from c2corg_api.views.validation import association_keys
 from sqlalchemy.dialects import postgresql
 from sqlalchemy.dialects.postgresql.array import ARRAY
 from sqlalchemy.orm import relationship, column_property
 from sqlalchemy.sql.elements import literal_column
-from sqlalchemy.sql.expression import select, text
+from sqlalchemy.sql.expression import select, text, cast
 from sqlalchemy.sql.functions import func
 from sqlalchemy.sql.schema import Column, ForeignKey, PrimaryKeyConstraint, \
     CheckConstraint, Index
@@ -137,6 +139,10 @@ class DocumentChange(Base):
     activities = Column(
         ArrayOfEnum(enums.activity_type), nullable=False, server_default='{}')
 
+    # langs of the document locales
+    langs = Column(
+        ArrayOfEnum(enums.lang), nullable=False, server_default='{}')
+
     # For performance reasons, areas and users are referenced in simple integer
     # arrays in 'feed_document_changes', no PK-FK relations are set up.
     # To prevent inconsistencies, triggers are used.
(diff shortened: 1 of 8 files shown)


alembic_migration/versions/8c230a4a0284_add_column_feed_filter_langs_to_users.py
```diff
@@ -0,0 +1,36 @@
...
+
+def upgrade():
+    langs = ('fr', 'it', 'de', 'en', 'es', 'ca', 'eu')
+    lang_enum = sa.Enum(*langs, name='lang', schema='guidebook')
+    lang_enum.create(op.get_bind(), checkfirst=False)
+    op.add_column(
+        'user',
+        sa.Column(
+            'feed_filter_langs',
+            ArrayOfEnum(lang_enum),
```
tsauerwein: Elsewhere `lang` is a foreign key to [guidebook.langs](https://github.com/c2corg/v6_api/blob/db2dc7412001d5d60e62f97272177db6534e0138/alembic_migration/versions/38df9393c9a9_init.py#L629):
https://github.com/c2corg/v6_api/blob/db2dc7412001d5d60e62f97272177db6534e0138/alembic_migration/versions/38df9393c9a9_init.py#L730-L732

asaunier: @tsauerwein Thanks! I have made a change https://github.com/c2corg/v6_api/pull/690/commits/f161373298fed122e534c296b14277fe3af145d8

asaunier: But I am not sure it is really possible to use such a foreign key for an ARRAY typed column.
With the current definition
```
feed_filter_langs = Column(
￼    ARRAY(String(2), ForeignKey(schema + '.langs.lang')),
    nullable=False, server_default='{}')
```
the foreign key is ignored. Most likely because it is not a supported argument for ``ARRAY()``:
http://docs.sqlalchemy.org/en/latest/core/type_basics.html#sqlalchemy.types.ARRAY

If I move the foreign key definition outside of ``ARRAY()``, I get errors such as 
> sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) foreign key constraint "feed_document_changes_langs_fkey" cannot be implemented
DETAIL:  Key columns "langs" and "lang" are of incompatible types: character varying[] and character varying.
 [SQL: 'ALTER TABLE guidebook.feed_document_changes ADD FOREIGN KEY(langs) REFERENCES guidebook.langs (lang)']

tsauerwein: Oh yes, sorry! I didn't notice it's an array.

The cleanest would probably be to use an enum everywhere, but that doesn't have to be done in this PR.

asaunier:
1d15e5c adds the update of the new ``langs`` attibute in table ``feed_document_changes`` to the alembic script. It directly uses an SQL query as ``op.execute()`` argument instead of sqlalchemy functions as done for instance in https://github.com/c2corg/v6_api/blob/master/alembic_migration/versions/d4d360ef69bc_update_public_transportation_types.py#L57-L79

But it works and seems quite simpler. Any objection?

Please note the query is a bit long to run (~90 sec on the dev server) because it updates > 226000 entries.

asaunier:
The PR is ready to be reviewed. @julsbreakdown perhaps you could have a look? Thanks :)

c2corg_api/models/feed.py
```diff
@@ -199,9 +206,12 @@ def update_feed_document_create(document, user_id):
     DBSession.flush()
 
     activities = []
-    if document.type in [ARTICLE_TYPE, OUTING_TYPE, ROUTE_TYPE]:
+    if document.type in [ARTICLE_TYPE, OUTING_TYPE, ROUTE_TYPE,
+                         BOOK_TYPE, XREPORT_TYPE]:
```
tsauerwein: :+1:

asaunier:
Thanks @tsauerwein @julsbreakdown for your reviews!

asaunier:
I have updated the demo with changes about the lang filtering of the feed (+has_geom_detail etc.) http://www.demov6.camptocamp.org/

The filtering seems to more or less work except that some unrelated cards are sometimes shown (especially when the filters are quite restrictive). That's weird because this behaviour is not observed in my dev instance. Could it be for instance a caching problem?

PR #694, opened 2017-10-02
Remove package postgresql-9.4-postgis-scripts from travis

arnaud-morvan:


arnaud-morvan:
Minor fix for travis to be green in current pull requests, merging.

Issue #692, opened 2017-09-26
Protected then unprotected articles


desnoes:
When an article is protected and then unprotected it is not possible to save a modification. If the category is modified (eg. add a new category) then the problem disappears.

asaunier:
Here are the logs when trying to edit the title of a route formerly protected then unprotected:
```
2017-10-27 16:22:29,716 ERROR [c2corg_api.views][waitress] Unexpected error: No row was found for one()
Traceback (most recent call last):
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/tweens.py", line 22, in excview_tween
    response = handler(request)
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid_tm/__init__.py", line 109, in tm_tween
    reraise(*exc_info)
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid_tm/compat.py", line 15, in reraise
    raise value
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid_tm/__init__.py", line 88, in tm_tween
    response = handler(request)
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/router.py", line 158, in handle_request
    view_name
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/view.py", line 547, in _call_view
    response = view_callable(context, request)
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/config/views.py", line 182, in __call__
    return view(context, request)
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/viewderivers.py", line 393, in attr_view
    return view(context, request)
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/viewderivers.py", line 371, in predicate_wrapper
    return view(context, request)
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/viewderivers.py", line 354, in _authdebug_view
    return view(context, request)
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/viewderivers.py", line 302, in _secured_view
    return view(context, request)
  File "/home/alex/v6_api/.build/venv/lib/python3.4/site-packages/pyramid/viewderivers.py", line 442, in rendered_view
    result = view(context, request)
[...]

PR #696, opened 2017-10-12
[WIP] Define ratings attributes for outings

julsbreakdown:
#685

PR #699, opened 2017-10-27
Add missing archive version when (un)protecting a document

asaunier:
Related to https://github.com/c2corg/v6_api/issues/692

TODO:
- [x] fix unit tests
- [x] add an alembic script to fix versions of already impacted documents

### DIFF
--- a/c2corg_api/views/document_protect.py
+++ b/c2corg_api/views/document_protect.py
@@ -2,8 +2,9 @@ import logging
 
 from c2corg_api import DBSession
 from c2corg_api.models.cache_version import update_cache_version_direct
-from c2corg_api.models.document import Document
+from c2corg_api.models.document import Document, UpdateType
 from c2corg_api.views import cors_policy, restricted_json_view
+from c2corg_api.views.document import DocumentRest
 from colander import MappingSchema, SchemaNode, Integer, required
 from cornice.resource import resource
 from cornice.validators import colander_body_validator
@@ -47,7 +48,18 @@ class DocumentProtectRest(object):
         """
         document_id = self.request.validated['document_id']
         document = _get_document(document_id)
+
+        # Do nothing if document is already protected.
+        if document.protected:
+            return {}
+
         document.protected = True
+
+        user_id = self.request.authenticated_userid
+        DocumentRest.update_version(
+            document, user_id, 'Protected document',
+            [UpdateType.FIGURES], [])
+
         update_cache_version_direct(document_id)
 
         return {}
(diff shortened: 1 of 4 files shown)


tsauerwein:
Good analysis in https://github.com/c2corg/v6_api/issues/692#issuecomment-339986001, Alex! :+1: When `document.protected = True/False` is set, the version number is incremented. The fix you are proposing looks reasonable (unless if you want to avoid that a new version is created when protecting/unprotecting documents).

To fix up the documents in the database where the version number in `documents` is out of sync with the latest version in `documents_archives`, you could probably update the version of those documents in `documents` with the latest version. This *should* work.

asaunier:
Thanks @tsauerwein !

That looks indeed relevant to me to add a row in the archive tables when updating the ``protected`` flag: this way the flag values between the standard ``documents`` table and ``documents_archives`` are consistent and we can have the protect/unprotect actions in the history (that happens not that often so I don't think it will dramatically overweight the archives).

>  To fix up the documents in the database where the version number in documents is out of sync with the latest version in documents_archives, you could probably update the version of those documents in documents with the latest version. This should work.

It seems to work indeed and I have add an alembic script that reset the ``version`` attribute from ``documents`` to the max value of ``version`` in the archives.

I had also errors in the unit tests (testing what happens when protecting an already protected document or unprotecting an already unprotected document) 
> psycopg2.IntegrityError: duplicate key value violates unique constraint "uq_documents_archives_document_id_version"
DETAIL:  Key (version, document_id)=(1, 1799) already exists.

because my changes tried to add an archive entry anyway, which caused a duplicate since the version value was not incremented (no change detected). I have then added a test to skip the whole action in case the document is already protected (resp. unprotected).

c2corg_api/views/document_protect.py
```diff
@@ -47,7 +48,18 @@ def post(self):
         """
         document_id = self.request.validated['document_id']
         document = _get_document(document_id)
+
+        # Do nothing if document is already protected.
```
julsbreakdown: Is there a reason why we duplicate the code (DocumentProtectRest&DocumentUnprotectRest). Couldn't we call the same function, with extra params ?

asaunier: You're right. This code could be factorized. I prefer not doing it in this PR since the fix is not really ordered by the association (as far as I know), so I prefer not spending more time on it :)

asaunier:
Thanks for reviewing!
@julsbreakdown Do you think it is better to merge your PR https://github.com/c2corg/v6_api/pull/696 first ?
I guess one of our 2 PRs will have to be updated concerning the alembic references. Is it more relevant to run your alembic script first or this one?

julsbreakdown:
Merge first, as the rating migration takes longer, i think it would be better for this migration to be done before hand. I will change the alembic ref on the other PR.
So you can merge =)

PR #704, opened 2017-11-02
Merge alembic heads

arnaud-morvan:


arnaud-morvan:
@asaunier : this should fix travis on v6_ui

asaunier:
Thanks @arnaud-morvan !
Do we really need to add a new alembic script for this. Isn't this change enough: https://github.com/c2corg/v6_api/pull/703

arnaud-morvan:
Yes, changing the downversion of one migration it is enough, I don't mind about one or the other.

arnaud-morvan:
The merge should work for the ones who have already executed one of them.

asaunier:
OK, since the other PR is merged, I guess the current one is not needed then?

> The merge should work for the ones who have already executed one of them.

I think no one has run the alembic scripts again except Travis.

arnaud-morvan:
If the other one has been merged, this one is invalid now.

Issue #708, opened 2017-12-01
Anonymising a DB dump


asaunier:
To make dumps of the database publicly available (eg. for installing a dev API with real data), we need to write a few SQL queries that would remove/replace private data such as email addresses, usernames, passwords (even though already encrypted) etc.

asaunier:
The PR adding SQL queries to remove sensitive data from the DB has been merged.
Now we have to set a routing that uses those queries and make a special dump available somewhere.

brunobesson:
I still don't know the production infrastructure enough, but could this be done through a systemd timer + service on compose0? @mfournier?

PR #709, opened 2017-12-04
SQL script to prepare anonymized DB dumps

asaunier:
Related to https://github.com/c2corg/v6_api/issues/708

Questions before the PR can be reviewed:
- [x] what password to use for accounts? (username is now ``user<user_id>``) + make sure the value is the correctly encrypted one
- [x] what data should be purged for user profiles?

Other data to remove?

### DIFF
--- a/scripts/anonymize-data.sql
+++ b/scripts/anonymize-data.sql
@@ -1,0 +1,36 @@
+-- This file contains SQL queries used to remove personal data before exporting the database to a public dump
+
+-- Replace users'data by fake data. Except (real) name to preserve the authors of the guidebook contributions.
+-- password = c2c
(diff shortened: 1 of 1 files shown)


cbeauchesne:
> what data should be purged for user profiles?

I would say all but locales :  It would be conservative to purge everything, and later ask ourself the question if needed.

> what password to use for accounts? (username is now user<user_id>) + make sure the value is the correctly encrypted one

'c2corg' ?

brunobesson:
>Questions before the PR can be reviewed

I agree with @cbeauchesne answer: the profiles should be emptied. Or replaced with a default content.
As for the password, any default password would be fine. Either `c2corg` or something similar, or a password identical to the username.

asaunier:
This PR is ready to be reviewed.

It removes all sensitive data: authentication tokens, sso stuff, mailing lists registrations, preferences, users' private data (emails, usernames, passwords, etc.) as well as profiles and profiles archives data.

The users' names (and id) remain though in order to preserve the authors of the documents.
Usernames are replaced by ``user<user_id>``, faked emails are used and passwords are all replaced by "c2c".

The queries take about 3 min on c2c SA's dev server (with a quite old DB though):
```
UPDATE 51041
TRUNCATE TABLE
TRUNCATE TABLE
TRUNCATE TABLE
TRUNCATE TABLE
UPDATE 234
UPDATE 382
UPDATE 51041
UPDATE 51157
UPDATE 50123
UPDATE 55590
UPDATE 56873
UPDATE 54442
TRUNCATE TABLE
TRUNCATE TABLE
TRUNCATE TABLE
TRUNCATE TABLE
TRUNCATE TABLE
```

asaunier:
@brunobesson Thanks!

PR #716, opened 2018-01-16
Add rate limiting support for write requests

asaunier:
* [x] Move existing tween (JWT validation) to a dedicated dir
* [x] Create a rate limiting tween that intercepts write requests (POST, PUT, DELETE) only
* [x] Alembic script
* [x] Unit tests

Ordered evolutions:
* [x] Send an email to moderators when a user is rate limited
* [x] Block user after x (in a row?) rate limited statuses
* [x] Have special settings depending on user categories (eg. moderators)
* [x] Add robot to user categories

### DIFF
--- a/c2corg_api/tweens/rate_limiting.py
+++ b/c2corg_api/tweens/rate_limiting.py
@@ -41,7 +41,6 @@ def rate_limiting_tween_factory(handler, registry):
                 'rate_limiting.limit_moderator' if user.moderator else
                 'rate_limiting.limit'))
             user.ratelimit_reset = now + datetime.timedelta(seconds=span)
-            user.ratelimit_limit = limit
             user.ratelimit_remaining = limit - 1
         elif user.ratelimit_remaining:
             user.ratelimit_remaining -= 1
--- a/c2corg_api/tweens/rate_limiting.py
+++ b/c2corg_api/tweens/rate_limiting.py
@@ -36,7 +36,7 @@ def rate_limiting_tween_factory(handler, registry):
             # No window exists or it is expired: create a new one.
             span = int(registry.settings.get('rate_limiting.window_span'))
             limit = int(registry.settings.get('rate_limiting.limit'))
-            user.ratelimit_reset = now + datetime.timedelta(minutes=span)
+            user.ratelimit_reset = now + datetime.timedelta(seconds=span)
             user.ratelimit_limit = limit
             user.ratelimit_remaining = limit - 1
         elif user.ratelimit_remaining:
--- a/c2corg_api/tweens/rate_limiting.py
+++ b/c2corg_api/tweens/rate_limiting.py
@@ -3,6 +3,7 @@ import datetime
 import pytz
 
 from pyramid.httpexceptions import HTTPBadRequest, HTTPTooManyRequests
+from c2corg_api.emails.email_service import get_email_service
 from c2corg_api.views import http_error_handler
 from c2corg_api.models import DBSession
 from c2corg_api.models.user import User
@@ -53,10 +54,15 @@ def rate_limiting_tween_factory(handler, registry):
             if user.ratelimit_last_blocked_window != current_window:
                 user.ratelimit_last_blocked_window = current_window
                 user.ratelimit_times += 1
+
                 max_times = int(
                     registry.settings.get('rate_limiting.max_times'))
                 if user.ratelimit_times > max_times:
                     user.blocked = True
+
+                # An alert message is sent to the moderators
+                email_service = get_email_service(request)
+                email_service.send_rate_limiting_alert(user)
 
             return http_error_handler(
                 HTTPTooManyRequests('Rate limit reached'), request)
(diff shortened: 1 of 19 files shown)


asaunier:
This PR should be ready for reviewing.

What it does:
* move the existing JWT validation tween to a dedicated directory and file
* create a new tween that performs the rate limiting when write requests are detected
* a "robot" flag has been added to the User data model in order to flag (authorized) robot accounts
* moderators and robots have their own (bigger) amount of authorized write actions
* an email is sent to moderators when a contributor reaches their max number of write actions

I have added quite extensive unit tests. The email part is not really tested though (except with the usual existing tests), I am not sure it is really possible to do something about that. My instance have no email tool (yet?), as a result I have not really been able to make sure this feature actually work. Perhaps we could test it on the demo server.

Please note the alert email address is currently a fake one and should be configured in the demo/prod ``docker-compose.yml`` files using a ``rate_limiting_alert_address`` environment var. I have already added this parameter to the demo config, using the dev email address (in prod the topoguide moderators list address must be used).

c2corg_api/tweens/rate_limiting.py
```diff
@@ -0,0 +1,72 @@
...
+            if user.ratelimit_last_blocked_window != current_window:
+                user.ratelimit_last_blocked_window = current_window
+                user.ratelimit_times += 1
+
+                max_times = int(
+                    registry.settings.get('rate_limiting.max_times'))
+                if user.ratelimit_times > max_times:
+                    user.blocked = True
+
+                # An alert message is sent to the moderators
```
asaunier: Please note sending the email postpones a little the "429 Too Many Requests" response.
If it is too long, a timeout is hit and I think it causes a 500 Internal Server Error (which is quite the same than the 429 error :P for the end user).

I don't know if it is possible to send the email asynchronously. Or maybe I could embed the 2 lines below in a try-except block?

asaunier:
> Perhaps we could test it on the demo server.

I have just tested it: emails are sent as expected on the demo server when a user is rate limited.

c2corg_api/tweens/rate_limiting.py
```diff
@@ -0,0 +1,72 @@
...
+        now = datetime.datetime.now(pytz.utc)
+        if user.ratelimit_reset is None or user.ratelimit_reset < now:
+            # No window exists or it is expired: create a new one.
+            span = int(registry.settings.get('rate_limiting.window_span'))
+            limit = int(registry.settings.get(
+                'rate_limiting.limit_robot' if user.robot else
+                'rate_limiting.limit_moderator' if user.moderator else
+                'rate_limiting.limit'))
+            user.ratelimit_reset = now + datetime.timedelta(seconds=span)
+            user.ratelimit_limit = limit
```
arnaud-morvan: I do not understand why you need to store limit in database as it is calculated from user type and settings ?

c2corg_api/tweens/rate_limiting.py
arnaud-morvan: Seems it is possible to use another process to send the mail (through a queue): https://docs.pylonsproject.org/projects/pyramid_mailer/en/latest/#queue

c2corg_api/tweens/rate_limiting.py
asaunier: Right, I have removed this attribute which is indeed not that useful.

c2corg_api/tweens/rate_limiting.py
asaunier: Thanks for the tip!

To be honest I feel a bit lazy doing this because it adds quite some work to set the separate process whereas sending such mails should not be that frequent, especially when the result is refusing a document edition anyway :)

c2corg_api/tweens/rate_limiting.py
```diff
@@ -59,9 +59,12 @@ def tween(request):
...
 
-                # An alert message is sent to the moderators
-                email_service = get_email_service(request)
-                email_service.send_rate_limiting_alert(user)
+                try:
+                    # An alert message is sent to the moderators
+                    email_service = get_email_service(request)
+                    email_service.send_rate_limiting_alert(user)
+                except:
+                    log.debug('RATE LIMITING: error while sending email')
```
arnaud-morvan: Can you add exc_info=True, to get the trace in the log.
Are you sure the send mail timeout occurs before the http server timeout ?

arnaud-morvan: As I understand, the mail is sent when the transaction is commited (pyramid_tm tween) not immediately, except you use the appropriate method.
I do not have found timeout parameter or setting in the pyramid_mailer docs.

arnaud-morvan: Does this occurs on production, or only on your workstation because you cannot reach default config smtp server (psemail) ? Maybe we do not really need to handle this error.

asaunier: I only had this case on my workstation because indeed psemail cannot be used. It seems to work on the demo (not in prod yet). Maybe you're right and this change is not really useful, this can be removed.

arnaud-morvan: The prod server already send other emails using the same method for password recovery.
This try / except block has no effect on timeout that occurs in pyramid_tm tween.

asaunier: You're right. I have removed the commit adding the try/except block.

I will merge the PR when Travis has passed. Thanks for the feedback and advices.

Issue #717, opened 2018-01-31
Add username in password resetting emails


asaunier:
Users often have problems authenticating after resetting their password because they don't realize they don't use the right username (or the right case for it). Thus they think the problem is the password and try to reset it again and again.

It would be relevant to remind the username in the password resetting emails.

stef74:
Not good pratice in terme of security. for me the real problem is the multiple login/pseudo approach

cbeauchesne:
I agree with stef. See discussion 'qu'est-ce qu'il faut ameliorer' o  the forum, where we discuss about this.

IMO,we should have 

* a unique pseudo, used to login, and displayed on forum. user mail is also accepted for login 
* a name, which is basiccly a label used, if filled, in replacment of the pseudo, in outings views.

asaunier:
c2corg is not a bank account, is it really a security issue to remind users of their username when sending emails? If we use the forum_username (or the email) as login username it will be a public information anyway (the email address is obviously available in the message the user receives).

Even if we use a unique username, there will still be people not reminding exactly what their username actually is.

In addition adding the username in the validation message is an easy solution that can be implemented quickly:
* [ ] pass the ``user.username`` attribute to the email body at https://github.com/c2corg/v6_api/blob/master/c2corg_api/emails/email_service.py#L78
* [ ] adapt the email templates (in all languages), eg. https://github.com/c2corg/v6_api/blob/master/c2corg_api/emails/i18n/fr/password_change_body

cbeauchesne:
I also agree with Alex 😊. Actually, I don't think it's a problem to remind username in email, BUT the as the first cause of forgotten username is our multi pseudo system, we should simplify it.

asaunier:
@stef74 @cbeauchesne A PR is ready at #719 Please review :)

PR #719, opened 2018-01-31
Add username in the password change email body

asaunier:
Related to https://github.com/c2corg/v6_api/issues/717

Not tested because I have no way to send emails in my instance. The branch could be tested on the demo instance.


asaunier:
The docker image of this branch has just been deployed on the demo. I have tested the "reset password" email in all language (+ the "change email" confirmation email in FR) and it works as expected.

Ready for review!

Issue #700, opened 2017-10-29
Emails sent by the API are not received by free.fr email addresses


asaunier:
For instance registration or email change emails. See the UI issue @desnoes has created: https://github.com/c2corg/v6_ui/issues/1763

It is maybe related to this function https://github.com/c2corg/v6_api/blob/master/c2corg_api/emails/email_service.py#L53-L72
We used to send emails with ``transfer_encoding`` defaulting to ``quoted-printable`` but had to use ``base64`` instead (using an explicit ``Attachment``) because of encoding problems in messages received by hotmail addresses.

asaunier:
It would be useful to figure out if emails sent by Discourse are also not delivered to free.fr addresses. If not, it probably means the problem is related to the third-party email sending service we use (SparkPost).

asaunier:
According to https://forum.camptocamp.org/t/probleme-de-reception-des-mails-c2c-par-les-adresses-free-fr/198972 emails sent by Discourse (also using the SparkPost service) are correctly received by free.fr addresses. Which means the problem is not related to SparkPost but to the API email system, pointed above.

stef74:
I have identified the error on sparkpost log : 
550 mail dropped, bare LF found

After quick googling : we have found a similar 

> host free.fr[212.27.48.7}] said: 550
mail dropped, bare LF found


Sur internet, des internautes semblent avoir des problèmes de ce genre et disent que dans la classe:
class.phpmailer.php il faut modifier une variable de "\n" à \r\n"
Ce qui, si j'ai bien compris est un problème de terminaison de fin de ligne.

asaunier:
> mail dropped, bare LF found

Ok thanks for the hint!

I have read messages suggesting that free.fr rejects emails with an empty body (which is probably what "bare LF found" means?).

I suspect it is related to the way we build emails with pyramid_mailer:
https://github.com/c2corg/v6_api/blob/master/c2corg_api/emails/email_service.py#L57-L66
=> we convert the email body to an attachment in order to set the ``transfer_encoding`` encoding to ``base64`` instead of the default ``quoted-printable`` that caused encoding problem with hotmail recipients. Then all the message content is in the attachment, probably leaving the body itself empty.

I think we found no other way to set the transfer_encoding:
https://docs.pylonsproject.org/projects/pyramid_mailer/en/latest/index.html#pyramid_mailer.message.Attachment

[...]

PR #723, opened 2018-02-08
[WIP] Add a default body to email messages

asaunier:
Else they are rejected by some email providers (eg. free.fr)

Related to https://github.com/c2corg/v6_api/issues/700


asaunier:
Docker image of the current branch "email-body-attachement" has been pushed to the demo.

Here is an example of the email I have received after requesting a password reset:

capture du 2018-02-08 17-44-12

The first line ("FIXME: base email body") is the default email content introduced by this PR. If the suggested change works with free.fr addresses, we could change this test content to something more relevant.

@desnoes Could you please test with your free.fr address?
Testing with other email providers would make sense as well to make sure it is still OK with them.

brunobesson:
As said on the mailing list, this does not seem entirely satisfying.
What about a solution distinguishing between the servers the mail is sent to?
AFAIU, we could do the following:

* attachment only for hotmail (which is the only case we have (?) with encoding problems)
* text only for others

julsbreakdown:
@asaunier 
I've modified the code to its original state (before the hotmail fix).
It stills works with hotmail adresses, but this does not fixe the initial free problem.

julsbreakdown:
Update : when changing SMTP server to gmail, mails sent to free adresses work.
So the problem seems to be with the SMTP server sparkpost.
In order to try to fix this problem, would it be possible to have access to the sparkpost server?
@stef74 @desnoes @asaunier

stef74:
I need your mail to configure your account

2018-05-14 11:48 GMT+02:00 jwaddle <notifications@github.com>:

julsbreakdown:
Here is mine julien.waddle@camptocamp.com could you also add arnaud.morvan@camptocamp.com

stef74:
Invitation Sent !

2018-05-14 12:16 GMT+02:00 jwaddle <notifications@github.com>:

asaunier:
Closing because replaced by https://github.com/c2corg/v6_api/pull/725

PR #725, opened 2018-05-16
Send emails using smtplib directly (without pyramid mailer).

arnaud-morvan:
Also note the email template files openned using universal newline support.

This fix error 550 "Bare LF found" with free.fr.

Relate #700

### DIFF
--- a/c2corg_api/emails/email_service.py
+++ b/c2corg_api/emails/email_service.py
@@ -1,8 +1,11 @@
 import transaction  # NOQA
 
-from pyramid_mailer import get_mailer
-from pyramid_mailer.message import Message, Attachment
+from email.mime.text import MIMEText
+from email.mime.multipart import MIMEMultipart
 from functools import lru_cache
+from pyramid.settings import asbool
+import smtplib
+
 from c2corg_common.attributes import default_langs
 
 import logging
@@ -20,8 +23,9 @@ class EmailLocalizator(object):
         filepath = os.path.dirname(__file__) + '/i18n/%s/%s' % (lang, key)
         if not os.path.isfile(filepath):
             filepath = os.path.dirname(__file__) + '/i18n/%s/%s' % ('fr', key)
-        f = open(filepath, 'r')
-        return f.read().rstrip()  # No trailing new line in email subject!
+        with open(filepath, 'rU') as f:  # Use universal newline support.
+            content = f.read().rstrip()  # No trailing new line in subject!
+        return content
 
     def get_translation(self, lang, key):
         if lang not in default_langs:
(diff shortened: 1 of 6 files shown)


arnaud-morvan:
@asaunier: I will put that branch on demo for testing after travis finished to build the docker image.

julsbreakdown:
:1st_place_medal:

arnaud-morvan:
@asaunier @gberaudo : Can someone have a look, this pass Travis now and is testable on demov6.

asaunier:
@stef74 @desnoes Could you have the demo tested to make sure emails sent by the demo server are actually received by any kind of email addresses? Thanks!

c2corg_api/emails/email_service.py
```diff
@@ -20,8 +23,9 @@ def _get_file_content(self, lang, key):
         filepath = os.path.dirname(__file__) + '/i18n/%s/%s' % (lang, key)
         if not os.path.isfile(filepath):
             filepath = os.path.dirname(__file__) + '/i18n/%s/%s' % ('fr', key)
-        f = open(filepath, 'r')
-        return f.read().rstrip()  # No trailing new line in email subject!
+        with open(filepath, 'rU') as f:  # Use universal newline support.
+            content = f.read().rstrip()  # No trailing new line in subject!
+        return content
```
asaunier: Don't you need to move the ``return`` statement to the ``with``block where ``content`` is defined?

desnoes:
@asaunier: i will do some tests today.

c2corg_api/emails/email_service.py
arnaud-morvan: Yes it should be better for the case the file does not exist.

arnaud-morvan:
I've tried to send a password recover email to an address in free.fr
The new error is 550 spam detected.

arnaud-morvan:
I sent an email to proxad.net as explained in http://postmaster.free.fr/#faux_positifs, waiting for a response now ...

arnaud-morvan:
Seems that Free SPAM policy has changed, I've just received a recover password email from demov6 in a free.fr mailbox.

desnoes:
@arnaud-morvan : i have also tested and receive the email with my yahoo, gmail and free adresses :). Thanks. I have asked people from the association to do more tests with other email adresses. I would like to check it works with hotmail. We had some problem with this mail in the past.

asaunier:
It seems Free has changed things about the emails we sent and now emails sent by the demo are received as expected. Is this PR needed as well?

arnaud-morvan:
@asaunier : Note that this does not fix the "550 spam detected" but "550 bare LF found"

asaunier:
> Note that this does not fix the "550 spam detected" but "550 bare LF found"

@arnaud-morvan OK, thanks for explaining :)

> Seems we should merge this PR now, if no one emits comments about it, I will merge today.

LGTM!

Issue #731, opened 2018-12-12
Add markdown parser


cbeauchesne:
* [x] markdown service for document 
* [x] cooker 
* [x] versions 
* [x] tests

### markdown service for document

GET `/routes/12345?cook=fr`

-> will return a document, without locales, but with a `cooked` property. This property is a object corresponding to a unique locale. This locale is in the lang asked in query parameter, or in a lang regarding `lang_priority` list

### cooker 

POST `/cooker`

* body : key-value object, where values are markdown 
* response : key-value object, where values are html, only if key correspond to a markdown field.

PR #732, opened 2018-12-18
Markdown cooker

cbeauchesne:
#731 

Add cook=lang in query for document
Add cooker service
And and unit tests

### DIFF
--- a/c2corg_api/tests/views/test_cooker.py
+++ b/c2corg_api/tests/views/test_cooker.py
@@ -6,8 +6,14 @@ class TestCookerRest(BaseTestRest):
         super(TestCookerRest, self).setUp()
 
     def test_get(self):
-        body = {
+        markdowns = {
             "lang": "fr",
-            "oui": "**coucou** *italic*"
+            "description": "**strong emphasis** and *emphasis*"
         }
-        self.app_post_json('/cooker', body, status=200)
+        response = self.app_post_json('/cooker', markdowns, status=200)
+
+        htmls = response.json
+
+        # lang is not a markdown field, it must be untouched
+        self.assertEqual(markdowns['lang'], htmls['lang'])
+        self.assertNotEqual(markdowns['description'], htmls['description'])
--- a/c2corg_api/tests/views/test_cooker.py
+++ b/c2corg_api/tests/views/test_cooker.py
@@ -1,0 +1,13 @@
+from c2corg_api.tests.views import BaseTestRest
+
+
+class TestCookerRest(BaseTestRest):
+    def setUp(self):  # noqa
+        super(TestCookerRest, self).setUp()
+
+    def test_get(self):
+        body = {
+            "lang": "fr",
+            "oui": "**coucou** *italic*"
+        }
+        self.app_post_json('/cooker', body, status=200)
--- a/c2corg_api/views/cooker.py
+++ b/c2corg_api/views/cooker.py
@@ -1,0 +1,38 @@
+import logging
+
+from c2corg_api.views import cors_policy
+from cornice.resource import resource, view
+from c2c_markdown import parse_code, configure_parsers
+
+log = logging.getLogger(__name__)
+
+configure_parsers({'api_url': 'https://api.camptocamp.org/'})
+
+# this is locale property, but they must not be cooked by markdown parser
+NOT_MARKDOWN_PROPERTY = {
+    'lang',
+    'version',
+    'title',
+    'slope',
+    'conditions_levels',
+    'topic_id',
+    'participants'
+}
+
+
+def cook(data):
+    result = {}
+    for key in data:
+        if key not in NOT_MARKDOWN_PROPERTY and data[key]:
+            result[key] = parse_code(data[key])
+        else:
+            result[key] = data[key]
+
+    return result
+
+
+@resource(path='/cooker', cors_policy=cors_policy)
+class CookerRest(object):
+    @view()
+    def post(self):
+        return cook(self.request.json)
--- a/c2corg_api/views/cooker.py
+++ b/c2corg_api/views/cooker.py
@@ -15,4 +15,12 @@ class CookerRest(object):
 
     @view()
     def post(self):
+        """
+        This service is a stateless service that returns HTML values.
+
+        * Input and output are dictionaries
+        * keys are keep unmodified
+        * values are parsed from markdown to HTML, only if key is not in
+          c2corg_api.views.markdown.NOT_MARKDOWN_PROPERTY
+        """
         return cook(self.request.json)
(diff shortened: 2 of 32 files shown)


docker-compose.yml
```diff
@@ -1,5 +1,5 @@
 api:
-  image: 'docker.io/c2corg/v6_api:latest'
+  image: 'docker.io/c2corg/v6_api:markdown-req'
```
brunobesson: Do not forget to remove this when merging.

requirements.txt
```diff
@@ -45,6 +45,9 @@ git+https://github.com/tsauerwein/ColanderAlchemy.git@c2corg
 # needs: https://github.com/mozilla-services/cornice/pull/359
 git+https://github.com/tsauerwein/cornice.git@nested-none-2.1.0-c2corg
 
+# c2c_markdown repo
+git+https://github.com/c2corg/c2c_markdown.git@master
```
brunobesson: Do we really need it from master? I'd rather reference a tag here.

cbeauchesne: A commit is not possible, because for a reason I ignore, pip won't install dependancies.

It's a commit for common, but it does no fail only because v6_common has not dependancies...

I propose to let it as is for a couple of month, because I will probably modify this dep.

As we will be in a more stable situation, we'll use a tag (if it works) or a dedicated branch.

docker-compose.yml
cbeauchesne: I'm not sure to perfectly handle docker image generation process, but as far as I've understood, image docker.io/c2corg/v6_api:latest won't include good dependancies, so travis process will, so it won't build the docker image.

So I think we'll need at least one commit on master with `docker.io/c2corg/v6_api:markdown-req`, and then we could use `docker.io/c2corg/v6_api:master`

@mfournier, am I wrong ?

c2corg_api/views/document.py
```diff
@@ -125,18 +144,25 @@ def create_response():
...
                      clazz_locale=None, adapt_schema=None,
                      include_maps=True, include_areas=True,
-                     set_custom_associations=None, set_custom_fields=None):
-        document = self._get_document(
-            clazz, id, clazz_locale=clazz_locale, lang=lang)
+                     set_custom_associations=None, set_custom_fields=None,
+                     cook_locale=False):
+
+        if cook_locale:
+            document = self._get_document_for_cooking(
```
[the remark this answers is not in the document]
cbeauchesne: it's not the same function because `_get_document` returns the expected locale, or no locale if this lang does not exists, where `_get_document_for_cooking` returns the expected locale, or another one, based on lang_priority array.

c2corg_api/tests/views/test_cooker.py
```diff
@@ -0,0 +1,13 @@
...
+class TestCookerRest(BaseTestRest):
+    def setUp(self):  # noqa
+        super(TestCookerRest, self).setUp()
+
+    def test_get(self):
+        body = {
+            "lang": "fr",
+            "oui": "**coucou** *italic*"
+        }
+        self.app_post_json('/cooker', body, status=200)
```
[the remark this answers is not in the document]
cbeauchesne: Actually, it's not the job of this repo to check markdown. By the way, I can add two assertion : 

* lang value's must not be touched, as it's not a markdown field
* oui value's must be modified, because it's not part of NOT_MARKDOWN_PROPERTY

c2corg_api/views/cooker.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@resource(path='/cooker', cors_policy=cors_policy)
+class CookerRest(object):
+
+    def __init__(self, request):
+        self.request = request
+
+    @view()
+    def post(self):
```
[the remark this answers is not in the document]
cbeauchesne: sure

c2corg_api/views/document.py
```diff
@@ -110,12 +112,29 @@ def _get(self, clazz, schema, clazz_locale=None, adapt_schema=None,
         id = self.request.validated['id']
         lang = self.request.validated.get('lang')
         editing_view = self.request.GET.get('e', '0') != '0'
+        cook = self.request.validated.get('cook')
+
+        if cook and lang:
```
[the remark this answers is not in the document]
cbeauchesne: no, because the inside logic is different. 

* `/waypoints/123?lang=en` will returns `en` locale, or no locale. 
* And `/waypoints/123?cook=en` will aways returns a locale and a cooked, even if it's not `en`

I figure that a query parameter must not touch the behavior of another query parameter, and prefer to use another one.

docker-compose.yml
brunobesson: Nope. Once your branch is merged into master, it will be latest, so it will be ok. You only need it in the meantime.

requirements.txt
brunobesson: I would prefer to create a version in c2c_markdown. And create as many as needed later on. You can't rely on a moving target, so a tag is appropriate.

c2corg_api/tests/views/test_cooker.py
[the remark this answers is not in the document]
tsauerwein: True, in-depth tests of the Markdown parser are not required. I was more thinking of a very basic test to see that it works at all.

requirements.txt
cbeauchesne: See the key point in my previous response : 

> It's a commit for common, but it does no fail only because v6_common has not dependancies...

docker-compose.yml
cbeauchesne: Ok.

requirements.txt
tsauerwein: I am not sure why it didn't work for you. But for pip it makes no difference if you use a branch name, commit or tag. For testing, I set your latest commit on master in this branch:
https://github.com/c2corg/v6_api/commit/f042a8314932f9e6addc59ebfe59b231c5fbd259

And the Travis build passed:
https://travis-ci.org/c2corg/v6_api/jobs/470141545

cbeauchesne: Well, to be frank, I can't explain why I came to this conclusion. Your test is crystal clear...

For the next few week, as I'll probably iterate a lot to fix small bugs, would you mind if we  keep it on master branch?

Then, when we'll be in production, we'll set it to a commit, or a branch (I prefer this last option, the branch name is more explicit).

brunobesson: I insist, rather than a branch, the tag is the way to go. A branch is still a moving target.
I you know I don't understand why you don't want to create tags :) but anyway, as long as there is a tag when merged into master, I'm happy.

cbeauchesne: I wanted to use master to keep a moving target during these week. But in fact, it seems taht docker image doesn't update itself, as it consider it still contains the dependancy.

So I just commited this line with an explicit commit : 7012ce6

This said, I also agree that a tag is better than a commit : more explicit. I just saw your message after my last commit 5 minutes ago, sorry.

Issue #733, opened 2019-02-11
Add sitemap.xml route


cbeauchesne:
See https://forum.camptocamp.org/t/projet-ui-moteur-de-recherche-et-sitemaps/237172 

As the new UI won't be able to transform rest info to XML, api will have to do the job.

PR #734, opened 2019-02-13
Sitemaps xml #733

cbeauchesne:
Add routes : 

* /sitemaps.xml
* /sitemaps.xml/b/0.xml 

...and all set b, r, w.... 

It will replace actual https://www.camptocamp.org/sitemaps.xml

### DIFF
--- a/c2corg_api/views/sitemap_xml.py
+++ b/c2corg_api/views/sitemap_xml.py
@@ -83,7 +83,10 @@ class SitemapXml(object):
         self.request.response.content_type = "text/xml"
 
         return get_or_create(
-            cache_sitemap_xml, cache_key, _get_sitemap_index)
+            cache_sitemap_xml,
+            cache_key,
+            _get_sitemap_index
+        )
 
     @view(validators=[validate_page, validate_document_type])
     def get(self):
--- a/c2corg_api/views/sitemap_xml.py
+++ b/c2corg_api/views/sitemap_xml.py
@@ -43,7 +43,7 @@ validate_page = create_int_validator('i')
 
 
 @resource(
-    collection_path='/sitemaps.xml', path='/sitemaps.xml/{doc_type}/{i}',
+    collection_path='/sitemaps.xml', path='/sitemaps.xml/{doc_type}/{i}.xml',
     cors_policy=cors_policy, renderer='string')
 class SitemapXml(object):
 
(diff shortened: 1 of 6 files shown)


docker-compose.yml
```diff
@@ -1,5 +1,5 @@
 api:
-  image: 'docker.io/c2corg/v6_api:markdown-cooker'
+  image: 'docker.io/c2corg/v6_api:suglify-dep'
```
brunobesson: shouldn't this be reset to dev-environment or something similar?

c2corg_api/views/sitemap_xml.py
```diff
@@ -0,0 +1,222 @@
...
+    'r': 'routes',
+    'w': 'waypoints',
+    'x': 'xreports'
+}
+
+validate_page = create_int_validator('i')
+
+
+@resource(
+    collection_path='/sitemaps.xml', path='/sitemaps.xml/{doc_type}/{i}.xml',
```
brunobesson: `sitemap.xml` instead of `sitemaps.xml`?

c2corg_api/views/sitemap_xml.py
```diff
@@ -0,0 +1,222 @@
...
+                </sitemap>
+            </sitemap>
+        """
+        cache_key = _get_cache_key()
+        etag_cache(self.request, cache_key)
+
+        self.request.response.content_type = "text/xml"
+
+        return get_or_create(
+            cache_sitemap_xml, cache_key, _get_sitemap_index)
```
brunobesson: no line return (does not seem needed?)

Otherwise something like 
```
return get_or_create(
            cache_sitemap_xml,
            cache_key,
            _get_sitemap_index
)
```

would be more readable IMHO

docker-compose.yml
cbeauchesne: yes :)

c2corg_api/views/sitemap_xml.py
cbeauchesne: I've stick to api logic : /routes and /routes/12345

brunobesson: Depends what is referenced in `robots.txt`

cbeauchesne: In robots.txt, you will have a reference to sitemap index, which is a list of site maps files. This one is on plural form.

The question can be asked for real sitemaps, between this four form : 

1. `sitemaps.xml/b/0.xml`
2. `sitemaps/b/0.xml`
3. `sitemap.xml/b/0.xml`
4. `sitemap/b/0.xml`

1) strictly stick to API route naming logic 
2) stick to API logic a little less, by removing useless extension in first 
3) make no sense :) 
4) the best IMO
 
All that said, I think it's a detail :)

brunobesson: Yes, it is.

Issue #746, opened 2019-08-10
When a new version of an image is upload, notify image backend to publish it


cbeauchesne:
See c2corg/c2c_ui#388, on document new version, check if filename has been modified, and if yes, send publish request to image backend.

PR #747, opened 2019-08-11
Publish new image version

cbeauchesne:
PR for #746, itself for c2corg/c2c_ui#388

The idea is, if document.filename is updated on image, then send `publish` request to image backend : [diff](https://github.com/c2corg/v6_api/compare/publish-new-image-version?expand=1#diff-0ed2cf87a7823395be03efb0711e74beR119)

I've made some revamp to achieve this. The best place for this feature is `before_update()` callback : it's called before any update : [link](https://github.com/c2corg/v6_api/compare/publish-new-image-version?expand=1#diff-b906af94797c5db6c394971ac60bf7a4L379). But, surprisingly, even if its arguments are `actual_doc` and `new_doc`, it was called **after** update in memory.

So, I fixed that : 

* move `before_update()` call before `document.update(document_input)`
* remove useless `user_id` argument from `before_update` signature
* <del>then, it appears that `route.before_update` is now useless. It's pupose was to update geometry based on linked waypoints. New UI **always** send geometry (means that it's handled by users)</del>
* <add>`route.before_update()`purpose was to always recompute localisation from GPS trace, or associated waypoints. This behaviour is now obsolete, as we let contributor handle geolocalisation point. But, as @saunier noticed it (see below), other UI may send empty geometry. In this cas, we set this point from the old version, as we always need it. Btw, I didn't set GPS trace from old version, because we may want to offer the possibility to remove GPS trace</add>
* same on `outing.before_update`, at this small exception : i keep it to keep update on localisation base on GPS trace.
* And at the end, remove useless unittest.

### DIFF
--- a/c2corg_api/tests/views/test_outing.py
+++ b/c2corg_api/tests/views/test_outing.py
@@ -839,9 +839,6 @@ class TestOutingRest(BaseDocumentTestRest):
                 ],
                 'geometry': {
                     'version': self.outing.geometry.version,
-                    'geom_detail':
-                        '{"type": "LineString", "coordinates": ' +
-                        '[[635956, 5723604], [635976, 5723654]]}',
                     'geom':
                         '{"type": "Point", "coordinates": [635000, 5723000]}'
                 },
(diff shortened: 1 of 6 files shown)


asaunier:
> then, it appears that route.before_update is now useless. It's pupose was to update geometry based on linked waypoints. New UI always send geometry (means that it's handled by users)

Is this a correct approach?
The API is UI-agnostic. It's mainly used by c2c_ui now but should work with any UI (eg. the mobile app or whatever). Then the API code should not assume that geometries are *always* provided. Having a default behaviour if none is provided still makes sense.

cbeauchesne:
@asaunier good catch, I'll fix that

Issue #707, opened 2017-11-14
A bad doc url with an area ID return a error 500 instead of 404


fbunoz:
An url using a area ID but a bad doc type, return an error 500 instead of 404 : 
https://www.camptocamp.org/routes/14403
https://www.camptocamp.org/articles/14403

For a waypoint ID, it return an error 404 : 
https://www.camptocamp.org/routes/43071
https://www.camptocamp.org/articles/43071/fr

But if there is a slug, it returns a 500 : 
https://www.camptocamp.org/articles/43071/fr/slug

cbeauchesne:
I observe : 

## 14403 is an area : https://www.camptocamp.org/areas/14403

* https://api.camptocamp.org/routes/14403 => 200 io 404
* https://api.camptocamp.org/articles/14403 => 200 io 404
* https://api.camptocamp.org/waypoints/14403 => 200 io 404

## 43071 is a waypoint : https://www.camptocamp.org/waypoints/43071

* https://api.camptocamp.org/routes/43071 => 200 io 404
* https://api.camptocamp.org/articles/43071 => 200 io 404
* https://api.camptocamp.org/articles/43071?cook=fr => 200 io 404

Basiclly, it returns the good document id, but foes not check the document type

cbeauchesne:
FI, it's a cahcing issue. If the document is not in cache, it returns 404. If it's the cache, the cache key does not include document type, and thus returns the document only based on the id.

cbeauchesne:
fixed one year ago

Issue #123, opened 2016-01-09
migration.py chokes on an sqlalchemy.exc.IntegrityError


mfournier:
When running the import script against current production database, it failed with this error:

```
$ python c2corg_api/scripts/migration/migrate.py
Importing users
Total: 46021 rows
1 of 46021
4004 of 46021
7007 of 46021
10530 of 46021
13013 of 46021
Traceback (most recent call last):
  File "c2corg_api/scripts/migration/migrate.py", line 65, in <module>
    main()
  File "c2corg_api/scripts/migration/migrate.py", line 54, in main
    MigrateUsers(connection_source, session, batch_size).migrate()
  File "/usr/src/v6_api/c2corg_api/scripts/migration/users.py", line 70, in migrate
    email_validated=id not in pending
  File "/usr/src/v6_api/c2corg_api/scripts/migration/batch.py", line 52, in add
    self.flush_or_not()
  File "/usr/src/v6_api/c2corg_api/scripts/migration/batch.py", line 14, in flush_or_not
    self.flush()
  File "/usr/src/v6_api/c2corg_api/scripts/migration/batch.py", line 59, in flush
    self.session.bulk_insert_mappings(self.model, self.entities)
  File "/usr/src/v6_api/.build/venv/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 2280, in bulk_insert_mappings
    mapper, mappings, False, False, return_defaults, False)
  File "/usr/src/v6_api/.build/venv/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 2351, in _bulk_save_mappings
    transaction.rollback(_capture_exception=True)
  File "/usr/src/v6_api/.build/venv/local/lib/python2.7/site-packages/sqlalchemy/util/langhelpers.py", line 60, in __exit__
    compat.reraise(exc_type, exc_value, exc_tb)
  File "/usr/src/v6_api/.build/venv/local/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 2346, in _bulk_save_mappings
    mapper, mappings, transaction, isstates, return_defaults)
  File "/usr/src/v6_api/.build/venv/local/lib/python2.7/site-packages/sqlalchemy/orm/persistence.py", line 67, in _bulk_insert
    bookkeeping=return_defaults)
  File "/usr/src/v6_api/.build/venv/local/lib/python2.7/site-packages/sqlalchemy/orm/persistence.py", line 752, in _emit_insert_statements
[...]

PR #752, opened 2019-10-16
When asked doc type is wrong, returns 404 io 200

cbeauchesne:
Fix for #707.

So, basicly, if a document is asked, the API look for it in a cache. The cache key does not include the document type : 

https://github.com/c2corg/v6_api/blob/3a5d5202da0ee49903b7ce495e141c41dce27892/c2corg_api/models/cache_version.py#L166

and 

https://github.com/c2corg/v6_api/blob/3a5d5202da0ee49903b7ce495e141c41dce27892/c2corg_api/views/document.py#L121

<small>Note that on the second line, `custom_cache_key` is sent in place of document_type... As far as I know, it does not cause a bug, hopefully...</small>

In consequence, imagine that you have an area with id 123. From an empty cache, do :  

1. `/routes/123`=> 404 ✅  
2. `/areas/123`=> 200 and the API save the doc in the cache
2. `/areas/123`=> 200, served from the cache 👍 
2. `/routes/123`=> 200, area #123 served from the cache, instead of 404 😡 

What I've done ? 

Considering that the philosophy followed by the API is to precise document types in URL, I've changed the `document_type` argument to mandatory on `get_cache_key()` : 

https://github.com/c2corg/v6_api/pull/752/files#diff-71235d266344278ce3ed31500c0df856R166

And small hack : history entry point does not precise the document type : `/document/124/history/fr`.

For this, I've used the generic `DOCUMENT_TYPE` type : 

https://github.com/c2corg/v6_api/pull/752/files#diff-d997b23b03a7bb5f76a2c2297fc55b6bR35

Sorry for this big commit, but it appears that there are quite a lot of boilerplate code in the API.

@asaunier, @b_b, would it be possible that you have a look on this commit ? :pray:


tsauerwein:
> * `/routes/123`=> 404 white_check_mark
> 
> * `/areas/123`=> 200 and the API save the doc in the cache
> 
> * `/areas/123`=> 200, served from the cache +1
> 
> * `/routes/123`=> 200, served from the cache

Sorry, I didn't read everything in your description. But it is not possible that there is an area and a route with the same document-id. All documents, no matter what type, have a row in the `documents` table. So, the document-id is unique.

cbeauchesne:
@tsauerwein yes, that's the point : the document served in the 4th request `/routes/123`** is the area**.  

You can easily try in prod: just look for a random document, request it with it's good document type, **then** request it with another document type.

cbeauchesne:
And f**k, Travis has updated ubuntu from 14 to 16, it breaks the CI.

cbeauchesne:
Thanks for the review @asaunier.

The demo server is actually down for a reason I ignore : https://api.demov6.camptocamp.org => I'l try to have look, but very little chance that I understand what's hapenning. Maybe @mfournier  ?

PR #920, opened 2020-11-16
Remove warnings from tests

cbeauchesne:


asaunier:
Could you explain why you made those function renamings? Related to the switch from nosetests to pytest?

cbeauchesne:
It would be more precise to say that the problem was present, but hidden. Switching to pytest reveal them, as it reports warnings, and nostests don't.

In a word, there were a ton of deprecation warnings in test output. I plan to update to a newer version of python and update deps, I need to anticipate. And also, it cleans test outputs :)

asaunier:
OK, I guess there were aliases such as "assertEquals" instead of "assertEqual" that are now no longer supported or warned as deprecated?

asaunier (approved):

PR #921, opened 2020-11-17
Do not rely on Referer HTTP header data when creating topic on Discourse

### LINKED ISSUE #736: Do not use referer to identify destination url on topic creation on discourse
First of all, read this : https://blog.mozilla.org/security/2018/01/31/preventing-data-leaks-by-stripping-path-information-in-http-referrers/

Modern browsers offers an option on privacy settings that remove any url content in referer field, in order to prevent personal data leak.

In consequence, this code does not contains the good URL, and then, the topic does not links to the good document : 

https://github.com/c2corg/v6_api/blob/3a5d5202da0ee49903b7ce495e141c41dce27892/c2corg_api/views/forum.py#L91-L94

We must change the way we get the destination url : 

* rebuild it from document_id, lang, and title (we have everything we need, except the disctinction between prod/demo domain. We could hard-write production domain 
* or send it in the body request

tristan0x:
Assume that the web browser wiped the referrer data (path, query, ...)
because calling the API endpoint.
This changes rebuilds the URL manually by using both the document identifier
passed in the request payload and the database to retrieve the document type.

Fixes #736

### DIFF
--- a/c2corg_api/views/forum.py
+++ b/c2corg_api/views/forum.py
@@ -1,12 +1,12 @@
 import colander
-from urllib3.util.url import parse_url
 
 from c2corg_api.models.common import document_types
 
 from c2corg_api.models import DBSession
-from c2corg_api.models.association import Association
+from c2corg_api.models.association import Association, \
+    association_keys_for_types
 from c2corg_api.models.cache_version import update_cache_version_direct
-from c2corg_api.models.document import DocumentLocale
+from c2corg_api.models.document import Document, DocumentLocale
 from c2corg_api.models.document_topic import DocumentTopic
 from c2corg_api.models.user import User
 from c2corg_api.security.discourse_client import get_discourse_client
@@ -66,6 +66,16 @@ def validate_topic_create(request, **kwargs):
         return
     request.validated['locale'] = locale
 
+    document = DBSession.query(Document) \
+        .filter(Document.document_id == document_id) \
+        .one_or_none()
+    if document is None:
+        request.errors.add('body',
+                           '{}/{}'.format(document_id, lang),
+                           'Document not found')
+        return
+    request.validated['document'] = document
+
     if locale.topic_id is not None:
         request.errors.add('body',
                            '{}_{}'.format(document_id, lang),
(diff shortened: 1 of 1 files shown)


tristan0x:
ping @cbeauchesne

c2corg_api/views/forum.py
```diff
@@ -86,12 +97,25 @@ def __init__(self, request):
     def collection_post(self):
         settings = self.request.registry.settings
 
+        import pprint
         locale = self.request.validated['locale']
+        document = self.request.validated['document']
+        log.warning(pprint.pformat(dir(locale)))
```
cbeauchesne, line 103: Is warning the good level? Maybe debug ?

c2corg_api/views/forum.py
```diff
@@ -86,12 +97,25 @@ def __init__(self, request):
...
+        document_type = association_keys_for_types[document.type]
+
+        referer = self.request.referer or '/'
+        referer_path = parse_url(referer).path
+        # only keep the left part of referer path
+        referer_scheme_host = referer.rsplit(referer_path, 1)[0]
```
cbeauchesne, line 109: if `self.request.referer` is missing, then `referer_scheme_host` will be an empty string.

As the host site will always be www.camtocamp.org, I think we should rather keep it very simple: 

``` python
content = '<a href="https://www.camptocamp.org{}">{}</a>'.format(referer_path)
``` 

And IMO,  rename `referer_path` to `document_path`

cbeauchesne (changes requested):

c2corg_api/views/forum.py
tristan0x: This was debug code I removed in 7bf9533

c2corg_api/views/forum.py
tristan0x: I applied your suggestion in 7bf9533

tristan0x:
One test fails in Travis CI fails with 7bf9533

> c2corg_api/tests/views/test_association.py:454: AssertionError

So far, I cannot reproduce the issue on my workstation. I will try later on with a fresh database.

lbesson:
could you rebase against latest master?

tristan0x:
> could you rebase against latest master?

It's done @lbesson 
Sorry for the delay...

lbesson:
No problem, I'm just trying to make teh api live again.
Your response time is fine to me :)

I didn't follow this from the beginning, what is the status in your opinion? Should it be merged? Is som additional work required?

tristan0x:
> No problem, I'm just trying to make the api project live again.
> Your response time is fine to me :)
> 
> I didn't follow this from the beginning, what is the status in your opinion? Should it be merged? Is some additional work required?

Charles's points have been addressed, so it is mergeable in my opinion. Maybe @cbeauchesne can take a last look before we merge this.
Cheers!

cbeauchesne:
If I remember correctly, it's all good.!

lbesson:
Great. Let's merge then.

PR #922, opened 2020-11-17
Use the good name for ES dev index

cbeauchesne:


asaunier:
No clue about what the good name is but what was the problem with this one?

cbeauchesne:
The index on  our image elasticsearch containing anonymized data was `c2corg`. In consequences, any url with a filter was not working (url without a filter hit the postgresDB, it explains why we missed it, we never checked with a filter ).

asaunier:
OK thanks

PR #923, opened 2020-11-17
Add tests in github actions

cbeauchesne:


cbeauchesne:
Issue on waitress that prevent from sevring on `0.0.0.0` on github action : https://github.com/Pylons/waitress/pull/153

It's possible to bypass by changing host to 127.0.0.1 in development.ini, but it make the api unreachable from outside the container for regular usage.

Also an issue on healthcheck on postgres.

So wait for dependancy upgrade.

PR #1017, opened 2021-03-27
Use larger limit for retrieved routes associated to waypoint 

brunobesson:
30 > 400. This should cover the worst case scenarii described in https://github.com/c2corg/c2c_ui/issues/754 and allow to solve this issue from the frontend side.

### DIFF
--- a/c2corg_api/views/waypoint.py
+++ b/c2corg_api/views/waypoint.py
@@ -35,7 +35,7 @@ from sqlalchemy.sql.elements import literal_column
 from sqlalchemy.sql.expression import and_, union
 
 # the number of routes that are included for waypoints
-NUM_ROUTES = 30
+NUM_ROUTES = 400
 
 validate_waypoint_create = make_validator_create(
     fields_waypoint, 'waypoint_type', waypoint_types)


brunobesson:
After some thought, the worst case I could think of is the calanques, with 857 results ad of today. I agree with @cbeauchesne comment: does it have really a meaning? Shouldn't it be an area instead? Anyway, should we use a greater limit?

lbesson (approved):

cbeauchesne:
> Shouldn't it be an area instead? 

IMO, yes. 

> Anyway, should we use a greater limit?

My only concern is about performance. If the perfs are ok, I do not have any blocker to put this limit to whatever is needed (and let moderators decide what's best in term of topoguide organization).

So to give a objective answer, I would say (magical wet finger) that we should set this limit to a number that do not decrease the perf more than 15% regarding the actual status.

ataraxido (approved):

cbeauchesne (approved):

lbesson:
can you rebase against latest master ?

brunobesson:
Done.

lbesson:
so, seems it doesn't break unit tests ;)

brunobesson:
OK, this is merged and tested. Documents like Gorges du Verdon or Aiguilles de Chamonix do have a slight latency on first call (< 2s in my tests), which drops down to 50 ms once cache is populated (all this on the demo server).

I do think this is acceptable.

PR #1019, opened 2021-03-29
Remove duplicates in recent outings

brunobesson:
As usual, I do not know python and the pyramid framework, nor am I testing this.
But we know the request can return duplicates (see e.g. Meij waypoint with says it has >2000 outings, but really only ~800), I assume this could fix this.
As a consequence we could enhance UI a little bit and remove some useless code on the frontend part.

### DIFF
--- a/c2corg_api/views/route.py
+++ b/c2corg_api/views/route.py
@@ -140,7 +140,7 @@ class RouteRest(DocumentRest):
                 Association,
                 Outing.document_id == Association.child_document_id).
             filter(Association.parent_document_id == route.document_id).
-            group_by(Outing.document_id).
+            distinct().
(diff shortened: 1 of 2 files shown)


lbesson (commented):
Moment mal.

When I find time I can try to merge in my update branch and run tests for this in there

brunobesson:
I tried to deploy the docker image on demo server, using first `group_by` then `distinct`, but it didn't seem to work 🙁

cbeauchesne:
If I'm not wrong, the code you modified is about routes associated outings.

But the bug is on Waypoints associated outings, am I right ?

brunobesson:
This could explain why my tests didn't work 🙂 
I'll try to fix this.

Issue #589, opened 2016-12-15
Duplicated routes and outings in the lists on a waypoint page


fbunoz:
- [ ] If a route is associated to several sub-waypoints of the same over-waypoint (or associated to over-waypoint + sub-waypoint), it appears several times in the route list on the over-waypoint page.
The sub-waypoint routes must appear, but the duplicated results returned by the BD request must be removed (or maybe the BD can do this ?).
- [ ] If an outing is associated to several routes of the same waypoint, it appears several times in the outing list on the waypoint page.

fbunoz:
Another issue exist in the UI (but it is an API issue) : https://github.com/c2corg/v6_ui/issues/1213

asaunier:
Another example: https://forum.camptocamp.org/t/les-deux-soeurs-en-doublon/192884

brunobesson:
Fixed by #1022 and #1019

brunobesson:
Fixed.

PR #1023, opened 2021-04-01
Have distinct routes for waypoint results (one more try, DISTINCT is not enough)

brunobesson:
Looks like returned routes of a waypoint still have duplicates

### DIFF
--- a/c2corg_api/tests/views/test_waypoint.py
+++ b/c2corg_api/tests/views/test_waypoint.py
@@ -171,19 +171,17 @@ class TestWaypointRest(BaseDocumentTestRest):
         all_linked_routes = associations.get('all_routes')
         linked_routes = all_linked_routes['documents']
         self.assertEqual(2, len(linked_routes))
-        linked_route = linked_routes[0]
+        linked_route = linked_routes[1]
         self.assertIn('type', linked_route)
         self.assertEqual(
             linked_route['document_id'], self.route1.document_id)
         self.assertEqual(
             linked_route.get('locales')[0].get('title_prefix'), 'Mont Blanc :')
-        # the route associated to a child waypoint comes after the route
-        # directly associated to the waypoint
         self.assertEqual(
-            linked_routes[1]['document_id'], self.route3.document_id)
-        self.assertIn('geometry', linked_routes[0])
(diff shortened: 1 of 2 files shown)


brunobesson:
Ok, I think I get the issue: to sort routes by waypoint priority (direct waypoint, then children waypoint, then grandchildren waypoints), the request make a join on the waypoint. If routes are associated to several of these waypoints, we get several rows with same route id, but different associated waypoint.

To distinguish thoses, I had to remove this sort and use DISTINCT ON (route.document_id) rather than a simple DISTINCT.

So we win no-dup routes, but loose the ordering based on associated waypoints. As in paralell we aim at no limiting the number of results returned, except for exceptional or buggy cases (except Calanques waypoint, we never have over 300 associated routes and aim at using a limit of 300 or 400), I do think we can do so, unless someone knows how to solve this best.

What do you think?

brunobesson:
This would fix #589 remaining issue.

lbesson:
imho ordering a few docs in the UI is manageable, since it's not dynamic (but maybe I don't know vue enough on that)

brunobesson:
The problem is not ordering results, but filtering.

Without this fix, API is supposed to return 30 results, but there may be some duplicates within. So the frontend has to filter this (which it does, currently). In addition to this useless hassle, you do not present 30 results, but only a fraction of it. In worst case scenarios, you may have less than half of the results.

P.S. : I am already working on a PR on the frontend side to sort results, that's not a problem.

cbeauchesne (approved):
Well, I do not have any knowledge with SQL Alch. so, I won't be a great help here. 

Sorry for stating the obvious, but I guess you did exhaustive tests on it? ;) 

For my own curiosity, what was the purpose of `literal_column('1').label('priority'))` lines, and why is not not necessary anymore?

brunobesson:
It is used to order results based on the waypoint 'proximity' to the main waypoint : retrieve first routes associated to the main waypoint, then children waypoint then grandchildren ones.
To me it is ok to loose this sort of we augment the number of routes retrieved from 30 to e.g. 400 as proposed. Because only in exceptional circumstances would this sort be useful. And 'big' waypoints such as Calanques or Verdon or Aiguilles de Cham should not have routes directly attached to this main waypoint.

PR #1033, opened 2021-04-12
Migrate to github actions and update libs

lbesson:
WIP


lbesson:
seems to be not breaking at least, I'm merging to test the generated image on demo

Issue #1107, opened 2021-06-15
BUG: missing property 


cbeauchesne:
in `_add_redis_status` function, line 82, prod logs has dazillion lines saying that `client` is not a property.

cbeauchesne:
To reproduce, simply run `curl localhost:6543/health` on dev env

PR #1108, opened 2021-06-15
Fix health view #1107

cbeauchesne:


lbesson:
I can only trust you on that :)

lbesson (approved):

lbesson:
There is one warning from codacy regarding the use of assert.
Being quite incompetent in python, I'll let you have a look

Issue #1133, opened 2021-07-23
Remove Warnings from tests 


cbeauchesne:
Let's close this issue when warning count will be 0 on tests. 

* 2020-07-23: 1932

cbeauchesne:
Status on 2020-07-23: 

Most of warning are DeprecationWarning. And the pain part is that, often, they come from 3rd party lib. For each of them we must check if : 

1. we need to update our code
2. or we need to update the lib

So, for now, I've ignored them, because I think there is no sense to bother devs for day-to-day dev. And I'll try to reduce the count time to time.

And by the way, there is no more warning, except Deprecation warning 😎

PR #1134, opened 2021-07-23
Warning count from 1932 to 1842

cbeauchesne:
See #1133 for more details

### DIFF
--- a/c2corg_api/caching.py
+++ b/c2corg_api/caching.py
@@ -79,6 +79,7 @@ def configure_caches(settings):
             'dogpile.cache.redis',
             arguments={
                 'connection_pool': redis_pool,
+                "thread_local_lock": False,
                 'distributed_lock': True,
                 'lock_timeout': 15,  # 15 seconds (dogpile lock)
                 'redis_expiration_time': int(settings['redis.expiration_time'])
--- a/c2corg_api/markdown/__init__.py
+++ b/c2corg_api/markdown/__init__.py
@@ -136,6 +136,7 @@ def _get_markdown_parser():
         _markdown_parser = markdown.Markdown(output_format='xhtml5',
                                              extensions=extensions,
                                              enable_attributes=False)
+
     return _markdown_parser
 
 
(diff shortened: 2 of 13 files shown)


cbeauchesne:
@tsauerwein by chance, would it possible to have your opinion about this change : https://github.com/c2corg/v6_api/pull/1134/files#diff-bcdcacbeeac9f27f3ad65cda8f9decaee8aa8358a2cc6bf4709d821a2ccd637aR82 

I added it because we have a warning saying  `The Redis backend thread_local_lock parameter should be set to False when distributed_lock is True` > https://github.com/sqlalchemy/dogpile.cache/blob/master/dogpile/cache/backends/redis.py#L108 

Do you know if there any side-effect with this parameter ?

lbesson:
I think that comes from one of the (outdated) dependencies. We already have this warning in the master branch and that still works.

tsauerwein:
> @tsauerwein by chance, would it possible to have your opinion about this change : https://github.com/c2corg/v6_api/pull/1134/files#diff-bcdcacbeeac9f27f3ad65cda8f9decaee8aa8358a2cc6bf4709d821a2ccd637aR82
> 
> I added it because we have a warning saying `The Redis backend thread_local_lock parameter should be set to False when distributed_lock is True` : https://github.com/sqlalchemy/dogpile.cache/blob/master/dogpile/cache/backends/redis.py#L108
> 
> But I have absolutely no idea of its purpose, and it scares me a little bit :)
> 
> Do you know if there any side-effect with this parameter ?

Checking the [dogpile.cache documentation](https://dogpilecache.sqlalchemy.org/en/latest/api.html#dogpile.cache.backends.redis.RedisBackend), your change makes sense. The configuration uses `distributed_lock=True` because the API runs in multiple processes/containers/servers and the lock should be synchronized across all. As I understand it, `distributed_lock` and `thread_local_lock` are mutually exclusive. So it would make sense to set  `thread_local_lock` to `False`.

lbesson (approved):
don't want to update bbcode while you're at it? :D

cbeauchesne:
@lbesson I've a loooooong todo list on API. I'm focusing on the easy one, and on ones that makes dev's life easier, sort of getting a foot in the door in a way :)

Issue #557, opened 2016-11-30
Improve simple search accuracy


tsauerwein:
Reported in https://github.com/c2corg/v6_ui/issues/351#issuecomment-247276509:

>  ... but there is still a problem when searching for "dent de barme" for example: corresponding results are not displayed first:

![image100](https://cloud.githubusercontent.com/assets/20488365/18544434/da15dda4-7b34-11e6-89fa-b8fa61d848b8.png)

asaunier:
Another example that the simple search / association tool behaviour is not satisfaying, making it hard to use:

<img width="586" alt="capture d ecran 2017-02-27 a 15 52 33" src="https://cloud.githubusercontent.com/assets/1192331/23366140/eb5c101c-fd05-11e6-980c-89da5fde495f.png">

Typing "Bénévent face NW", I expect the results to contain all 3 words. Instead of that I get results with absolutely no connection with "Bénévent". For instance "Bionnassay face NW" => as if the words were used as OR clauses instead of AND clauses.

brunobesson:
Another example: "gollachon" returns nothing while "golachon" exists in the database (well, that was the case until I created "Col du Gollachon" in prod - still valid on demo).

This is not related to an AND or OR clause, but I find it strange that such spelling errors are not accepted.

PR #1388, opened 2022-09-28
Fix #755: multi-word search

### LINKED ISSUE #755: Feature: Multi-words search
**Community discussion about this feature (mandatory)**
https://forum.camptocamp.org/t/nouvelle-interface-bugs-questions-suggestions/241148/586?u=florence_b

**Is your feature request related to a problem? Please describe.**
When searching for "word1 word2", the results seem to be results for "word1"&"word2" + results for "word1" + results for "word2" and not results for "word1 word2".

**Describe the solution you'd like**
When "word1 word2" is entered in search input, the first results must contain "word1 word2" the exact string, then "word1" and "word2", then only "word1" or "word2".

jeanphilippeds:
Hello,

This PR fixes #755 and #557.

The bug was due to a bad multi-match request. Here, we want to search for the whole phrase instead of a list of separate words. By adding `operator="and"` (see [doc](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-multi-match-query.html#operator-min)), the issue is solved.

I also added the `fuzziness=auto` parameter (see [doc](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/query-dsl-match-query.html#query-dsl-match-query-fuzziness)) to allow misspellings.

At this point, it looks OK but there is still a minor point:
Capture d’écran 2022-09-28 à 23 59 45

We want the 5th result to be the 1st want as we typed `traversée des calanques`. Due to fuzziness, the first 6 results are "equivalent". It looks like the 5th result is after because it has only one available locale `fr` (but I can't find the real cause as multi_match is using `type=best_field` which keeps only the best field score for each document: only `score(title_fr)` and not `score(title_fr) + score(title_en)`)
Ideally, we would use `fuzziness=auto` and `type=phrase` in order to keep the words order into account. But this is not working.

**Possible next steps**:
- We say it's OK like that
- We investigate [span queries ](https://stackoverflow.com/questions/53652765/elastic-search-match-phrase-fuzziness)

### DIFF
--- a/c2corg_api/search/__init__.py
+++ b/c2corg_api/search/__init__.py
@@ -80,20 +80,6 @@ def create_search(document_type):
         index=elasticsearch_config['index'],
         doc_type=search_documents[document_type])
 
-
-def get_text_query(search_term, lang=None):
-    # search in all title* (title_en, title_fr, ...), summary* and
-    # description* fields. "boost" title fields and summary fields.
-    return Bool(
-        should=[
-            MultiMatch(
-                query=search_term,
-                operator='and',
-                fields=['summary*^2', 'description*'], boost=0.5),
-            get_text_query_on_title(search_term, lang)
-        ])
-
-
 def get_text_query_on_title(search_term, search_lang=None):
     # search in all title* (title_en, title_fr, ...) fields.
     if not search_lang:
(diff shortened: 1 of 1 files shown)


brunobesson:
Thanks for all this work!

I personnally do not touch the API side, because I am no python guy, so any help on the backend side is a must 👍 

As for what to do with your results, I would say let's start with merging this state of the art, because it is already a huge improvement over the current situation.

And if you want to continue investigating then... 😇

brunobesson:
Beware of checks failures however.

jeanphilippeds:
> Beware of checks failures however.

Yep, I'll fix that!

PR #1390, opened 2022-09-30
Strip leading/trailing whitespaces from usernames when registering

### LINKED ISSUE #1383: Validate usernames
It is quite frequent to have users complaining they cannot sign in with their username until one figures out the username has a trailing " " in the database
=> additional username validation should be performed when creating a user account:

- at least strip the leading and trailing whitespaces
- most likely also get rid of the case (case insensitive). Note: there are username duplicates in the database if case is not considered => should we then automatically change those existing duplicates?

TODO :

- [ ] strip leading and trailing whitespaces in usernames (and passwords etc.) when creating an account
- [ ] strip leading and trailing whitespaces in usernames and passwords when authenticating
- [ ] make usernames case-insensitive
- [ ] add migration steps to prevent duplicates in existing usernames

asaunier:
Fixes #1383

### DIFF
--- a/c2corg_api/tests/views/test_user.py
+++ b/c2corg_api/tests/views/test_user.py
@@ -197,6 +197,35 @@ class TestUserRest(BaseUserTestRest):
                          'already used forum_username')
 
     @patch('c2corg_api.emails.email_service.EmailService._send_email')
+    def test_register_stripped_username(self, _send_email):
+        request_body = {
+            'username': ' contributor ',
+            'forum_username': 'Foo',
+            'name': 'Max Mustermann',
+            'password': 'super secret',
+            'email': 'some_user@camptocamp.org'
+        }
+        url = self._prefix + '/register'
+        json = self.app_post_json(url, request_body, status=400).json
+        self.assertEqual(json['errors'][0]['description'],
+                         'This username already exists')
+
+        request_body = {
+            'username': ' username with spaces ',
+            'forum_username': 'Spaceman',
+            'name': 'Max Mustermann',
+            'password': 'super secret',
+            'email': 'space@camptocamp.org'
+        }
+        url = self._prefix + '/register'
+        body = self.app_post_json(url, request_body, status=200).json
+        self.assertBodyEqual(body, 'username', 'username with spaces')
+        user_id = body.get('id')
+        user = self.session.query(User).get(user_id)
+        self.assertIsNotNone(user)
+        self.assertEqual(user.username, 'username with spaces')
+
+    @patch('c2corg_api.emails.email_service.EmailService._send_email')
     def test_register_username_email_not_equals_email(self, _send_email):
         request_body = {
             'username': 'someone_else@camptocamp.org',
--- a/c2corg_api/views/user.py
+++ b/c2corg_api/views/user.py
@@ -68,6 +68,27 @@ def validate_json_password(request, **kwargs):
         request.errors.add('body', 'password', 'Invalid')
 
 
+def validate_json_username(request, **kwargs):
+    """Checks if the username was given, removes leading and trailing
+       whitespaces and eventually checks it's unique.
+    """
+
+    if 'username' not in request.json:
+        request.errors.add('body', 'username', 'Required')
+        return
+
+    username = request.json['username'].strip()
+    if not username:
+        request.errors.add('body', 'username',
+                           'Username cannot be empty or whitespaces')
+        return
+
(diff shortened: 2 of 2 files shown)


c2corg_api/views/user.py
```diff
@@ -58,6 +58,25 @@ def validate_json_password(request, **kwargs):
...
+
+    username = request.json['password'].strip().lower()
+    if not username:
+        request.errors.add('body', 'username', 'Username cannot be empty or whitespaces')
+        return
+
+    if not is_unused_user_attribute('username', username, lowercase=True):
+        request.errors.add('body', 'username', 'This username already exists')
+
+    request.validated['password'] = username
```
brunobesson, line 78: isn't there a typo here?

asaunier: Indeed! Thanks!
And also at line 70.
I have copied-pasted the `validate_json_password` function to create a new one and validate usernames, but forgotten to update it completely :P

c2corg_api/views/user.py
```diff
@@ -442,7 +461,7 @@ def __init__(self, request):
         validators=[colander_body_validator, validate_json_password])
     def post(self):
         request = self.request
-        username = request.validated['username']
+        username = request.validated['username'].strip().lower()
```
asaunier, line 464: Not needed if eventually we decide not to convert the existing usernames with leading or trailing whitespaces.

asaunier:
At the stage the PR at least checks new accounts do not have usernames with leading or trailing spaces + lowercase them.

Next step would be to convert all existing usernames with such leading or trailing spaces (~900) and remove those extra spaces, making sure we don't create duplicates doing so.
We would also have to lowercase all existing usernames, also potentially creating duplicates.

Then we need to figure out how to deal with those duplicates, possibly by adding some suffix (numbering?)? People consequently having troubles to sign in (because their username would have been changed) could follow the "forgotten password?" process, receiving an email including their (updated) username.

I wonder if it is really worth the case?!
If not, I may amend the changes by removing the (embryonic) alembic script file and maybe also cancel the usename stripping and lowercasing when signing in.

brunobesson:
> We would also have to lowercase all existing usernames, also potentially creating duplicates.

Quite easy to test how many it would create using a request on the BD, isn't it? I'll have a look.

IMHO, we should at least keep the changes you propose to introduce for new users.

brunobesson:
```
c2corg=# select count (distinct trim(username)) from users.user;
 count
-------
 89772
(1 row)

c2corg=# select count (distinct username) from users.user;
 count
-------
 89997
(1 row)

c2corg=# select count (distinct trim(lower(username))) from users.user;
 count
-------
 87157
(1 row)
```

So, considering only the spaces, we have few problems. But when using lowercase, it is far worse.

My choice would then be:
* avoid problems with new users (existing users have already dealt with the issue, most of our problems come from users creating new accounts)
* in the futur, allow login with email

c2corg_api/tests/views/test_user.py
```diff
@@ -195,6 +195,35 @@ def test_register_forum_username_unique(self, _send_email):
...
+            'password': 'super secret',
+            'email': 'space@camptocamp.org'
+        }
+        url = self._prefix + '/register'
+        body = self.app_post_json(url, request_body, status=200).json
+        self.assertBodyEqual(body, 'username', 'username with spaces')
+        user_id = body.get('id')
+        user = self.session.query(User).get(user_id)
+        self.assertIsNotNone(user)
+        self.assertEqual(user.username, 'username with spaces')
```
asaunier, line 225: No clue if the last 4 lines are really needed or if the check at line 221 is enough to make sure the user was created with a stripped username.

asaunier:
@brunobesson Thanks for the feedback! I agree.
Therefore I have removed:

- the alembic script as we no longer need to update existing usernames
- the lowercase converting as we still want existing usernames with upper case to be accepted. I have kept the lowercase in the username unicity check though in order to prevent adding more case-related duplicates.

lbesson:
I do think the current behaviour is fine (correct if I misunderstood):
- We don't touch existing usernames (people already managed to workaround potential spaces)
- For new users, you strip username

This looks fine to me. Do you know if browsers save username + password on the register page? Because in that case, someone using a leading or trailing whitespace would try to login with it, and would fail?

lbesson:
@asaunier I merged #1416

asaunier:
@lbesson I have updated this PR to fix the conflicts after the merging of yours and to combine our username validators into a single one. Looks good to you?

lbesson (approved):

PR #1393, opened 2022-10-01
Add support of version masking

asaunier:


asaunier:
UI PR at https://github.com/c2corg/c2c_ui/pull/2947

c2corg_api/views/document_version.py
```diff
@@ -66,23 +69,28 @@ def _load_version(
...
-            schema = adapt_schema(schema, archive_document)
+            if adapt_schema:
+                schema = adapt_schema(schema, archive_document)
+
+            document = to_json_dict(
+                archive_document,
+                schema,
+                cook_locale=True
+            )
+        # TODO if masked, return limited info as in DocumentInfoRest?
```
brunobesson, line 86: this TODO is not yet resolved?

asaunier: No, it's not: as for now when a version is masked, the API returns nothing as the archived documents. See https://github.com/c2corg/v6_api/pull/1393/files#diff-33ce4bfbec662d496fcd00822bfdb6174d23586982306f66c15194dba40f771cR72
Perhaps returning at least the version title as in `DocumentInfoRest` would be nice? See https://github.com/c2corg/v6_api/blob/969bf13ed8e4ad5f0e08694694394e0edba3f313/c2corg_api/views/document_info.py#L17-L20
I have not tried yet to figure out if this is easily feasible though.

asaunier:
There's still an issue as the cache of the versions and of the document history is not refreshed when a version is masked/unmasked.

lbesson (commented):
(Stupid?) remarks / questions, please keep in mind I'm dumb as f*** with python:

- you're not testing that only a moderator can read a version that is masked?
- what about diffs? If we can make the diff between a masked version and another, then it's the same as if we didn't mask it.

brunobesson:
> what about diffs? If we can make the diff between a masked version and another, then it's the same as if we didn't mask it.

As far as I invstigated it some time ago, a moderator will see the diff between all versions, but a normal user will se the diff between unmasked versions only.

asaunier:
@lbesson A version is returned only if not masked or if the user is a moderator. See
https://github.com/c2corg/v6_api/pull/1393/files#diff-33ce4bfbec662d496fcd00822bfdb6174d23586982306f66c15194dba40f771cR72-R85

About the diff: there is no diff view on the API side. It's done on the UI side making requests to the API to get both versions to diff. If one version is masked it won't be exposed on the UI side either when diffing.

lbesson:
> @lbesson A version is returned only if not masked or if the user is a moderator

Yes. I was just saying there are no unit tests added to check that this is properly implemented.

asaunier:
@lbesson Yes there are, see https://github.com/c2corg/v6_api/pull/1393/files#diff-a8b07e15af99ba3249c67ac5324df8bc9506be37f544d2f6e5ebde1be76a0ac2R180-R196

lbesson:
So we're back to the 'stupid' remark qualification :)

asaunier:
@lbesson I had planned to do so when the UI part would be ready but I think we can merge it anyway as the PR only adds a new attribute and new services/tests, not changing the existing API behaviour.
And that might make it easier to have other people test the UI :)

brunobesson:
> How is db migration handled?

https://github.com/c2corg/v6_api/wiki/Upgrade-the-prod-server#full-how-to

asaunier:
Thanks @brunobesson for the link to the doc.
In short, when upgrading the demo/prod API, one has to activate the maintenance mode and run the alembic ugrade command manually.

asaunier:
@lbesson @brunobesson I will merge the PR this evening if it's OK for you.

PR #1394, opened 2022-10-04
Fix #1101: Add period filter

### LINKED ISSUE #1101: Add the possibility to filter outings by a period of the year
For example, get all outings in may, or between the 15th and 20th of december, without having to stick to a specific year

linked issue on the UI: https://github.com/c2corg/c2c_ui/issues/1979

jeanphilippeds:
Hello,

I managed to fix #1101 to add the possibility to filter by date range regardless of the year.

First of all, it is not possible with the elastic native `DateRange` filter. So, I saw 2 options:

**1. Update indexation**
- Index 2 new fields `date_start_day_of_year` and `date_end_day_of_year` that contains the date's day number of the year (eg: 1 for 01/01 and 288 for 15/10).
- Migrate all the existing documents to fill those 2 fields.
- Add a new period filter.

**2. Update only the query**
- Compute the position of a date from the existing date field (which value contains the number of milliseconds since epoch). For that, I calculate the `(number of milliseconds since the beginning of the year) = (milliseconds since epoch) % (milliseconds in a year)`.
- Add a new period filter using this calculation.

I chose option n°2 for two reasons:
- Less code to update (so less risky)
- Easier to implement/deploy

Note: As I had to update ElasticSearch options (see updates in `docker-compose.yml`), we will have to do it on the deployed instances (demo & prod). How can we achieve this? Is it deployed also with Docker?

Next step: use this filter to fix: https://github.com/c2corg/c2c_ui/issues/1979


brunobesson:
> For that, I calculate the (number of milliseconds since the beginning of the year) = (milliseconds since epoch) % (milliseconds in a year).

Isn't there in Python a library to handle datetimes properly rather than using this 'fake' computation with a quarter of a day more per year? (remember I do not code in python)

> Note: As I had to update ElasticSearch options

How are these modifications relevant? I don't understand the need of the host configuration, nor the inline stuff?

Another point: could you add some tests?

Anyway, great work 👍

jeanphilippeds:
> > For that, I calculate the (number of milliseconds since the beginning of the year) = (milliseconds since epoch) % (milliseconds in a year).
> 
> Isn't there in Python a library to handle datetimes properly rather than using this 'fake' computation with a quarter of a day more per year? (remember I do not code in python)
> 
I could use a library but only in the Python part. The inline script (sent to elastic) is Java code and we can't use any lib in the json query.

> > Note: As I had to update ElasticSearch options
> 
> How are these modifications relevant? I don't understand the need of the host configuration, nor the inline stuff?
> 
2 lines here:
- `network.host` which was already here in the pulled docker image for elastic
- `script.inline` allows to use an inlined script in our query (which is the case here)

> Another point: could you add some tests?
> 
Yes, for sure!
> Anyway, great work 👍

brunobesson:
Once merged, what actions should be done on the deployment? I confess I have forgotten what you mentioned earlier and prefer you to list it again clearly 🙂 
@lbesson

jeanphilippeds:
> Once merged, what actions should be done on the deployment? I confess I have forgotten what you mentioned earlier and prefer you to list it again clearly 🙂 @lbesson

@brunobesson Once deployed, we need to check that elasticsearch engine has the `script.inline: true` option.
To check it, we can call:
`https://api.demov6.camptocamp.org/outings?bbox=-431698,3115462,1931123,8442818&period=10-05,10-07` and expect results from 10-05 to 10-07.

jeanphilippeds:
> ok, I guess best is now to merge and try to deploy a new version of the api in the demo env, by modifying `-Des.script.inline=false` to `-Des.script.inline=true`
> 
> First we should merge and produce a version. Maybe it's worth waiting the ongoing PR from @asaunier on the login?
> 
> Regarding elastic search, I don't know if the network host option is necessary (it worked before?) Nevertheless could you modify a bit your PR so that all options are passed via -D, or via the yaml configuration file, but not spread in several places?

Good idea @lbesson , I did not think about it :)

I pushed a new commit, you can merge when you want.

PR #1416, opened 2022-10-17
Allow to use email as login

lbesson:
Also prevents creating new accounts with a username that would be an email-like, except if that's the same as the user's email

### DIFF
--- a/c2corg_api/tests/views/test_user.py
+++ b/c2corg_api/tests/views/test_user.py
@@ -102,7 +102,8 @@ class TestUserRest(BaseUserTestRest):
     @patch('c2corg_api.emails.email_service.EmailService._send_email')
     def test_always_register_non_validated_users(self, _send_email):
         request_body = {
-            'username': 'test', 'forum_username': 'test',
+            'username': 'test',
+            'forum_username': 'test',
             'name': 'Max Mustermann',
             'password': 'super secret',
             'email_validated': True,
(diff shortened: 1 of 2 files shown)


brunobesson (approved):
Looks good to me (with limited language skills). Maybe sync with @asaunier which is also working on the login part?

c2corg_api/views/user.py
```diff
@@ -113,6 +122,24 @@ def validate_forum_username(request, **kwargs):
             request.errors.add('body', attrname, res)
 
 
+def validate_username(request, **kwargs):
+    """
+    Check that the username is not an email,
+    or that it is the same as the actual email.
+    """
+    attrname = 'username'
+    if attrname in request.json and 'email' in request.json:
```
asaunier, line 131: That's not very consistent to use a variable for username but not for email. Why not simply write `'username'` everywhere as you did for `'email'`?
That would be more readable I think.

c2corg_api/views/user.py
```diff
@@ -113,6 +122,24 @@ def validate_forum_username(request, **kwargs):
...
+        value = request.json[attrname]
+        email = request.json['email']
+        if (valid_email(value) and email != value):
+            request.errors.add(
+                'body',
+                attrname,
+                'You cannot use someone else\'s email as username')
```
asaunier, line 138: That could be your email address though, for instance if you have several...
The message should rather be something like:
`An email address used as username should be the same as the one used as the account email address.`.

c2corg_api/views/user.py
```diff
@@ -166,11 +193,13 @@ def __init__(self, request):
             partial(validate_unique_attribute,
                     "forum_username",
                     lowercase=True),
+            validate_username,
             validate_forum_username,
             validate_captcha])
     def post(self):
         user = schema_create_user.objectify(self.request.validated)
         user.password = self.request.validated['password']
+        user.email = self.request.validated['email']
```
asaunier, line 202: I don't think this line is useful, as you make no change on the email attribute anyway.
And it's whitelisted in https://github.com/c2corg/v6_api/blob/master/c2corg_api/models/user.py#L180
Not sure exactly what it means but I think it means the attribute is directly created from ``self.request.validated`` by the 2 lines above:
``user = schema_create_user.objectify(self.request.validated)``

c2corg_api/tests/views/test_user.py
```diff
@@ -195,10 +196,51 @@ def test_register_forum_username_unique(self, _send_email):
         self.assertEqual(json['errors'][0]['description'],
                          'already used forum_username')
 
+    @patch('c2corg_api.emails.email_service.EmailService._send_email')
+    def test_register_username_not_email(self, _send_email):
```
asaunier, line 200: To be consistent with the following test, it should rather be `test_register_username_email_not_equals_email`

c2corg_api/tests/views/test_user.py
```diff
@@ -195,10 +196,51 @@ def test_register_forum_username_unique(self, _send_email):
...
+            'forum_username': 'Contributor4',
+            'name': 'Frankie Vincent',
+            'password': 'super secret',
+            'email': 'some_user@camptocamp.org'
+        }
+        url = self._prefix + '/register'
+        self.app_post_json(url, request_body, status=200)
+
+    @patch('c2corg_api.emails.email_service.EmailService._send_email')
+    def test_register_invalid_email(self, _send_email):
```
asaunier, line 226: This one was missing? Gasp :P 
Good catch then :)

asaunier (commented):
A few comments but overall the PR looks good to me :) 
I'm afraid it will be conflicting with my pending PR https://github.com/c2corg/v6_api/pull/1390 though :P

c2corg_api/views/user.py
lbesson: I tried to be a bit consistent with other cases, but agreed

c2corg_api/views/user.py
lbesson: k

c2corg_api/views/user.py
```diff
@@ -34,6 +34,15 @@
 MINIMUM_PASSWORD_LENGTH = 3
 
 
+def valid_email(email):
```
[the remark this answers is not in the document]
lbesson: > I'm afraid it will be conflicting with my pending PR #1390 though :P

It does conflict as such, but I don't think that they are incompatible, do they?
Ho would you rather proceed? Which PR should we merge to master first?

lbesson:
> I'm afraid it will be conflicting with my pending PR https://github.com/c2corg/v6_api/pull/1390 though :P

It does conflict as such, but I don't think that they are incompatible, do they?
Ho would you rather proceed? Which PR should we merge to master first?

asaunier:
That's probably not a big deal combining them. but for instance I think we'll have to combine the ``validate_username`` functions introduced by both PRs.

Please merge your PR first and I will adapt mine then.

lbesson:
Ok, I'll take your comments into account and merge.

But in the end, my PR is mainly at the login level. Yours at the register one. The only thing I do at the register level is prevent someone using an email for a username (except if that's the same)

@brunobesson could you easily look in the database if we have currently usernames that match an email from another account? (I doubt it)

brunobesson:
> could you easily look in the database if we have currently usernames that match an email from another account? (I doubt it)

Alas... 68 results!

```sql
c2corg=# select u1.id, u1.username, u1.email, u2.id, u2.username, u2.email from users.user u1 inner join users.user u2 on u1.email = u2.username where u1.email != u2.email;
```

Looking at it, it looks like in all (most?) cases, it consists of people who created duplicate accounts, using either the same email but changing capitals, adding/rmoving dots, or, in most cases another email.
We must think on how to solve this...

PR #1420, opened 2022-10-22
WIP: Add bouldering rating in database

lechinoix:
Linked to this MR : https://github.com/c2corg/c2c_ui/pull/2971
And this issue : https://github.com/c2corg/c2c_ui/issues/2552

DIdn't figure out yet how to run migrations (it will run on start ? Did not spawn docker yet)

Need then to migrate existing bouldering cotation from `rock_free_rating`


asaunier:
Hi @lechinoix 

Migrations are run using Alembic. It's documented at https://github.com/c2corg/v6_api/tree/master/alembic_migration
You have to run the migration tool manually.

As the bouldering ratings are very similar to "rope" ratings, wouldn't it be simpler to keep using the same rating attribute but to capitalize it on the UI side when the "route_type" is "boulder"?

BorutK:
Yup. The point (and probably a problem) is not so much about capitalizing or not. The specific of Bleau bouldering grades (which ARE accepted since they even were the start for any kind of purely climbing grades) is the fact that the "letter following a number system" only starts with 6a (sorry 6A). So from 1 to 5 it's : **1-** , **1** , **1+** , **2-** ... 
I on purpose gave the example with **1** because wikipedia in its comparison table puts grades 1, 2 and 3 in a bundle, which is not what is found in real life. Those low grades (with 3 subdivisions) are important to point out : think of small kids or absolute greenhorns. 
In any case, it's nice to see that people care about all this. 

Thx to all a ya !

Borut

PR #1423, opened 2022-10-24
Update markdown minor version

lbesson:


lbesson:
oops, that was the patch version.

anyway, ...

PR #1458, opened 2022-11-24
Remove useless values in test_search.py

cbeauchesne:
* Articles and books can't have geometry (and I'm a little bit surprised to see the data model accepts it...)
* fixing document id on test setup is useless in test setup (and painful to handle for any revamp)


asaunier:
No need for geometry for articles and books, indeed.
No clue for the document_id attribute.

PR #1502, opened 2023-01-30
Increment document views when getting a document

jaweherCherif:
This PR is in progress.
For now I :
- Created a new table document_views which store the view_count and enable_view_count of each document
- Incremented the view_count when getting a document

### DIFF
--- a/alembic_migration/versions/576e341ee94d_add_document_views_table.py
+++ b/alembic_migration/versions/576e341ee94d_add_document_views_table.py
@@ -1,0 +1,30 @@
+"""add document_views table
+
+Revision ID: 576e341ee94d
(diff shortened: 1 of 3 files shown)


cbeauchesne:
Hi @jaweherCherif,

Thank you for the PR. Unfortunately, I see a major issue : it will perform a write transaction with the DB for every view on every page. As you do not have any performance test tool, we cannot ensure it'll be performant enough on production context :( 

As now, the way the API works makes write access only for page updates (maybe 0.001% of all requests), and read access very rare too, as it use lot of caching.

I'm 99.99% sure that moving to 1 request = 1 write request to the DB will cause lot of trouble.

PR #1523, opened 2023-02-12
add version_date to /outing response

jaweherCherif:


lbesson:
this looks good, do you think you could add some unit tests?

PR #1542, opened 2023-03-03
ToS acceptation

lbesson:
This PR forces the validation of ToS

Existing users won't have tos_validated field set
On login, if user hasn't accepted ToS, a 403 with specific message will be returned (to be caught by UI)
The UI should ask to login again, with a small explanation + ToS checkbox

Once the login + accept ToS is done, the user can access camptocamp, and the date is stored in the db.

Note: do we know since when the ToS checkbox is there on register? We should skip this for such users (or more precisely, store a tos_validated field corresponding to user creation?)


c2corg_api/views/user.py
```diff
@@ -486,15 +490,37 @@ def post(self):
...
             user = DBSession.query(User). \
                 filter(User.email == username).first()
 
         token = try_login(user, password, request) if user else None
         if token:
+            # Check if the user has validated Terms of Service, if not,
+            # return a 403 with an explicit message that can be caught
+            # by the frontend
+            if user.tos_validated is None and accept_tos is not True:
+                raise HTTPForbidden('TOS not accepted')
```
brunobesson, line 508: I haven't checked the other error messages, but do we really need to shorten to `TOS`?
Couldn't `Terms for service not accepted` be used instead?

brunobesson (approved):

brunobesson:
> Note: do we know since when the ToS checkbox is there on register? We should skip this for such users (or more precisely, store a tos_validated field corresponding to user creation?)

I'll check

c2corg_api/views/user.py
lbesson: sure

brunobesson:
@lbesson what's the status on this? https://github.com/c2corg/c2c_ui/pull/3838 does the frontend part. Could we merge this PR? Is there sil some work needed?

lbesson:
mostly asked to be put on hold by Loic, probably this should be fine now

PR #1726, opened 2024-01-13
feat: increment documents views using Redis queue and scheduled job

florentcadot:
The beginning of a potential solution for https://github.com/c2corg/c2c_ui/issues/2648.

~~I've taken some chunk of code written by @jaweherCherif here https://github.com/c2corg/v6_api/pull/1502 for the moment.~~

## In a nutshell

Taking into account @cbeauchesne 's [comment](https://github.com/c2corg/v6_api/pull/1502#issuecomment-1413555114)  + the [specs](https://forum.camptocamp.org/t/cahier-des-charges-compteur-de-vues-pour-les-sorties/305151/18?u=loic_p) from the forum I've used a Redis queue (created similarly to the one used by the syncer) to buffer all the documents viewed and created a job (similar to the 2 background jobs) that update the documents views every 5 minutes (<-- interval chosen randomly for now). The counter of each document viewed is increase only if the client is not a c2c bot or a crawler.  

## Bulk update

For the moment each bulk update query will update 4000 documents max by default. So if we have to update an amount of docs > 4000 it will be divided into several bulk update query that will update 4000 documents max each.

## Performance

Here is a small project to have an idea of how much time it gets to perform the bulk updates using this approach:
https://github.com/florentcadot/c2c-update-view-counter-benchmark/tree/master.
The idea is just to push some documents ids to a Redis Queue, perform the bulk update and measure the amount of time required to perform this update thanks to sqlalchemy [profiling](https://docs.sqlalchemy.org/en/13/faq/performance.html#query-profiling). It uses the same dev database and same libs versions than this project.
It is possible to change the number of docs to update and the bulk update amount (4000 by default from now) to try to find a sweet spot.

Here are some results:
- For 100 documents to update:  **0.001287s**   
- For 1000 documents to update: **0.011257s**
- For 10000 documents to update:  **0.076611s** (bulk update of 4000 docs n°1) **0.040134s** (bulk update of 4000 docs n°2)  **0.015170s** (bulk update of 2000 docs n°3) 

## In another PR:

See https://github.com/c2corg/v6_api/pull/1731

- [x] Update API route to allow authorized users (ACL management) to update ~~`enable_view_count`~~ `disable_view_count state` 
- [x] Get API route to allow authorized users (ACL management) to read the counter of a document if ~~`enable_view_count`~~ `disable_view_count` is ~~True~~ False       

Feel free to give a feedback anytime :smile:   

EDIT: use `disable_view_count` instead of `enable_view_count` to be consistent with `disable_comments` outing column and add other PR link

### DIFF
--- a/alembic_migration/versions/2d710311ac56_add_view_count_to_document.py
+++ b/alembic_migration/versions/2d710311ac56_add_view_count_to_document.py
@@ -1,0 +1,30 @@
+"""add view count to document
+
+Revision ID: 2d710311ac56
+Revises: 305b064bdf66
+Create Date: 2024-01-08 13:28:50.090334
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '2d710311ac56'
+down_revision = '305b064bdf66'
+branch_labels = None
+depends_on = None
+
+def upgrade():
+    op.add_column(
+        'documents',
+        sa.Column(
+            'view_count',
+            sa.Integer(),
+            nullable=False,
+            server_default='0',
+        ),
+        schema='guidebook')
+
+def downgrade():
+    op.drop_column('documents', 'view_count', schema='guidebook')
--- a/c2corg_api/__init__.py
+++ b/c2corg_api/__init__.py
@@ -6,7 +6,8 @@ from sqlalchemy import engine_from_config, exc, event
 from sqlalchemy.pool import Pool
 
 from c2corg_api.models import DBSession, Base
-from c2corg_api.search import configure_es_from_config, get_queue_config
+from c2corg_api.queues.queues_service import get_queue_config
+from c2corg_api.search import configure_es_from_config
 
 from pyramid.security import Allow, Everyone, Authenticated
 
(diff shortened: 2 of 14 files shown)


brunobesson:
I am not fluent in python so won't review the PR, but thanks a lot for taking a look into this.

@lbesson and @asaunier maybe?

lbesson:
i'll try to make a review asap, but I'm not really fluent in python either :)

eddy-geek:
As an experienced python dev I can say the code is quite clean, even beautiful :) 
but having zero c2c backend experience I'd need more time for a useful review ;-)

personally I was not really worried about the perf impact (maybe I underestimate traffic?) but indeed the batch design  should alleviate this concern.

lbesson:
coming back a long time after...

@florentcadot do you think we could merge this and test it in the demo env?

florentcadot:
> coming back a long time after...
> 
> @florentcadot do you think we could merge this and test it in the demo env?

@lbesson sounds good, worst case scenario we revert

loicperrin:
@florentcadot some checks are not successful. Could you fix them so that we can merge?

PR #1747, opened 2024-02-18
feat: add external_resources column to waypoints

florentcadot:
https://github.com/c2corg/c2c_ui/issues/3697

## How to test:

Run the added migration using `docker-compose exec api .build/venv/bin/alembic upgrade head`
Pull the associated front branch: feature/#3697-add-waypoints-external-resources

From the front app, create a new waypoint and fill the `external_resources` field. Then update the external resources of this.

### DIFF
--- a/alembic_migration/versions/626354ffcda0_add_column_external_resources_to_waypoints.py
+++ b/alembic_migration/versions/626354ffcda0_add_column_external_resources_to_waypoints.py
@@ -1,0 +1,37 @@
+"""add_column_external_resources_to_waypoints
+
+Revision ID: 626354ffcda0
+Revises: 305b064bdf66
+Create Date: 2024-02-18 08:36:44.714268
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '626354ffcda0'
+down_revision = '305b064bdf66'
+branch_labels = None
+depends_on = None
+
+def upgrade():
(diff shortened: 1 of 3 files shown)


brunobesson (approved):
LGTM, although I am not fluen with the backend. But the modifications ook consistent, and you tested it already :)

lbesson:
lgtm, maybe we could add some unit tests around this? That shoudln't be too complicated (mostly copy pasting how it's done with other documents?)

florentcadot:
> lgtm, maybe we could add some unit tests around this? That shoudln't be too complicated (mostly copy pasting how it's done with other documents?)

You're right!

I've added a simple model unit test in `c2corg_api/tests/models/test_waypoint.py`. 
Do you want me to test if `Waypoint` instantiation with wrong local external_resource does not work ?
Maybe I should add some tests in `views/test_waypoint` as well ?

eddy-geek (approved):

florentcadot:
> Could you lint the files so I can merge?


Do you need me to also fix the codacy error or it's fine ?

image

(It's just a convenient way to destructure the tuple returned by the function)

eddy-geek:
> Do you need me to also fix the codacy error or it's fine ?
> 
> (It's just a convenient way to destructure the tuple returned by the function)

FYI `_` as variable name is a usual way to signal you don't use it, tools will usually understand it.

ie
```py
(_, waypoint) = self.put_success_all( ...
```

PR #1760, opened 2024-03-08
fix: c2corg#1226 prevent climbing indoor waypoints association with routes

Nayor:


c2corg_api/views/validation.py
```diff
@@ -624,18 +643,25 @@ def _add_associations(
...
-                    'is_parent': is_parent
-                } for doc in associations_in[document_key]
-                ]
+            return
+        
+        if not _are_valid_waypoints(associations_in, main_document_type, document_key):
+            errors.add(
+                'body', 
+                'associations.climbing_indoor_waypoint', 
+                'climbing_indoor waypoint cannot be linked to a route')
```
brunobesson, line 652: this will require to add those translations for the frontend. I'll try not to forget it when merged.

c2corg_api/views/validation.py
```diff
@@ -604,6 +604,25 @@ def _get_linked_document_ids(associations):
...
+    waypoints_id = [doc['document_id'] for doc in associations_in[document_key]]
+
+    # load the waypoint type for each linked waypoint
+    if waypoints_id:
+        query_waypoints_with_type = DBSession. \
+            query(Waypoint.waypoint_type). \
+            filter(Waypoint.document_id.in_(waypoints_id))
+        for waypoint in query_waypoints_with_type.all():
+            if waypoint.waypoint_type == 'climbing_indoor':
+                return False
```
brunobesson, line 622: maybe the function could return the error description instead of a boolean, so that it can be more easily handled on line 648 if several cases arise?

currently, the error handling there is not really clear because the function name doesn't explicit it handles a single case

brunobesson (commented):
Thanks again for this PR.

I do think we could make the code more explicit by expressing it a little bit diffeently. Otherwise, looks 👌

brunobesson:
Please check the linter here too.

Nayor:
> Thanks again for this PR.

My pleasure !

> I do think we could make the code more explicit by expressing it a little bit diffeently. Otherwise, looks 👌

I agree, I wasn't really certain about the refactoring, if you have any suggestion, I'm taking it !

> I wonder if we could also fix the input part, i.e. not offer to associate indoor climbing sites to routes?

It's already done (by [#2422](https://github.com/c2corg/c2c_ui/pull/2422)) isn't it ?
https://github.com/c2corg/c2c_ui/blob/master/src/js/associations-rights-mixin.js#L21
Or I am missing something ?

brunobesson:
> It's already done (by https://github.com/c2corg/c2c_ui/pull/2422) isn't it ?

You're right, it should be enough like this.

c2corg_api/views/route.py
```diff
@@ -63,7 +63,7 @@ def validate_main_waypoint(is_on_create, request, **kwargs):
 
     # no association found
     request.errors.add(
-        'body', 'main_waypoint_id', 'no association for the main waypoint')
```
brunobesson, line 66: must check the translation on the frontend part when merged

brunobesson:
> any suggestion

* the function `_are_valid_waypoints` doesn't really need the `document_key` argument, does it? It could be filtered before caling the function (because its name indicates it deals only with waypoints)
* instead of returning a boolean, it could return either `null` or the error key. This way you still build the error in the `_add_associations` method, but you can handle other putative cases.

What do you think?

c2corg_api/views/validation.py
Nayor: I've changed the fonction name, so it is more clear that it handles a single case

c2corg_api/views/route.py
Nayor: I didn't found any translation for this error

Nayor:
> * the function `_are_valid_waypoints` doesn't really need the `document_key` argument, does it? It could be filtered before caling the function (because its name indicates it deals only with waypoints)

True, I wanted to make it kind of generic but if it became something specific, it's not needed anymore

> * instead of returning a boolean, it could return either `null` or the error key. This way you still build the error in the `_add_associations` method, but you can handle other putative cases.

The error handling if a bit heavy for this single verification, IMHO. May be it has to be done if this function have to cover more cases in the futur ?

brunobesson:
> The error handling if a bit heavy for this single verification, IMHO. May be it has to be done if this function have to cover more cases in the futur ?

Agreed, coud be handled later.

c2corg_api/views/route.py
brunobesson: thus it is not translated by the frontend yet 🙂

brunobesson (approved):

PR #1767, opened 2024-03-20
Update docker-compose services

BboyKeen:
Hey there,

You'll find in this PR a few updates 

- Schema update scripts are inline with the use of the standard `postgres` user.
- $USER variables are not used anymore in schema update scripts
- Previously built images for Postgres/Postgis and Redis are not used anymore in docker-compose
- We use the dedicated volumes feature for Postgres data now
- The obsolete links directive was removed from the docker-compose

All tests are green in local with these updates

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         image: postgis/postgis
         env:
           POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: postgres
+          POSTGRES_PASSWORD: test
         options: >-
           --health-cmd pg_isready
           --health-interval 10s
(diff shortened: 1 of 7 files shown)


lbesson:
Hello, thanks for the PR (and the other ones)

I didn't see a notification, thus only seeing this now.

I'm wondering about the postgre image change: this image contains anonymized data from an old extract of c2c db, it's not used in the tests?

BboyKeen:
Interesting. I may have missed something for this part. However all tests are green without it so I think it's only used as dev fixtures.

I can revert this part if necessary but this image is quite old and it's not easily updatable..

Maybe we should prefer having a dumped anonymized SQL file to load at dev database init.

lbesson:
Hello, again sorry for having been very absent recently

I think we can merge this anyway, all tests pass, and I don't really see why we would need existing data for tests that don't use it

romain-dartigues:
Should been included through #1988 (well, at least I hope so, my eyes are getting heavy; please reach us out if needed).

PR #1769, opened 2024-03-21
[WIP] Refactor the Makefile

BboyKeen:
This PR brings in a new Makefile structure to offer a simpler development environment.
For the sake of simplicity and readibility, I purposefully used the Makefile as a task runner and removed almost all rules dependencies.

It introduces two Makefiles : one for the local environment and the other to build the production container.

Instead of relying on config files, we now rely on default env files. I introduced a possibility to override defaults by using a `config/env.local` which will be automatically read by the Makefile.

- **The production Makefile still needs testing to make sure it bootstraps the production container properly.**
- **Same for Github Actions**

---
This PR depends on #1767 and on #1768


romain-dartigues:
Should been included through #1988.

Issue #1067, opened 2021-04-29
Add public transport information in waypoints (accesses) listing


brunobesson:
See #https://github.com/c2corg/c2c_ui/issues/846

Field `public_transportation_rating` should be included. I assume we do not need the `public_transportation_types` field.

AntoineMarnat:
Any progress on this ?

cbeauchesne:
nope ...

florentcadot:
Hey!
My understanding is that we should add a `public_transportation_rating` in the `Route` table. It would make sense to also add a field in the `Route` create/update form front-end side, do you confirm ?

eddy-geek:
mmm I'm discovering this, but I don't think new columns are needed, instead the existing `public_transportation_rating` data of the waypoints can be reused?

![image](https://github.com/c2corg/v6_api/assets/2772505/88e0480d-bfcf-4bfc-9aaf-aabec96236f7)

I'm guessing a *route* is accessible by public transport if at least one of its linked *waypoints* is accessible?

florentcadot:
> mmm I'm discovering this, but I don't think new columns are needed, instead the existing `public_transportation_rating` data of the waypoints can be reused?
> 
[...]

PR #1797, opened 2024-05-14
feat: c2corg#1067 add public transport information in routes

Nayor:


eddy-geek:
xref #1067 and c2corg/c2c_ui#3936 for frontend

c2corg_api/views/route.py
```diff
@@ -282,3 +312,206 @@ def set_title_prefix_for_ids(ids, title):
     """
     DBSession.query(RouteLocale).filter(RouteLocale.id.in_(ids)). \
         update({RouteLocale.title_prefix: title}, synchronize_session=False)
+
+
+def update_all_pt_rating(waypoint_extrapolation=True):
```
eddy-geek, line 317: a bit worried about using the ORM for such a giant query... best way to run this would be in batches ?

Nayor: @eddy-geek 
The first request runs in 3209.054 ms on my laptop
The whole process is done in a few minutes and is supposed to be used only once, afterwards the informations are supposed to be automaticaly updated on route/waypoint creation/updates

With python log level = INFO and postgresql log level to all, the whole process took about 2-3 hours on my laptop and generates huge amount of logs (since that for each route, every linked waypoints are read and the route updated)

Nayor: May be the best is to not expose this function to the REST API and run it with a command line ? So only devs can run it

eddy-geek: did you run your tests on the full production database?

if it's just a one-time migration then yes we should not leave it in the API, but I guess it's ok to leave it as-is for now and remove the API part after we're done with the initial migration

eddy-geek (approved):
looks great.

c2corg_api/views/route.py
Nayor: I have run the tests on the docker's data, I don't have access to the production database data

The other way to make the migration is to call the function using a python command

Nayor:
> What's the status there? Should be merged, isn't it?

From my point of view, it could be merged ! It seems @eddy-geek had checked the code some time ago

eddy-geek:
yess

PR #1815, opened 2024-06-17
Update deletion time from 24 to 720 hours

AntoineMarnat:
Solves https://github.com/c2corg/c2c_ui/issues/3976


eddy-geek:
nitpick: you forgot to rename the enclosing function: `def is_less_than_24h_old(document_id):`

PR #1860, opened 2024-08-24
Implement new LTag aliases: C# and V# for crag climbing routes

bohwaz:
This commit adds two new syntaxes for tables of lines, intended for use with crag climbing routes:

```
V#1 | Name A | 6a+
C#2 | Name B | 5b
```

V is for "Voie", C is for "Climb", as R (could be used for Route) is already used for "Relay"  and L (could mean "Line") is used for "Longueur".

When transformed to HTML, the V or C is removed, resulting in this:

> | - | - | - |
> | -- | -- | -- |
> | 1 | Name A | 6a+ |
> | 2 | Name B | 5b |

(Table header added here to make it work with Github markdown syntax)

Apart from this, there is no difference with the handling of regular `R#`  and `L#` tags.

This is useful for writing topos for regular crags, where a climb can have multiple lengths (longueurs) and relays. It's closer to what you would find in topo guidebooks.

I added a new test to make sure the aliases work properly.

### DIFF
--- a/c2corg_api/markdown/ltag.py
+++ b/c2corg_api/markdown/ltag.py
@@ -27,6 +27,8 @@ def _get_ltag_pattern():
     Build the big ugly fat regexp for L# numbering. It's based on named
     patterns : (P?<pattern_name>pattern) and decomposed part by part.
 
+    C#/V# numbering is for crag routes (climb/voie)
+
     Please have a look on
     https://forum.camptocamp.org/t/question-l/207148/69
     """
@@ -53,7 +55,7 @@ def _get_ltag_pattern():
     text_in_the_middle = p("text_in_the_middle", "~")
     header = p("header", "=")
 
-    typ = p("type", "[LR]")
+    typ = p("type", "[LRVC]")
 
     text = "(" + header + "|" + text_in_the_middle + "|" + numbering + ")"
 
@@ -78,6 +80,7 @@ class LTagNumbering(object):
     FORMAT = ('<span class="pitch">'
               '<span translate>{type}</span>'
               '{text}</span>').format
+    FORMAT_CLIMB = ('<span class="pitch">{text}</span>').format
     FORMAT_UNMATCHED = '<code>{}</code>'.format
 
     def __init__(self, markdown):
@@ -85,7 +88,7 @@ class LTagNumbering(object):
         self.get_placeholder = markdown.htmlStash.store
 
         # Values for relative patterns
-        self.value = {"R": 0, "L": 0}
+        self.value = {"R": 0, "L": 0, "V": 0, "C": 0}
 
         # One way switch
         self.supported = True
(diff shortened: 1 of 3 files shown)


cbeauchesne (approved):

cbeauchesne:
Hi @bohwaz ,

Thanks for your work. I've approved the PR, as the original writer of this extension to confirm that your work is fine. 

Though, I'm not anymore part of c2c anymore, so I can't make the decision to merge it. You can send a message to the developper forum on c2c to get someone to merge it if nobody answers here.

Best!

bohwaz:
Thank you @cbeauchesne I'll do that :)

eddy-geek:
however as mentioned in the forum there is a standard markdown extension / widely adopted syntax for table

eg the githiub syntax using ` | --- | --- |` as header and `|:--- | ---:|` for let/right alignment

eg there is this extension of the markdown engine we use, we could try to integrate it: https://python-markdown.github.io/extensions/toc/

that would more in-line with most of the existng symtax is standard markdown, and less confusing for new users who know markdown

cbeauchesne:
> eg there is this extension of the markdown engine we use, we could try to integrate it: https://python-markdown.github.io/extensions/toc/

You probably mean https://python-markdown.github.io/extensions/tables/ ?

bohwaz:
> however as mentioned in the forum there is a standard markdown extension / widely adopted syntax for table
> 
> eg the githiub syntax using ` | --- | --- |` as header and `|:--- | ---:|` for let/right alignment
> 
> eg there is this extension of the markdown engine we use, we could try to integrate it: https://python-markdown.github.io/extensions/toc/
> 
> that would more in-line with most of the existing syntax which is standard markdown, and less confusing for new users who know markdown

I would be happy to have markdown tables as well, but we need to have some semantic meaning to identify climbing routes.

Or I won't be able to automatically generate our local climbing topos books from the data I am entering in Camp2camp.

I am sad that the forum guys pushed against this. I spent a considerable amount of time on this, and it just seems we can't get to have any evolution, which makes camptocamp irrelevant if it can't suit the open data needs of our days.

bohwaz:
Adding the link to the forum: https://forum.camptocamp.org/t/merge-pr-pour-les-tags-v-and-c/337899

cbeauchesne:
> I am sad that the forum guys pushed against this. I spent a considerable amount of time on this, and it just seems we can't get to have any evolution, which makes camptocamp irrelevant if it can't suit the open data needs of our days.

I totally understand this feeling, and I deeply regret that the workflow `decision -> implementation` didn't worked as it should.

> Or I won't be able to automatically generate our local climbing topos books from the data I am entering in Camp2camp.

Could you give more details about those difficulties ? Maybe I could help, I've handled the post-migration v5 -> v6 and cleaned the whole DB (it was in a terrible shape after the migration), I may have some idea to help you to tackle this.

PR #1969, opened 2025-03-23
Make ext colander tests automatically run again

romain-dartigues:
The ext-colander tests were not run because they failed to be [discovered by Pytest](https://docs.pytest.org/en/8.3.x/explanation/goodpractices.html#conventions-for-python-test-discovery).

If fixed them the best I could understand on the spur of the moment.
Ping @lbesson


lbesson:
?

not sure to follow what I'm supposed to say / understand?

PR #2042, opened 2025-07-18
Smart origin/prod

eddy-geek:


get_public_transports_from_France.sh
```diff
@@ -0,0 +1,203 @@
+#!/bin/bash
+
+# Configuration
+SERVICE_NAME="postgresql"
```
eddy-geek, line 4: used with compose, so needs to be adjusted

PR #2226, opened 2026-09-02
fix: Smart/Origin: apply 5km fix on bm script

lgourdin:


eddy-geek (approved):

eddy-geek:
@lgourdin do we need to rerun this script ?

lgourdin:
you can rerun it to fix the distance issue, but you can also directly apply the fix on the DB if you don't want to spend navitia credits.
As i mentioned it in our ticketing platform (issue number 2208), you can run `delete from guidebook.waypoints_stopareas where distance > 5;` directly on the prod database to avoid rerunning the whole script.

